PR #5, opened 2021-06-09
Add basic CLI for creating transactions

anchpop:
This PR has two components:

1) It extends the validators to support gossiping about new operations to include on the chain. It uses a super-simple flood system, where any validator can receive an operation and, if it's new to them, will send it to all the other validator they're connected to (who will then do the same thing etc). Someone could easily DOS the network by sending a bunch of operations really fast, there's no protection against that, and I'm not sure how other blockchains deal with this issue. An easy fix would be to require PoW on each transaction, but that has tradeoffs. 

2) It adds a simple CLI, implemented with Cmdliner, for creating wallets and submitting transactions to the pool of operations.

What still needs to be done, possibly in a future MR:

1) Double check to make sure that validators actually include pending side ops when it's their turn to create blocks.

2) Add tests. The cli currently contains something that should be part of the automated test suite.

3) Move the rest of the command-line tools into this binary, or at least make them use cmdliner so we get nice man pages.

### DIFF
--- a/bin/sidecli.re
+++ b/bin/sidecli.re
@@ -1,0 +1,242 @@
+open Helpers;
+open Node;
+open State;
+open Protocol;
+open Cmdliner;
+
+Printexc.record_backtrace(true);
+
+// Helpers
+
+let read_file = file => {
+  let.await ic = Lwt_io.open_file(~mode=Input, file);
+  let.await lines = Lwt_io.read_lines(ic) |> Lwt_stream.to_list;
+  let.await () = Lwt_io.close(ic);
+  await(lines |> String.concat("\n"));
+};
+let read_file = file => read_file(file) |> Lwt_main.run;
+let read_identity_file = file => {
+  let file_buffer = read_file(file);
+  let json = Yojson.Safe.from_string(file_buffer);
+  identity_of_yojson(json) |> Result.get_ok;
+};
+
+// Todo from Andre: we may have to do peer discovery?
+// I don't know anything about that, except for having played with hyperswarm.
+let validators = () => [
+  read_identity_file("0/identity.json"),
+  // read_identity_file("1/identity.json"),
+  read_identity_file("2/identity.json"),
+  // read_identity_file("identity_3.json"),
+];
+
+let make_filename_from_address = wallet_addr_str => {
+  Printf.sprintf("%s.tzsidewallet", wallet_addr_str);
+};
+
+module Serializable = {
+  [@deriving yojson]
+  type wallet_file = {
+    address: Wallet.t,
+    priv_key: Address.key,
+  };
+};
+
+let exits =
+  Term.default_exits
+  @ Term.[exit_info(1, ~doc="expected failure (might not be a bug)")];
+
+// Commands
+// ========
+
+// Create wallet
+
+let info_create_wallet = {
+  let doc = "Creates a wallet file. The wallet file's filename is its address. The wallet file contains the private key corresponding to that address.";
+  let man = [
+    `S(Manpage.s_bugs),
+    `P("Email bug reports to <contact@marigold.dev>."),
+  ];
+  Term.info("create-wallet", ~version="%‌%VERSION%%", ~doc, ~exits, ~man);
+};
+
+let create_wallet = () => {
+  let (key, wallet) = Wallet.make_wallet();
+
+  let swallet: Serializable.wallet_file = {priv_key: key, address: wallet};
+  let to_yojson = Serializable.wallet_file_to_yojson;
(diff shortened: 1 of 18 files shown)


bin/sidecli.re
```diff
@@ -0,0 +1,243 @@
...
+// I don't know anything about that, except for having played with hyperswarm.
+let validators = () => [
+  read_identity_file("0/identity.json"),
+  // read_identity_file("1/identity.json"),
+  read_identity_file("2/identity.json"),
+  // read_identity_file("identity_3.json"),
+];
+
+let make_filename_from_address = wallet_addr_str => {
+  String.concat("", [wallet_addr_str, ".tzsidewallet"]);
```
EduardoRFS, line 34:
```suggestion
  Printf.sprintf("%s.tzsidewallet", wallet_addr_str);
```

Nice to have specific extensions instead of `.bin`

bin/sidecli.re
```diff
@@ -0,0 +1,243 @@
...
+  await(lines |> String.concat("\n"));
+};
+let read_file = file => read_file(file) |> Lwt_main.run;
+let read_identity_file = file => {
+  let file_buffer = read_file(file);
+  let json = Yojson.Safe.from_string(file_buffer);
+  identity_of_yojson(json) |> Result.get_ok;
+};
+
+// Todo from Andre: we may have to do peer discovery?
```
EduardoRFS, line 24: We do, opened an issue #13

bin/sidecli.re
```diff
@@ -0,0 +1,243 @@
...
+
+  let swallet: Serializable.wallet_file = {priv_key: key, address: wallet};
+  let to_yojson = Serializable.wallet_file_to_yojson;
+  let wallet_json = swallet |> to_yojson;
+
+  let wallet_addr_str = Wallet.address_to_string(wallet);
+
+  let filename = make_filename_from_address(wallet_addr_str);
+
+  Yojson.Safe.to_file(filename, wallet_json);
```
EduardoRFS, line 74: We need to handle exceptions here, or use something from Lwt, probably wrap it and return `Lwt.return(`Error(something))`

bin/sidecli.re
```diff
@@ -0,0 +1,243 @@
...
+
+  Lwt.return(`Ok());
+};
+
+let create_wallet_t =
+  Term.(ret(const(() => create_wallet() |> Lwt_main.run) $ const()));
```
EduardoRFS, line 80: I think we don't need lwt here

bin/sidecli.re
```diff
@@ -0,0 +1,243 @@
...
+  | (_, true, _, _) =>
+    Lwt.return(`Error((false, "Must provide a destination wallet address.")))
+  | (_, _, true, _) =>
+    Lwt.return(
+      `Error((false, "Must provide a transaction amount above zero.")),
+    )
+  | (_, _, _, None) =>
+    Lwt.return(
+      `Error((false, "Transaction destination must be a valid address.")),
+    )
```
EduardoRFS, line 128: Really nice to have nice error messages
cute anime girl happy with good error messages

node/networking.re
```diff
@@ -69,6 +66,13 @@ let broadcast = (endpoint, state, data) =>
        )
      );
 
+let broadcast = (endpoint, state, data) =>
```
EduardoRFS, line 69:
```suggestion
// TODO: this should probably be broadcast_to_validators
let broadcast = (endpoint, state, data) =>
```

protocol/address.re
```diff
@@ -25,6 +25,12 @@ let key_of_yojson =
   | _ => Error("invalid type");
 
 type t = Ed25519.pub_; // TODO: is okay to have this public
+let make_keypair: unit => (key, t) = Ed25519.generate;
```
EduardoRFS, line 28: I think this is dead code.

protocol/wallet.rei
```diff
@@ -1,6 +1,14 @@
...
 type t;
 
 let of_address: Address.t => t;
 let pubkey_matches_wallet: (Address.t, t) => bool;
 let get_pub_key: Address.key => Address.t;
+let make_wallet: unit => (Ed25519.priv, t);
+let make_address: unit => t;
+let address_to_blake: t => BLAKE2B.t;
+let address_of_blake: BLAKE2B.t => t;
+let address_to_string: t => string;
```
EduardoRFS, line 14: \n at the end of the file

EduardoRFS (commented):
Some small nitpicks, hopefully we can merge it soon

EduardoRFS (approved):
Thanks so much for this
anime girl bow while saying thank you

PR #10, opened 2021-06-15
Fix tmux script

anchpop:
We changed the structure of the folders that contain the validator info (`identity.json`, `validators.json`, and sometimes `state.bin`), which broke the tmux script we use for testing. This change fixes the tmux script to assume there are 4 folders (named `0`, `1`, `2`, and `3`) each containing a set of validator info.


EduardoRFS:
anime girl bow to nice work

PR #11, opened 2021-06-15
Make hashing and unhashing consistent

anchpop:
Operations could not be deserialized because the way hashes were calculated when serializing and deserializing was not consistent.  This should fix that issue. Ideally, the code would not be duplicated in the first place, I'm interested in comments on how that might be possible.


EduardoRFS (approved):
anime girl "fixing" the bugs

EduardoRFS:
This is exactly why duplicated code is bad grr

PR #19, opened 2021-07-06
Correct conceptual addresses/wallet mixup

anchpop:
Before, Wallet.t referred to a hash, while Address.t referred to a public key. This is a conceptual mixup - Wallet.t should really be a private key, and Address.t should really be a hash of a public key. This hopefully resolves that issue. 

Draft for now because I'm going to give it another look tomorrow to make sure I didn't miss anything/there's nothing in this PR that should be separated.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -10,6 +10,6 @@ Sometimes you will see it being used even for single cases `v |> f`, this is mor
 
 ### t
 
-A type named t means the main type of a module so, `Validators.t` can be read as `validators` and `Wallet.t` as `wallet`.
+A type named t means the main type of a module so, `Validators.t` can be read as `validators` and `Address.t` as `wallet`.
 
 An identifier named t on a module will always have type t, so under `Validators.re` `let t = empty` is the same as `let validators = empty`
--- a/HACKING.md
+++ b/HACKING.md
@@ -10,6 +10,6 @@ Sometimes you will see it being used even for single cases `v |> f`, this is mor
 
 ### t
 
-A type named t means the main type of a module so, `Validators.t` can be read as `validators` and `Address.t` as `wallet`.
+A type named t means the main type of a module so, `Validators.t` can be read as `validators` and `Wallet.t` as `wallet`.
 
 An identifier named t on a module will always have type t, so under `Validators.re` `let t = empty` is the same as `let validators = empty`
--- a/protocol/wallet.re
+++ b/protocol/wallet.re
@@ -1,15 +1,15 @@
 open Mirage_crypto_ec;
 
 type t = Ed25519.priv;
-type pub_ = Ed25519.pub_;
+type key = Ed25519.pub_;
 
-let wallet_of_privkey = priv => priv;
-let wallet_to_privkey = priv => priv;
+let of_privkey = priv => priv;
+let to_privkey = priv => priv;
 
-let pub_of_Ed25519pub = pub_ => pub_;
-let pub_to_Ed25519pub = pub_ => pub_;
+let key_of_Ed25519pub = pub_ => pub_;
+let key_to_Ed25519pub = pub_ => pub_;
 
-let pubkey_of_wallet = Ed25519.pub_of_priv;
+let to_key = Ed25519.pub_of_priv;
 
 let make_pair = () => Ed25519.generate();
 let make_pubkey = () => make_pair() |> snd;
@@ -35,9 +35,9 @@ let of_yojson =
     | _ => Error("failed to parse")
     }
   | _ => Error("invalid type");
-let pub_to_yojson = t =>
+let key_to_yojson = t =>
   `String(Ed25519.pub_to_cstruct(t) |> Cstruct.to_string |> to_hex);
-let pub_of_yojson =
+let key_of_yojson =
   fun
   | `String(key) =>
     try(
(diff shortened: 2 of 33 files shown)


HACKING.md
```diff
@@ -10,6 +10,6 @@ Sometimes you will see it being used even for single cases `v |> f`, this is mor
 
 ### t
 
-A type named t means the main type of a module so, `Validators.t` can be read as `validators` and `Wallet.t` as `wallet`.
+A type named t means the main type of a module so, `Validators.t` can be read as `validators` and `Address.t` as `wallet`.
```
EduardoRFS, line 13: hmm no?

protocol/wallet.rei
```diff
@@ -1,14 +1,26 @@
 open Mirage_crypto_ec;
-open Helpers;
 
 [@deriving (ord, yojson)]
 type t;
+[@deriving (ord, yojson)]
+type pub_;
+
+let genesis_key: t;
+let genesis_pubkey: pub_;
```
EduardoRFS, line 9: Let's go with `secret` for private keys and `key` to public keys. This is how it's used in michelson and it also makes clear, `secret` is a secret.

protocol/wallet.rei
```diff
@@ -1,14 +1,26 @@
...
+
+let wallet_of_privkey: Ed25519.priv => t;
+let wallet_to_privkey: t => Ed25519.priv;
+let pub_of_Ed25519pub: Ed25519.pub_ => pub_;
+let pub_to_Ed25519pub: pub_ => Ed25519.pub_;
```
EduardoRFS, line 14: If they're alias and you will export the type like this, it's better to just export them with as `t = pri Ed25519.priv`

protocol/wallet.rei
```diff
@@ -1,14 +1,26 @@
...
+
+let pubkey_of_wallet: t => pub_;
+let make_pair: unit => (t, pub_);
+let make_pubkey: unit => pub_;
+
+let to_hex: string => string;
+let of_hex: string => string;
```
EduardoRFS, line 21: Pretty sure of_hex cannot have this type and should not be exported here, currently we avoid exception as we don't have an effect system.

protocol/wallet.rei
```diff
@@ -1,14 +1,26 @@
...
-let get_pub_key: Address.key => Address.t;
-let make_wallet: unit => (Ed25519.priv, t);
-let make_address: unit => t;
-let address_to_blake: t => BLAKE2B.t;
-let address_of_blake: BLAKE2B.t => t;
-let address_to_string: t => string;
+/*
+ let compare_pub: (pub_, pub_) => int;
+ let compare: (t, t) => int;
+ */
```
EduardoRFS, line 26: If it's not used anymore remove it

protocol/wallet.re
```diff
@@ -1,29 +1,66 @@
...
-let address_to_string = wallet =>
-  wallet |> address_to_blake |> BLAKE2B.to_string;
+let of_hex = hex => Hex.to_string(`Hex(hex));
+
+let to_yojson = t =>
+  `String(Ed25519.priv_to_cstruct(t) |> Cstruct.to_string |> to_hex);
+let of_yojson =
+  fun
+  | `String(key) =>
+    // TODO: this raises an exception
```
EduardoRFS, line 28: This is probably an outdated comment

protocol/wallet.rei
```diff
@@ -1,14 +1,26 @@
...
+
+let genesis_key: t;
+let genesis_pubkey: pub_;
+
+let wallet_of_privkey: Ed25519.priv => t;
+let wallet_to_privkey: t => Ed25519.priv;
+let pub_of_Ed25519pub: Ed25519.pub_ => pub_;
+let pub_to_Ed25519pub: pub_ => Ed25519.pub_;
+
+let pubkey_of_wallet: t => pub_;
```
EduardoRFS, line 16: `to_pubkey` or `pubkey`, you will always read it with the module as `Wallet.pubkey` or `Wallet.to_pubkey`

EduardoRFS (changes requested):

protocol/wallet.re
anchpop: Thought so too, just didn't remove it to keep this PR small https://github.com/marigold-dev/sidechain/blob/main/protocol/address.re#L16

protocol/wallet.rei
anchpop: What type does it have here? https://github.com/marigold-dev/sidechain/blob/main/protocol/address.re#L10 my editor says `string => string`. Probably right that it shouldn't be exported

protocol/wallet.rei
anchpop: Should I be using the `:>` syntax>

protocol/wallet.rei
anchpop: Does this mean we should use `Wallet.secret` instead of `Wallet.t`?

EduardoRFS: A wallet is always a secret, saying `Wallet.secret` would be redundant.

protocol/wallet.rei
EduardoRFS: Yes, but we handle the exception internally

at > https://github.com/marigold-dev/sidechain/blob/main/protocol/address.re#L18

d4hines:
Hey @anchpop, can you rebase or close this PR?

PR #29, opened 2021-07-13
Add dockerfile for containerized development

anchpop:
There are two benefits to this. The first is that it makes the build environment more easily reproducible. The second is that it lets me use Github Codespaces to develop. (In fact I'm sending you this PR from codespaces.) You still have to run `esy` and `esy shell` and `esy dune build` yourself, for some reason I couldn't get those working in the dockerfile, but everything else works (syntax highlighting, jump to definition, etc.)

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,16 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.187.0/containers/javascript-node/.devcontainer/base.Dockerfile
+
+# [Choice] Node.js version: 16, 14, 12
+ARG VARIANT="16-buster"
+FROM mcr.microsoft.com/vscode/devcontainers/javascript-node:0-${VARIANT}
+
+# [Optional] Uncomment this section to install additional OS packages.
+RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+    && apt-get -y install --no-install-recommends rsync git m4 build-essential patch unzip wget pkg-config libgmp-dev libev-dev libhidapi-dev libffi-dev opam jq zlib1g-dev inotify-tools
+
+# [Optional] Uncomment if you want to install an additional version of node using nvm
+# ARG EXTRA_NODE_VERSION=10
+# RUN su node -c "source /usr/local/share/nvm/nvm.sh && nvm install ${EXTRA_NODE_VERSION}"
+
+# [Optional] Uncomment if you want to install more global node modules
+RUN su node -c "sudo npm install -g --unsafe-perm esy"
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,27 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.187.0/containers/javascript-node
+{
+	"name": "Node.js",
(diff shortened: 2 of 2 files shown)


EduardoRFS:
I'm not against it, but I would request review from @callistonianembrace if he approves we merge

anchpop:
Actually, no rush, I'm still figuring out how to get docker-in-docker to work, once that does it'll be more worthwhile to merge because you could do things like deploy contracts from the same environment you develop on

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+# [Optional] Uncomment if you want to install an additional version of node using nvm
+# ARG EXTRA_NODE_VERSION=10
+# RUN su node -c "source /usr/local/share/nvm/nvm.sh && nvm install ${EXTRA_NODE_VERSION}"
+
+# Install esy
+RUN su node -c "sudo npm install -g --unsafe-perm esy"
+
+# Install LIGO
+RUN wget https://ligolang.org/deb/ligo.deb && sudo apt install ./ligo.deb
```
callistonianembrace, line 19: Does LIGO provide exact compiler version and a checksum of the .deb file? This could be made more secure.

callistonianembrace: At the very least, it will serve as an integrity check (say, if the there was a network failure during the download?)

callistonianembrace (changes requested):

.devcontainer/Dockerfile
EduardoRFS: Also for ligo there is no need to download a `.deb` we could just download the statically linked binary

anchpop: > Does LIGO provide exact compiler version and a checksum of the .deb file? This could be made more secure.

I don't think they do, but I could generate a checksum myself and check it against that (would need to be changed every time there's a new version of ligo though)

> Also for ligo there is no need to download a .deb we could just download the statically linked binary

Is that better?

EduardoRFS: It is, btw you can also put the binary on the repository, that's okay

EduardoRFS:
@anchpop @callistonianembrace is this frozen?

anchpop:
@EduardoRFS No, it's actually good to merge I think. (Although if nobody uses the devcontainer besides me there may be no advantage in merging it.) Ligo is in the repo and docker-in-docker works so we can run a tezos node

anchpop:
Oh, I accidentally merged @callistonianembrace's branch, let's get that figured out before playing more with this one anyway

anchpop:
@EduardoRFS @callistonianembrace thoughts? I'm happy to maintain this if it's merged.

anchpop:
@d4hines 

Docker used to work but microsoft updated something and my docker-in-docker setup stopped working. Rather than committing the time to getting it working I just disabled docker-in-docker because I didn't think anyone was using it. Also, I don't even use these containers as my main dev environment anymore because I copied Thomas and GA and started using a digitalocean server to develop on and developing on a [remote docker host](https://code.visualstudio.com/remote/advancedcontainers/develop-remote-host) is a much less streamlined experience. 

Sometime down the line I think it would be worth supporting Codespaces as the primary dev environment for Deku (because it would pretty much resolve the "works on my machine" issues we tend to run into, among other things). But I don't think it's suitable just yet because of the restrictions on storage for the machines they provide. Also I'd like to learn nix and use nix to generate the dockerfiles

Issue #30, opened 2021-07-14
use key_hash by default


EduardoRFS:
A user or validator should only need to expose the public key when doing an operation. Currently that's not the case.

That's already the case for users, but not for validators.

## Changes required

- [x] #201
- [x] #202

rosalogia:
Closed by #208

PR #31, opened 2021-07-14
support key_hash on PACK

EduardoRFS:
## Problem

Most of the problem should be described at #30 but this specifically is being added so we can transmit proofs of withdraw to Tezos for tickets communication.

## Solution

This duplicates Tezos encoding logic to sidechain to avoid having a hard dependency on `tezos-crypto` the logic can be tracked from:

https://gitlab.com/tezos/tezos/-/blob/master/src/lib_crypto/ed25519.ml
https://gitlab.com/tezos/tezos/-/blob/master/src/proto_alpha/lib_protocol/script_ir_translator.ml#L514

## Manual Testing

```shell
echo "let pack = (data: key_hash) => Bytes.pack(data)" > test.religo
ligo evaluate-call test.religo pack "(\"tz1LzCSmZHG3jDvqxA8SG8WqbrJ9wz5eUCLC\": key_hash)"
```

## Related Issues

- #34

### DIFF
--- a/tezos_interop/tezos_interop.re
+++ b/tezos_interop/tezos_interop.re
@@ -11,6 +11,16 @@ let rec try_decode_list = (l, string) =>
   | [] => None
   };
 
+let blake2b_20_encoding =
+  Data_encoding.(
+    conv(
+      hash => BLAKE2B_20.to_raw_string(hash) |> Bytes.of_string,
+      // TODO: I don't like this exception below
+      bytes =>
+        Bytes.to_string(bytes) |> BLAKE2B_20.of_raw_string |> Option.get,
+      Fixed.bytes(20),
+    )
+  );
 module Base58 = Base58;
 
 module Ed25519 = {
@@ -40,7 +50,11 @@ module Ed25519 = {
     let hash_key = t =>
       BLAKE2B_20.hash(Ed25519.pub_to_cstruct(t) |> Cstruct.to_string);
 
-    // TODO: encoding
+    let encoding = {
+      let name = "Ed25519.Public_key_hash";
+      // TODO: in tezos this is splitted json is not same as bin
+      Data_encoding.(obj1(req(name, blake2b_20_encoding)));
+    };
 
     let prefix = Base58.Prefix.ed25519_public_key_hash;
     let to_raw = BLAKE2B_20.to_raw_string;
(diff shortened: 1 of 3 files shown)


tezos_interop/tezos_interop.re
```diff
@@ -40,7 +50,11 @@ module Ed25519 = {
     let hash_key = t =>
       BLAKE2B_20.hash(Ed25519.pub_to_cstruct(t) |> Cstruct.to_string);
 
-    // TODO: encoding
+    let encoding = {
+      let name = "Ed25519.Public_key_hash";
+      // TODO: in tezos this is splitted json is not same as bin
```
anchpop, line 55: I'm not sure what this comment means, could it be clarified?

tezos_interop/tezos_interop.re
```diff
@@ -11,6 +11,16 @@ let rec try_decode_list = (l, string) =>
   | [] => None
   };
 
+let blake2b_20_encoding =
```
anchpop, line 14: I was a little confused about this at first, is the size 20 bytes just because that's what's used by tezos here https://gitlab.com/tezos/tezos/-/blob/master/src/lib_crypto/ed25519.ml#L40 ?

tezos_interop/tezos_interop.re
anchpop: BTW not a big deal, but "splitted" isn't an english word, it's just "split"

anchpop (approved):

tezos_interop/tezos_interop.re
EduardoRFS: Yup, key_hash should always be 160bits, that's a tradition since bitcoin. Our default blake2b is the 256bits one

tezos_interop/tezos_interop.re
EduardoRFS: "splitted" seems to be a valid word in english but it is copied from tezos

https://gitlab.com/tezos/tezos/-/blob/master/src/lib_crypto/helpers.ml#L134

https://www.merriam-webster.com/dictionary/splitted

EduardoRFS: @anchpop do you have any suggestion on the comment that should go here?

tezos_interop/tezos_interop.re
callistonianembrace: Perhaps document this in the rei file?
https://github.com/marigold-dev/sidechain/blob/1d9c49c856bb7b7eac5dfa773b23d80fd3e836fb/tezos_interop/tezos_interop.rei#L14

EduardoRFS: This module models tezos data types, so any decision here should be "why is 20 bytes in Tezos?".

PR #32, opened 2021-07-14
support address on tezos_interop

EduardoRFS:
## Depends

- [x] **#31** 

## Problem

This is required so that we can transmit proofs of withdraw to Tezos for tickets communication.

## Solution

This duplicates Tezos logic to sidechain to avoid having a hard dependency on tezos-crypto the logic can be tracked from:

https://gitlab.com/tezos/tezos/-/blob/master/src/proto_alpha/lib_protocol/contract_repr.ml
https://gitlab.com/tezos/tezos/-/blob/master/src/proto_alpha/lib_protocol/script_ir_translator.ml#L466

## Manual Testing

```shell
echo "let pack = (data: address) => Bytes.pack(data)" > test.religo
ligo evaluate-call test.religo pack "(\"tz1LzCSmZHG3jDvqxA8SG8WqbrJ9wz5eUCLC\": address)"
```

## Related Issues

- #34

### DIFF
--- a/tests/test_tezos_interop.re
+++ b/tests/test_tezos_interop.re
@@ -1,5 +1,6 @@
 open Setup;
 open Protocol;
+open Address;
 open Tezos_interop;
 
 // TODO: maybe fuzz testing or any other cool testing magic?
@@ -9,7 +10,7 @@ describe("key", ({test, _}) => {
 
   // TODO: test encoding
 
-  let edpk = Ed25519(Address.genesis_address);
+  let edpk = Ed25519(genesis_address);
   test("to_string", ({expect, _}) => {
     expect.string(to_string(edpk)).toEqual(
       "edpkvDqjL7aXdsXSiK5ChCMAfqaqmCFWCv7DaT3dK1egJt136WBiT6",
@@ -54,7 +55,7 @@ describe("key_hash", ({test, _}) => {
   open Key_hash;
 
   // TODO: proper test of_key
-  let tz1 = of_key(Ed25519(Address.genesis_address));
+  let tz1 = of_key(Ed25519(genesis_address));
   test("to_string", ({expect, _}) => {
     expect.string(to_string(tz1)).toEqual(
       "tz1LzCSmZHG3jDvqxA8SG8WqbrJ9wz5eUCLC",
@@ -80,7 +81,7 @@ describe("key_hash", ({test, _}) => {
 describe("secret", ({test, _}) => {
   open Secret;
 
-  let edsk = Ed25519(Address.genesis_key);
+  let edsk = Ed25519(genesis_key);
   test("to_string", ({expect, _}) => {
     expect.string(to_string(edsk)).toEqual(
       "edsk4bfbFdb4s2BdkW3ipfB23i9u82fgji6KT3oj2SCWTeHUthbSVd",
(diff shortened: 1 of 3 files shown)


tests/test_tezos_interop.re
```diff
@@ -216,9 +245,19 @@ describe("pack", ({test, _}) => {
...
   );
+  test(
+    "key_hash(\"tz1LzCSmZHG3jDvqxA8SG8WqbrJ9wz5eUCLC\")",
+    key_hash(Key_hash.of_key(Ed25519(genesis_address))),
+    "050a00000015000ec89608700c0414159d93552ef9361cea96da13",
+  );
+  test(
+    "address(\"tz1LzCSmZHG3jDvqxA8SG8WqbrJ9wz5eUCLC\")",
+    address(Implicit(Key_hash.of_key(Ed25519(genesis_address)))),
+    "050a0000001600000ec89608700c0414159d93552ef9361cea96da13",
```
anchpop, line 259: Might want to include a comment referencing https://ide.ligolang.org/p/4fLtnmyjluBgNoZ-TvBu7A

tezos_interop/tezos_interop.re
```diff
@@ -131,6 +145,27 @@ module Key_hash = {
   type t =
     | Ed25519(Ed25519.Hash.t);
 
+  let name = "Signature.Public_key_hash";
+
+  let title = "A Ed25519, Secp256k1, or P256 public key hash";
```
anchpop, line 150: maybe include a comment indicating that this is copied from https://gitlab.com/tezos/tezos/-/blob/master/src/lib_crypto/signature.ml#L58 ?

anchpop (approved):

tezos_interop/tezos_interop.re
```diff
@@ -11,6 +11,16 @@ let rec try_decode_list = (l, string) =>
   | [] => None
   };
 
+let blake2b_20_encoding =
+  Data_encoding.(
+    conv(
+      hash => BLAKE2B_20.to_raw_string(hash) |> Bytes.of_string,
+      // TODO: I don't like this exception below
```
callistonianembrace, line 18: What exception?

EduardoRFS: Option.get in the line below

tezos_interop/tezos_interop.re
EduardoRFS: This is not from this PR, it was merged on #31

tests/test_tezos_interop.re
EduardoRFS: #40 opened an issue on this, I think it's important

tezos_interop/tezos_interop.re
callistonianembrace: Can you elaborate your comment (in the file)?

tests/test_tezos_interop.re
callistonianembrace: @anchpop Are you asking how the value expected key_hash was arrived at?

anchpop: It was arrived at using the ligo code @EduardoRFS wrote in the "manual testing" section, but I thought it was unclear in the code why it was being tested against that specific value

EduardoRFS: Closing this, if you want to discuss please use the issue

PR #42, opened 2021-07-19
CLI: Fix doc strings for create-transaction

callistonianembrace:
Before

```
ARGUMENTS
       MSG (required)
           The message to print.

       MSG (required)
           The message to print.

       MSG (required)
           The message to print.
```

Now,
```
SYNOPSIS
       sidecli create-transaction [OPTION]... sender receiver amount

ARGUMENTS
       amount (required)
           The amount to be transacted.

       receiver (required)
           The receiving address.

       sender (required)
           The sending address, or a path to a wallet. If a bare sending
           address is provided, the corresponding wallet is assumed to be in
           the working directory.
```

### DIFF
--- a/bin/sidecli.re
+++ b/bin/sidecli.re
@@ -183,33 +183,30 @@ let create_transaction = {
   let address_from = {
     let doc = "The sending address, or a path to a wallet. If a bare sending address is provided, the corresponding wallet is assumed to be in the working directory.";
     let env = Arg.env_var("SENDER", ~doc);
-    let doc = "The message to print.";
     Arg.(
       required
       & pos(0, some(wallet), None)
-      & info([], ~env, ~docv="MSG", ~doc)
+      & info([], ~env, ~docv="sender", ~doc)
     );
   };
 
   let address_to = {
     let doc = "The receiving address.";
     let env = Arg.env_var("RECEIVER", ~doc);
-    let doc = "The message to print.";
     Arg.(
       required
       & pos(1, some(address), None)
-      & info([], ~env, ~docv="MSG", ~doc)
+      & info([], ~env, ~docv="receiver", ~doc)
     );
   };
 
   let amount = {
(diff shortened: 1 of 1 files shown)


callistonianembrace:
Depends on https://github.com/marigold-dev/sidechain/pull/15

anchpop (approved):

anchpop:
Thanks, small change but big improvement in usability (my fault for forgetting to do this originally)

EduardoRFS:
cute small anime girl happy with small PR

bin/sidecli.re
```diff
@@ -183,33 +183,30 @@ let create_transaction = {
   let address_from = {
     let doc = "The sending address, or a path to a wallet. If a bare sending address is provided, the corresponding wallet is assumed to be in the working directory.";
     let env = Arg.env_var("SENDER", ~doc);
-    let doc = "The message to print.";
```
callistonianembrace, line 186: @EduardoRFS I wonder why Dune didn't complain of shadowing here. Are we missing flags? Is it because of `-p` in the esy build command?

callistonianembrace: I mean, of course, before this PR

callistonianembrace: (please free to mark this thread resolved once you see it. Not a blocker)

anchpop: I think (but I'm not sure) that -p causes dune to build only what's needed for the packages specified in sidechain.install, and sidecli and genesis_helper should be in there, so I don't think that should be the problem. I also wonder if merging #39 would affect this.

EduardoRFS: @callistonianembrace because it is used, we're not doing anything wrong

EduardoRFS: resolving this, as it doesn't seems related to the PR

EduardoRFS (approved):
anime girl happy with docs

PR #44, opened 2021-07-22
Merge sidecli with genesis helpers

callistonianembrace:
Draft because I need to look into

```
 $cur__target_dir/default/bin/sidecli.exe inject-genesis                                                                                                   
0                                                                                                                                                                                                                                      
block_hash: 4c3505a532b6cf21852c6d71b6c2be82e41d080fa431053ce44e3b94cb720047, state_hash: d35106d99c319fdb19fda58c316f967222f8edd96489a858a02ec4a897ed14e9, block_height: 1, validators: edpkuKhMZk3hsNSKZqQ4tCqWGfLsUot31KjbcUmxvKxwRc
UTH6LxFt,edpkv2Jr7qCp8FgRTapzRH6SyRNomkwEY154zXvKomGe5FRaERnVFT,edpktvWinmLsmCL7npxwKMeDSMzQZQLaX5Uv5Moie7zdmFb2MJPh36,edpktfHqDKsSoh5zPAQi3THLKxrakeSHzRfRVuLdLP34pCu5krf1Uosidecli: internal error, uncaught exception:
         Node.Networking.Error_status                                                                                                                                                                                                  
         Raised at Lwt.Miscellaneous.poll in file "src/core/lwt.ml", line 3072, characters 20-29
         Called from Lwt_main.run.run_loop in file "src/unix/lwt_main.ml", line 31, characters 10-20                                                                                                                                   
         Called from Lwt_main.run in file "src/unix/lwt_main.ml", line 118, characters 8-13                                                                                                                                            
         Re-raised at Lwt_main.run in file "src/unix/lwt_main.ml", line 124, characters 4-13
         Called from Cmdliner_term.app.(fun) in file "cmdliner_term.ml", line 25, characters 19-24                                                                                                                                     
         Called from Cmdliner.Term.run in file "cmdliner.ml", line 117, characters 32-39 
  ```

### CODE REVIEW

**Comment by callistonianembrace** [comment]
False alarm - the error was due to repeated genesis block injections. Clearing the `state.bin`s made the trace disappear. We need better error logs

**Comment by callistonianembrace** [comment]
For this PR, the scope has only been to merge the two CLIs. More improvements coming up in future PRs

File: bin/sidecli.re [hunk 1]
```diff
@@ -15,13 +15,38 @@ let read_file = file => {
...
+};
+
+let read_validators = file => {
+  let.await file_buffer = read_file(file);
+  await(
+    try({
+      let json = Yojson.Safe.from_string(file_buffer);
+      module T = {
+        [@deriving of_yojson]
+        type t = {
```
anchpop, line 38: Not blocking, because it's not any worse than it was before, but this is duplicated [here](https://github.com/marigold-dev/sidechain/pull/44/files#diff-f199cc2f35dec0a8571164460a4a4bfa6ca3100739f3f5b69314008de807ccfcR379) and we should remember to merge them

anchpop (approved):
Thanks for doing this haha, I haven't tested it locally but it looks good to me.

bin/sidecli.re
```diff
@@ -342,6 +553,8 @@ let () = {
       (create_transaction, info_create_transaction),
       (sign_block, info_sign_block),
       (produce_block, info_produce_block),
+      (gen_credentials, info_gen_credentials),
+      (inject_genesis, info_inject_genesis),
```
EduardoRFS, line 557: inject_genesis is not needed, it should be attainable with produce_block + sign_block

EduardoRFS (changes requested):

bin/sidecli.re
anchpop: I feel like it's a useful command just for convenience

EduardoRFS: Inject_genesis should almost never happen, we can have an alias, but not duplicate the logic. And inject_genesis also only makes sense if you control all the keys, which we will not in the future

anchpop: Made an issue for it: #45

bin/sidecli.re
EduardoRFS: @anchpop open an issue

bin/sidecli.re
callistonianembrace: How can the genesis block be created with `produce-block` command? `produce-block` only takes path to the `state.bin` file right now.

EduardoRFS: @callistonianembrace and it will produce a block, you can run a node to get the state.bin, will move this to an issue to document on how this can be done

#59

bin/sidecli.re
EduardoRFS: resolving the conversation to merge, but with a follow up #60

EduardoRFS (approved):
It's definitely an improvement and it highlighs the problem, so LET'S GO

anime girl concerned with duplicated logic

anchpop (approved):

PR #46, opened 2021-07-23
Toplevel for REPL driven development

callistonianembrace:
`rtop` doesn't understand OCaml syntax emitted by `dune top` and needs a workaround - load the `cma` files first from the `use_output`, then activate the reason parsers in the REPL.

We could take this up in the Reason community and see if this can be fixed natively there.

### DIFF
--- a/esy.lock/index.json
+++ b/esy.lock/index.json
@@ -1,5 +1,5 @@
 {
-  "checksum": "41a3676ed7f27e8413521ecd8f207b69",
+  "checksum": "7a1c8c30b81f193463d11335669bffac",
   "root": "sidechain@link-dev:./package.json",
   "node": {
     "yocto-queue@0.1.0@d41d8cd9": {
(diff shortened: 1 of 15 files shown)


callistonianembrace:
To test the REPL - ` esy b dune exec ./tools/dtop.exe`

anchpop:
I get this output when running that command:

```
node ➜ /workspaces/sidechain (callistonianembrace/add-rtop) $ esy b dune exec ./tools/dtop.exe
Fatal error: exception Unix.Unix_error(Unix.ENOENT, "open", "/home/node/.config/.lambda-term-inputrc")
Called from unknown location
Called from unknown location
Called from unknown location
Called from unknown location
```

Do I need to install rtop? If so, can it be included in the package.json so it's installed automatically after running `esy install`?

callistonianembrace:
@anchpop Can you `touch /home/node/.config/.lambda-term-inputrc` the file?

anchpop:
@callistonianembrace I also had to create `/home/node/.rtop-history` and `/home/node/.utoprc` in addition to `/home/node/.config/.lambda-term-inputrc`, but now it seems to work. Should I include commands to touch those in my docker configuration, or is there a different recommended way to set up an environment for rtop?

anchpop (approved):

callistonianembrace:
>  Should I include commands to touch those in my docker configuration, or is there a different recommended way to set up an environment for rtop?

@anchpop Yes, please touch those files.

EduardoRFS:
I don't like this, utop has a reason mode right? Why can't we just use it instead?

callistonianembrace:
@EduardoRFS I'm not sure what you're referring to. How do I try out this reason mode you're talking about?

EduardoRFS:
@callistonianembrace I don't remember that much, but isn't this https://github.com/ocaml-community/utop/blob/bac3946079a98df00f31656dc10c6c9f1a8bc422/src/top/utop.el#L264 ?

callistonianembrace:
@EduardoRFS That's an emacs plugin - utop minor mode works with reason major mode.

EduardoRFS:
Opened a follow up #118

EduardoRFS:
anime girl saying nice to developer experience being improved

PR #47, opened 2021-07-23
ensure that side chain operations are checked

EduardoRFS:
## Problem

This code was made in a couple of hours and was particularly brittle, and having two kinds of operations one checked and one not checked.

## This PR

The goal here is to simplify it and enforce the invariant that all side operations **must be signed by the source** of the money

### DIFF
--- a/protocol/operation.re
+++ b/protocol/operation.re
@@ -17,6 +17,7 @@ module Side_chain = {
   [@deriving (ord, yojson)]
   type t = {
     hash: BLAKE2B.t,
+    signature: Signature.t,
     nonce: int32,
     block_height: int64,
     source: Wallet.t,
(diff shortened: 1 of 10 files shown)


protocol/operation.re
```diff
@@ -38,33 +39,30 @@ module Side_chain = {
...
+    let.ok () =
+      verify(~hash, ~nonce, ~block_height, ~source, ~amount, ~kind)
+        ? Ok() : Error("Side operation invalid hash");
+    let.ok () =
+      Signature.verify(~signature, hash)
+      && Wallet.pubkey_matches_wallet(
+           Signature.public_key(signature),
+           source,
+         )
+        ? Ok() : Error("Side operation invalid signature");
```
anchpop, line 53: Should these be string errors or would variants be better?

anchpop (approved):

protocol/operation.re
EduardoRFS: variants would change the type and bind would fail as one would be `result('a, string)` and the other `result('b, some_other_type)`

EduardoRFS: resolving this to merge

PR #48, opened 2021-07-23
add tezos_hash to all main chain operations

EduardoRFS:
## Problem

Currently main chain operations have no identifier shared with tezos, so we cannot ensure that an operation already happened.

## This PR

It is a temporary fix, it adds the tezos_hash to allow main chain operations to be properly indexed.

### DIFF
--- a/protocol/operation.re
+++ b/protocol/operation.re
@@ -2,11 +2,21 @@ open Helpers;
 
 module Main_chain = {
   [@deriving (ord, yojson)]
-  type t =
+  type kind =
     // TODO: can a validator uses the same key in different nodes?
     // If so the ordering in the list must never use the same key two times in sequence
     | Add_validator(Validators.validator)
     | Remove_validator(Validators.validator);
+  [@deriving yojson]
+  type t = {
+    tezos_hash: BLAKE2B.t,
+    kind,
+  };
+  let compare = (a, b) => BLAKE2B.compare(a.tezos_hash, b.tezos_hash);
+
+  let make = (~tezos_hash, ~kind) => {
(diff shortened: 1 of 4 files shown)


protocol/operation.re
```diff
@@ -2,11 +2,21 @@ open Helpers;
...
     | Remove_validator(Validators.validator);
+  [@deriving yojson]
+  type t = {
+    tezos_hash: BLAKE2B.t,
+    kind,
+  };
+  let compare = (a, b) => BLAKE2B.compare(a.tezos_hash, b.tezos_hash);
+
+  let make = (~tezos_hash, ~kind) => {
+    {tezos_hash, kind};
```
anchpop, line 18: Should we be checking that hash actually points to a valid operation on the mainchain?

anchpop (approved):

protocol/operation.re
EduardoRFS: #55 it's more complex than that, so I opened an issue

callistonianembrace:
> It is a temporary fix

What do you think a better fix would be?

protocol/operation.re
callistonianembrace: Could you please elaborate #55 before we merge this PR?

callistonianembrace (changes requested):

callistonianembrace:
Except for that more descriptive issue, everything else looks fine to me.

protocol/operation.re
EduardoRFS: @callistonianembrace added a pointer to this PR

callistonianembrace (approved):

PR #50, opened 2021-07-23
improve validators module tests

EduardoRFS:
## This PR

Adds two additional test cases for the Validators module and some comments.

### DIFF
--- a/tests/test_validators.re
+++ b/tests/test_validators.re
@@ -56,6 +56,12 @@ describe("validators", ({test, _}) => {
       let t = t |> remove(b);
       expect.equal(t, empty);
     };
+    {
+      // remove unknown
+      let (t, _, _) = setup_two();
+      let unknown = make_validator();
+      expect.equal(remove(unknown, t), t);
+    };
     ();
   });
   test("after_current", ({expect, _}) => {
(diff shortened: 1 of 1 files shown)


anchpop (approved):

callistonianembrace (commented):
Can we have more descriptive test descriptions, atleast for the touched tests?

tests/test_validators.re
```diff
@@ -90,6 +96,12 @@ describe("validators", ({test, _}) => {
 
     // ensure this is a noop
     expect.equal(update_current(b.address, t), t);
+
```
callistonianembrace, line 99: Can we have more descriptive test descriptions, atleast for the touched tests?

callistonianembrace (changes requested):

tests/test_validators.re
callistonianembrace: For instance, this test could read "Only validators present in the authorised list can be made current validator". This could be documented in the doc string at the `update_current` function too.

EduardoRFS: openen an issue as a follow up #56

callistonianembrace (approved):

PR #51, opened 2021-07-23
add incremental patricia tree

EduardoRFS:
## Problem

To validate data on Tezos main chain during withdraws the user needs a way to provide a proof that he did burn the money on the sidechain.

## Solution

We use a binary radix tree, where the key is always an incremental int, with this every money burnt will receive an id back that can be used to construct the proof.

## Related issues

- #34


anchpop (approved):

anchpop:
Not guaranteeing this code is correct but I looked it over and it looks good to me

PR #58, opened 2021-07-26
Uses esy shell in .tmuxinator config

callistonianembrace:
Use pre_window field in the config to reliably set up the environment
in which cur__* variables are always present in all tmux panes.


anchpop (approved):
Haven't tested it locally, but lgtm!

EduardoRFS (approved):
anime girl happy with workflow improvements

PR #63, opened 2021-07-26
make a library for files

EduardoRFS:
## Depends

- [x] #44 

## Problem

We've been duplicating logic for a while to handle files, namely `identity.json` and `validators.json`, this is getting quite messy especially when adding new files.

## Solution

I made a library for file formats which currently sits as part of `bin`, this is not ideal but it solves the duplicated code problem without too many changes on the project structure.

### DIFF
--- a/bin/dune
+++ b/bin/dune
@@ -7,7 +7,13 @@
 
 (executable
  (name sidecli)
- (libraries node mirage-crypto-ec mirage-crypto-rng.unix helpers cmdliner)
+ (libraries
+  node
+  files
+  mirage-crypto-ec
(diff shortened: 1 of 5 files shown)


callistonianembrace (commented):
Opening an issue for error handling. Error logs could be more verbose and preferably at the site that has more context - either context can be built like `RunAsync.t` at esy (which is more scalable and composable) or we could simply guard more IO calls. 

Otherwise LGTM

Follow up issue here - https://github.com/marigold-dev/sidechain/issues/64

callistonianembrace (approved):

anchpop (approved):

PR #67, opened 2021-07-27
add MIT LICENSE

### LINKED ISSUE #66: Add license
The project should be under MIT but currently doesn't have a licence. I've borrowed the contract code for a ligo playground so it would be very good if we could clarify it

EduardoRFS:
## Problem

We're lacking a license on the project, this means that while actually we're open source no one is allowed to use this code without our permission. But this project should be open source to allow others to use it freely.

## Solution

As discussed with @Galfour in private everything at marigold is by default MIT unless explicitly discussed otherwise, so I'm just adding it here.

## Related Issues

- closes #66

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2021 Marigold <contact@marigold.dev>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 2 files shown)


callistonianembrace (approved):

callistonianembrace:
Are we ever going to distribute this via NPM? Should considering adding the license to the package.json too>

d4hines (approved):

EduardoRFS:
@callistonianembrace that's a good idea, added another commit on it

anchpop (approved):

Galfour (approved):

PR #68, opened 2021-07-27
Set block level in create-transaction command

callistonianembrace:
**Problem**
To prevent replay attacks, `create-transaction` command must stamp the transaction requests to the node with the block level at which the transaction is expected to be baked.

**Solution**
This PR adds a `/block-level` endpoint to query the current block level and use it while signing the transaction operation.

### DIFF
--- a/bin/http_server.re
+++ b/bin/http_server.re
@@ -46,7 +46,12 @@ let handle_request =
       | Ok(response) =>
         let response = E.response_to_yojson(response);
         await(Response.of_json(~status=`OK, response));
-      | Error(_err) =>
+      | Error(err) =>
+        switch(err) {
+        | `Not_a_json => print_endline("Invalid json");
+        | `Not_a_valid_request(err) => print_endline("Invalid request:"  ++ err);
+          | _ => print_endline("unhandled");
+        }
         await(Response.make(~status=`Internal_server_error, ()))
       };
     },
@@ -95,6 +100,12 @@ let handle_block_by_hash =
       Ok(block);
     },
   );
+
+let handle_block_level =
+  handle_request((module Block_level), (_update_state, _request) => {
+    Ok({ level : Flows.find_block_level(Server.get_state()) })
+  });
+
 let handle_protocol_snapshot =
   handle_request(
     (module Protocol_snapshot),
@@ -218,6 +229,7 @@ let () = Node.Server.start(~initial=node);
 let _server =
   App.empty
   |> App.port(Node.Server.get_port() |> Option.get)
+  |> handle_block_level
   |> handle_received_block_and_signature
   |> handle_received_signature
   |> handle_block_by_hash
--- a/bin/http_server.re
+++ b/bin/http_server.re
@@ -47,12 +47,13 @@ let handle_request =
         let response = E.response_to_yojson(response);
         await(Response.of_json(~status=`OK, response));
       | Error(err) =>
-        switch(err) {
-        | `Not_a_json => print_endline("Invalid json");
-        | `Not_a_valid_request(err) => print_endline("Invalid request:"  ++ err);
-          | _ => print_endline("unhandled");
-        }
-        await(Response.make(~status=`Internal_server_error, ()))
+        switch (err) {
+        | `Not_a_json => print_endline("Invalid json")
+        | `Not_a_valid_request(err) =>
+          print_endline("Invalid request:" ++ err)
+        | _ => print_endline("unhandled")
+        };
+        await(Response.make(~status=`Internal_server_error, ()));
       };
     },
   );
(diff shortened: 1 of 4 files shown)


callistonianembrace:
Untested for now, but would love a code review.

bin/http_server.re
```diff
@@ -46,8 +46,14 @@ let handle_request =
...
         await(Response.of_json(~status=`OK, response));
-      | Error(_err) =>
-        await(Response.make(~status=`Internal_server_error, ()))
+      | Error(err) =>
+        switch (err) {
+        | `Not_a_json => print_endline("Invalid json")
+        | `Not_a_valid_request(err) =>
+          print_endline("Invalid request:" ++ err)
+        | _ => print_endline("unhandled")
+        };
```
EduardoRFS, line 55: That's gonna be helpful

bin/sidecli.re
```diff
@@ -176,14 +176,30 @@ let info_create_transaction = {
...
+    |> Block_level.response_to_yojson
+    |> Yojson.Safe.to_basic;
+  let block_level =
+    switch (Yojson.Basic.Util.(member("level", block_level_yojson))) {
+    | `Int(block_level) => block_level
+    | _ =>
+      failwith(
+        "JSON returned from /block-level was expected to have field block_level",
+      )
+    };
```
EduardoRFS, line 194:
```suggestion
  let.await block_level_response = request_block_level((), validator_uri);
  let block_level = block_level_response.level;
```

block_level_response already exposes a level field

bin/sidecli.re
```diff
@@ -176,14 +176,30 @@ let info_create_transaction = {
...
   let transaction = {
     let.ok wallet = load_wallet_file(sender_wallet_file);
 
     Ok(
       Operation.Side_chain.sign(
         ~secret=wallet.priv_key,
         ~nonce=0l,
-        ~block_height=0L,
+        ~block_height=Int64.of_int(block_level + 1),
```
EduardoRFS, line 202: It needs to be bigger than +1, probably something like 100 is a good idea, this is the maximum block_height that an operation can be included

EduardoRFS (changes requested):
Two small changes

anchpop:
I don't understand how this prevents replay attacks. If I submit a transaction with a block_height of `block_level + 1`, can someone not copy the transaction 3 times and put it in that block? Or will only transaction per sender per block be allowed? 

Also, is the block_height an upper limit, or must it match exactly?

EduardoRFS:
@anchpop no, there is the hash of the operation, what prevents replay attack is the set of included operations. But because we cannot hold an infinite number of operations we have a block_height that cuts the number of blocks we need to hold an operation hash.

So the PR description is a bit off, but it's in the right direction

EduardoRFS:
rebased the PR

bin/sidecli.re
```diff
@@ -118,19 +118,34 @@ let info_create_transaction = {
...
+        "JSON returned from /block-level was expected to have field block_level",
+      )
+    };
   let.await wallet = Files.Wallet.read(~file=sender_wallet_file);
   let transaction =
     Operation.Side_chain.sign(
       ~secret=wallet.priv_key,
       ~nonce=0l,
-      ~block_height=0L,
+      ~block_height=Int64.of_int(block_level + 1),
```
EduardoRFS, line 142: It needs to be bigger than +1, probably something like 100 is a good idea, this is the maximum block_height that an operation can be included

anchpop (approved):
Oh okay, with Edu's suggestions this looks good then

bin/http_server.re
callistonianembrace: Cool. Shall I resolve this?

bin/sidecli.re
callistonianembrace: Rename the param to `max_block_height` ?

bin/http_server.re
callistonianembrace: GH is showing this comment as 'Requested Changes'. Not sure how to resolve this.

bin/sidecli.re
EduardoRFS: No actually I'm wrong this operation is done on the node itself, so this should be just `block_level`

EduardoRFS: https://github.com/marigold-dev/sidechain/blob/main/protocol/protocol.re#L46

bin/sidecli.re
```diff
@@ -118,19 +118,22 @@ let info_create_transaction = {
...
+  let validator_uri = List.hd(validators_uris);
+  let.await block_level_response = request_block_level((), validator_uri);
+  let block_level = block_level_response.level;
   let.await wallet = Files.Wallet.read(~file=sender_wallet_file);
   let transaction =
     Operation.Side_chain.sign(
       ~secret=wallet.priv_key,
       ~nonce=0l,
-      ~block_height=0L,
+      ~block_height=Int64.(add(block_level, of_int(100))),
```
EduardoRFS, line 131: Probably you misread, https://github.com/marigold-dev/sidechain/pull/68#discussion_r677777268

This should be just

```suggestion
      ~block_height=block_level,
```

EduardoRFS (changes requested):

EduardoRFS (approved):
anime girl happy with PR that fix important bug

anchpop (approved):

PR #74, opened 2021-07-27
add tickets to operations

EduardoRFS:
## Problem

Users should be able to operate on tickets.

## Solution

This is solved through adding tickets to the ledger and to Transaction operation, because of this essentially the key for a ledger is a tuple between the `ticket` and wallet `address`.

## Related Issues

- #34

### DIFF
--- a/protocol/ticket.re
+++ b/protocol/ticket.re
@@ -1,0 +1,28 @@
+open Tezos_interop;
+
+// TODO: I feel uneasy about this module
+module Address = {
+  include Address;
+  let with_yojson_string = (name, of_string, to_string) =>
+    Helpers.with_yojson_string(
+      string =>
+        of_string(string) |> Option.to_result(~none="invalid " ++ name),
+      to_string,
+    );
+  let (of_yojson, to_yojson) =
+    with_yojson_string("address", of_string, to_string);
+};
+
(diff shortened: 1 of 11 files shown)


anchpop (approved):

protocol/ticket.re
```diff
@@ -0,0 +1,28 @@
+open Tezos_interop;
+
+// TODO: I feel uneasy about this module
```
callistonianembrace, line 3: Why do you feel uneasy about this module?

callistonianembrace: I'm also reminded of functors like `Set_with_yojson_make` and `Map_with_yojson_make` - can they be unified?

callistonianembrace (approved):

protocol/ticket.re
callistonianembrace: (feel free to resolve)

EduardoRFS: A lot of duplicated code and re exporting a type from Tezos_interop, it's probably a bad abstraction. Also nope, `Set_with_yojson_make` is completely different from `Map_with_yojson_make`.

EduardoRFS: It's already slightly improved on a future PR tho

PR #79, opened 2021-07-28
add types for hashes in consensus contract

EduardoRFS:
## Depends

- [x] #76 

## Problem

Currently hashes are treated as sequence of bytes, but there is also some other things that are sequences of bytes like the result of `Bytes.pack`, it would be cool if we ensured that a hash is not just any sequence of bytes.

## Solution

Sadly Ligo doesn't have nominal typing so for now I'm just adding an alias to makes it clear for us while reading the consensus contract.

### DIFF
--- a/tezos_interop/consensus.mligo
+++ b/tezos_interop/consensus.mligo
@@ -1,3 +1,5 @@
+type blake2b = bytes
+
 (* store hash *)
 type validator = key
 type validators = validator list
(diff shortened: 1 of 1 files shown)


callistonianembrace (approved):
Diff looks safe to me, but I have a very shallow understanding of the contracts at this point.

anchpop (approved):

PR #81, opened 2021-07-28
Don't repeat man information all over the place in sidecli

anchpop:
## Problem

Currently, the sidecli `man` information is duplicated in many places. 

## Solution

We should have one man page "template", which is really just a list of man page items, and customize it as needed by prepending more information to the list. Since every man page is exactly the same right now, it's easy to move the repeated `man` definition to the top.

## Further work

We should probably write more detailed man page information for each option at some point. I've created #80 to track that.

### DIFF
--- a/bin/sidecli.re
+++ b/bin/sidecli.re
@@ -6,6 +6,11 @@ open Cmdliner;
 
 Printexc.record_backtrace(true);
 
+let man = [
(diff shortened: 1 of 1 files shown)


callistonianembrace (approved):

EduardoRFS (approved):
anime girl saying nice to simple improvement

PR #83, opened 2021-07-28
add deposit on consensus

EduardoRFS:
## Depends

- [x] #78 

## Problem

As per #34 we need to be able to move tickets to the sidechain.

## Solution

Add a deposit entrypoint on the contract, this should be detected by the node and calls `Ledger.deposit` when the operation is added to a block.

## Related

- #34

### DIFF
--- a/tezos_interop/consensus.mligo
+++ b/tezos_interop/consensus.mligo
@@ -116,15 +116,45 @@ let root_hash_main
       current_validators = validators;
     }
 
+(* vault contract *)
+type vault_storage = (address * bytes, bytes ticket) big_map
+type vault_deposit = {
+  ticket: bytes ticket;
+  (* WARNING: deposit address should be a valid sidechain address *)
+  address: address
+}
+let vault_deposit (deposit: vault_deposit) (vault: vault_storage) =
+  let (content, ticket) =  Tezos.read_ticket deposit.ticket in
+  let (ticketer, (data, _)) = content in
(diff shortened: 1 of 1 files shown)


tezos_interop/consensus.mligo
```diff
@@ -116,15 +116,45 @@ let root_hash_main
...
 
+(* vault contract *)
+type vault_storage = (address * bytes, bytes ticket) big_map
+type vault_deposit = {
+  ticket: bytes ticket;
+  (* WARNING: deposit address should be a valid sidechain address *)
+  address: address
+}
+let vault_deposit (deposit: vault_deposit) (vault: vault_storage) =
+  let (content, ticket) =  Tezos.read_ticket deposit.ticket in
```
anchpop, line 127:
```suggestion
  let (content, ticket) = Tezos.read_ticket deposit.ticket in
```

anchpop (approved):

tezos_interop/consensus.mligo
EduardoRFS: thanks

callistonianembrace (approved):
I have a very shallow understanding of these contracts. I need to study them better before I can provide meaningful contract. Approving to unblock merging.

anchpop:
@callistonianembrace Me too, let me know if you want to go over them together (or if you go over them with @EduardoRFS, let me know so I can join in)

PR #85, opened 2021-07-28
add handles hash to root hash

EduardoRFS:
## Depends

- [x] #83 

## Problem

To do withdraws from sidechain to Tezos we need to somehow provide proofs that the ticket was burned on the sidechain.

## Solution

The goal here is to add a hash on every root_hash update for `handles_hash` which is the root hash 
 of a patricia tree from an `id -> data`, this will allow a user to construct a proof when burning tickets.

Note that this is currently not implemented as the sidechain is committing always the same handles_hash and the goal of this PR is only to introduce it in the contract.

## Related

- #34

### DIFF
--- a/tests/test_tezos_interop.re
+++ b/tests/test_tezos_interop.re
@@ -372,7 +372,7 @@ describe("consensus", ({test, _}) => {
       );
     let hash = BLAKE2B.to_string(hash);
     expect.string(hash).toEqual(
-      "58b39faab167d45a69a0cf126bade7fd23932b2c48684adfbd164281ca2cc5ff",
+      "23cdb9e9ffef6dd17553b622af0c043f544daa18430dff295d04a9d46b3e5267",
     );
   });
 });
--- a/tezos_interop/consensus.mligo
+++ b/tezos_interop/consensus.mligo
@@ -11,6 +11,7 @@ type root_hash_storage = {
   current_block_hash: blake2b;
   current_block_height: int;
   current_state_hash: blake2b;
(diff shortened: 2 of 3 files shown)


EduardoRFS:
@anchpop this PR implements it https://github.com/marigold-dev/sidechain/pull/85/commits/15063adadc398cc789bfc55acf4fe6b19444ebe5

anchpop:
@EduardoRFS oh, that code corresponds to [this](https://github.com/marigold-dev/sidechain/commit/15063adadc398cc789bfc55acf4fe6b19444ebe5#diff-3d45d01dfacf702d630feafd03bf8f096009d6779a783e3a16a6a90784591827R116-R122)? (Btw if so I'm uncomfortable with how loosely coupled these are but I don't know if there's anything we can do about it.)

EduardoRFS:
@anchpop nope, it corresponds to this https://github.com/marigold-dev/sidechain/pull/85/commits/15063adadc398cc789bfc55acf4fe6b19444ebe5#diff-3d45d01dfacf702d630feafd03bf8f096009d6779a783e3a16a6a90784591827R60

You can see that there is a pack of it and we reimplement this pack on the sidechain.

> I'm uncomfortable with how loosely coupled these are
Same I would love to generate this from the contract itself, opened an issue #86

callistonianembrace (approved):
I don't understand this Pull Request completely, but wish to unblock it.

EduardoRFS:
@callistonianembrace what did you don't understand here?

PR #88, opened 2021-07-28
add withdraw consensus contract

EduardoRFS:
## Depends

- [x] #85

## Problem

To achieve #34 we need a way to move money outside of the sidechain and have a proof to it so that Tezos can verify it.

## Solution

As #85 we have the handles_hash here we add them for a set of handle_hash and when the user tries to withdraw it we verify the proof by assuming the hash represents a patricia tree where the key is always an incremental int.

After doing a withdraw the handle is added to a set of handles_id so that it cannot be used again.

## Related

- #34

### DIFF
--- a/tezos_interop/consensus.mligo
+++ b/tezos_interop/consensus.mligo
@@ -122,13 +122,25 @@ let root_hash_main
     }
 
 (* vault contract *)
-type vault_storage = (address * bytes, bytes ticket) big_map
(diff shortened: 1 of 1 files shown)


callistonianembrace (approved):

anchpop (approved):
I don't fully understand this code but it mostly makes sense so I'm approving it to unblock merging

PR #89, opened 2021-07-29
add hash_withdraw_handle on tezos interop

EduardoRFS:
## Depends

To keep the PR dependency graph simple, this one depends on

- [x] #74 

## Problem

#88 adds on the Tezos contract a `handle_structure` which when combined with the `handles_hash` allow us to communicate proofs about the sidechain to Tezos, but currently there is no way to hash `handle_structure` on the sidechain.

## Solution

This implements `hash_withdraw_handle` which creates a hash compatible with #88 `handle_structure`.

## Related

- #34
- #88

### DIFF
--- a/tezos_interop/tezos_interop.re
+++ b/tezos_interop/tezos_interop.re
@@ -348,6 +348,7 @@ module Pack = {
   type t = node(canonical_location, prim);
 
   let int = n => Int(-1, n);
+  let nat = n => Int(-1, n);
   let bytes = b => Bytes(-1, b);
   let pair = (l, r) => Prim(-1, D_Pair, [l, r], []);
   let list = l => Seq(-1, l);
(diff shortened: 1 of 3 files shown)


tezos_interop/tezos_interop.re
```diff
@@ -348,6 +348,7 @@ module Pack = {
   type t = node(canonical_location, prim);
 
   let int = n => Int(-1, n);
+  let nat = n => Int(-1, n);
   let bytes = b => Bytes(-1, b);
   let pair = (l, r) => Prim(-1, D_Pair, [l, r], []);
   let list = l => Seq(-1, l);
```
anchpop, line 354: For these, the first parameter (`-1`) is discarded during packing, so it doesn't affect the hash, which is all we care about, right?

anchpop (approved):

tezos_interop/tezos_interop.re
EduardoRFS: yup, it's the original location on the micheline, I think it's used only for error messages

PR #90, opened 2021-07-29
add withdraw handles on ledger

EduardoRFS:
## Depends

- [x] #89

## Problem

With #88 the consensus contract allow the user to provide a proof that he burned the money on the sidechain, but currently the patricia tree is not integrated in the sidechain itself.

## Solution

This doesn't solve the problem entirely, but it creates the needed infrastructure from the ledger to integrate the patricia tree in the future on the protocol.

## Related

- #34 
- #88

### DIFF
--- a/protocol/ledger.re
+++ b/protocol/ledger.re
@@ -17,10 +17,51 @@ module Wallet_and_ticket_map = {
   let find_opt = (address, ticket) => Map.find_opt({address, ticket});
   let add = (address, ticket) => Map.add({address, ticket});
 };
+module Handle = {
+  module Address = {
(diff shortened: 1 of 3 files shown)


anchpop (approved):

anchpop:
Can you point me to the bit of code that does increments the ID? I can't seem to find it. Or do we now use the position in the patricia tree as the ID?

EduardoRFS:
@anchpop https://github.com/marigold-dev/sidechain/blob/main/helpers/incremental_patricia.re#L86

TLDR if you're using Incremental_patricia the key is always incremented by one when adding.

Issue #22, opened 2021-07-09
support tz2 and tz3 on tezos_interop


EduardoRFS:
Currently we only support KT1 and tz1. So if someone sends a deposit to a tz2 or tz3 address that is gonna stay in the limbo.

PR #94, opened 2021-07-29
Protocol: Ed25519 addresses are now wrapped with Tz1 tag

callistonianembrace:
**Problem**
We only support tz1 addresses (ie ed25519 only)

**Solution**
In an attempt to address #22 , this PR, wraps address as tz1 explicitly and expects ed25519 address to be treated explicitly so. This lets future PRs bring in support for p256 and secp256k1 addresses.


EduardoRFS:
Sorry for the description on the issue, but we need to support tz2 and tz3 inside of Tezos_interop.Address, not in the sidechain. On the sidechain itself we don't support even KT1.

d4hines:
Can we safely close this now that we have #212 and company?

callistonianembrace:
@d4hines This PR isn't in our milestones I think. Let's revisit this once it is?

callistonianembrace:
It seems @Zett98 has taken up this in his own PR. Can you please comment the link here, Zett98? We can close this PR.

Issue #61, opened 2021-07-26
tezos integration should be driven by the node


EduardoRFS:
## Problem

Currently we don't have any tezos integration in the node. This means that you need weird external scripts to sync the sidechain with the mainchain.

This makes testing the full pipeline for sidechains much harder and essentially I'm the only one who can do it.

PR #98, opened 2021-07-30
make cli to run tezos contracts with taquito

EduardoRFS:
## Problem

To achieve #61 we should be able to run tezos contracts directly from the node, this currently is not possible.

## Solution

Instead of reimplementing the logic to talk to a Tezos RPC and handle things like waiting for confirmations, this PR proposes that we use taquito by making node CLIs that can be in the future called by the node itself.

Essentially an API to talk to Ligo contracts provided by the sidechain.

## About code

I've used JSDoc to at least document the interface as IIRC typescript doesn't work well with esy sadly, also because this avoids a build step for now.

## Related

- #61

### DIFF
--- a/esy.lock/index.json
+++ b/esy.lock/index.json
@@ -1,5 +1,5 @@
 {
-  "checksum": "c95887a15c60b768ecdf636377d6c25a",
+  "checksum": "961787d2abfcfb599ccae2acb23baada",
   "root": "sidechain@link-dev:./package.json",
   "node": {
     "yarn-pkg-config@github:esy-ocaml/yarn-pkg-config#db3a0b63883606dd57c54a7158d560d6cba8cd79@d41d8cd9": {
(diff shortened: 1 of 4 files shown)


anchpop (approved):
I don't really understand what this code is doing. It looks like it reads some json from stdin, extracts `payload` from that json, sorts that and discards the keys (I don't understand this part), then passes that to a contract determined by the input json, then waits for a confirmation?

EduardoRFS:
@anchpop that's exactly it, it sorts the payload by key, then drop the keys, because of how Ligo works all records are compiled in alphabetical order as `{ a, b, c, d, e }` will become something like `(pair (pair (pair a b) (pair c d)) e)` the following. So we just use the values sorted by name that will duplicate the behavior from Ligo. Also don't like it but it works.

Opened an issue so that we on the future don't need to rely on duplicated logic for this.

PR #100, opened 2021-08-01
add commit_state_hash to Tezos_interop

EduardoRFS:
## Problem

To achieve #61 we need to be able to commit the state hash from inside of the node.

## Solution

This is a movement in this direction, it integrates on the CLI provided by #98 and exposes `commit_state_hash` which allows the node to call and submit the operation to Tezos.

## Related

- #61

### DIFF
--- a/tezos_interop/tezos_interop.re
+++ b/tezos_interop/tezos_interop.re
@@ -391,6 +391,73 @@ module Context = {
     required_confirmations: int,
   };
 };
+module Run_contract = {
+  [@deriving to_yojson]
+  type input = {
+    rpc_node: string,
+    secret: string,
+    confirmation: int,
+    destination: string,
+    entrypoint: string,
+    payload: Yojson.Safe.t,
+  };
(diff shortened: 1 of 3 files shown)


tezos_interop/tezos_interop.re
```diff
@@ -422,6 +489,59 @@ module Consensus = {
       address(ticketer),
     )
     |> hash_packed_data;
+
+  // TODO: how to test this?
```
callistonianembrace, line 493: re: testing

We could start maintaining a mock tezos node for unit testing.

callistonianembrace: (feel free to resolve)

callistonianembrace (approved):

anchpop (approved):

tezos_interop/tezos_interop.re
EduardoRFS: #101 good idea

PR #106, opened 2021-08-02
Inline JS scripts inside the node binary

EduardoRFS:
## Problem

Since #105 we use a JS script to commit the sidechain state hash to Tezos, this is a problem as to run this script we need not only `node` but also `esy` which all the dependencies installed locally.

## Solution

This uses webpack to generate a standalone bundle and ppx_blob to inline the JS file in the binary, and during runtime the binary creates a temporary file with JS which will be used during the lifetime of the node.

With this you only need `node` available on the environment which is a much smaller requirement.

## **Warning**

This requires you to update your esy to the latest nightly as esy had a problem related to JS packages inside of the `_esy`.

### DIFF
--- a/tezos_interop/dune
+++ b/tezos_interop/dune
@@ -2,4 +2,13 @@
  (name tezos_interop)
  (libraries helpers lwt.unix mirage-crypto-ec tezos-micheline)
  (preprocess
-  (pps ppx_deriving_yojson)))
+  (pps ppx_deriving_yojson ppx_blob))
+ (preprocessor_deps
+  (file ./run_entrypoint.bundle.js)))
+
+(rule
+ (deps ./webpack.config.js ./run_entrypoint.js)
+ (targets ./run_entrypoint.bundle.js)
+ (mode fallback)
+ (action
+  (run webpack --env=%{profile})))
--- a/tezos_interop/tezos_interop.re
+++ b/tezos_interop/tezos_interop.re
@@ -439,6 +439,7 @@ module Run_contract = {
   };
 
   // TODO: probably we should add a ppx to load this file in the bundle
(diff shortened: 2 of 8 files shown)


tezos_interop/tezos_interop.re
```diff
@@ -456,8 +461,11 @@ module Run_contract = {
         (command, [|command, file|]),
         Yojson.Safe.to_string(input_to_yojson(input)),
       );
+    print_endline(output);
     switch (Yojson.Safe.from_string(output) |> output_of_yojson) {
-    | Ok(data) => await(data)
+    | Ok(data) =>
+      print_endline("there");
```
anchpop, line 467: I think this might be a mistake

anchpop (approved):
LGTM, I haven’t tested it locally but I’m approving it to unblock merging, will test it tomorrow

tezos_interop/tezos_interop.re
EduardoRFS: Oh nice, forgot some high level debugging symbols

tezos_interop/dune
```diff
@@ -2,4 +2,13 @@
...
+  (pps ppx_deriving_yojson ppx_blob))
+ (preprocessor_deps
+  (file ./run_entrypoint.bundle.js)))
+
+(rule
+ (deps ./webpack.config.js ./run_entrypoint.js)
+ (targets ./run_entrypoint.bundle.js)
+ (mode fallback)
+ (action
+  (run webpack --env=%{profile})))
```
callistonianembrace, line 14: Running webpack or any node binaries isn't well supported in esy. It's nice that running `esy` will be the JS artifacts too, but if it causes trouble, we'll have to move it outside the build environment (therefore, outside Dune too).

Great job though

PR #111, opened 2021-08-03
Docs: Explain node setup along with a sh script as illustrating example

callistonianembrace:
**Problem**
We haven't documented the new node setup - understandably since only recent we got the end-to-end infra ready.

**Solution**
This PR walks the reader through the setups involved in setting up validator identities, settings up initial set of config and data files and a working demo of producing and signing (validating) blocks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,75 @@ esy test
 
 ### Running the node
 
+#### Deployment
+
+`esy x http_server` would run the node (currently over HTTP). `http_server` takes one argument - folder containing the config and data files.
+
+Example usage,
+
+```sh
+esy x http_server data/0
+```
+
+##### TLDR; If you can follow annotated shell scripts, checkout [setup.sh](./setup.sh), which contains the following steps as sequence of sh commands. You'll need [jq](https://github.com/stedolan/jq)
+
+To configure and setup the node, 
+1. Create identity for the node with `esy x sidecli generate-identity`. 
+   Step #1 will most likely be run multiple times for each node run by each participant. We create a list of identities of such authorised participants (validators), and use them in the next step - originating the consensus contract.
+2. `consensus.mligo` (found in `tezos_interop/`) needs to be deployed - a preloaded LIGO IDE with the contract can be found [here](https://ide.ligolang.org/p/-x6CdYJ5tIEaVzD9lGYsaA). Specify the list of validators' public addresses in `current_validators` field of the storage. Example,
+
+```ml
+{
+  root_hash = {
+    current_block_hash = 0x;
+    current_block_height = 0;
+    current_state_hash = 0x;
+    current_handles_hash = 0x;
+    current_validators = ([("edpk...": key)]: validators);
+  };
+  vault = {
+    known_handles_hash = (Big_map.empty : vault_known_handles_hash_set);
+    used_handles = (Big_map.empty : vault_used_handle_set);
+    vault = (Big_map.empty : vault);
+  }
+}
+```
+3. Use `setup-node` subcommand from `sidecli` to generate the config and the data files
+
+```sh
+    esy x sidecli setup-node "$data_directory" \
+	--secret=< \
+	--tezos_consensus_contract $contract_address_from_step_2 \
+	--tezos_rpc_node https://testnet-tezos.giganode.io \
+	--tezos_secret $test_net_wallet_secret \
+	--uri=$uri
+```
+
+4. Run the node
+
+```sh
+esy x http_server $data_directory
+```
+
+##### Testing the node - producing a block
+
+  To see the chain in action, blocks can be produced by hand with the following steps
+  
+  1. Produce a block and note the block hash
+  
+  ```sh
+  HASH=$(esy x sidecli produce-block ./0/wallet.json ./0/state.bin | awk '{ print $2 }')
+  ```
+  
+  2. Sign the block at each validator
+  
+  ```sh
+  esy x sidecli sign-block 0/wallet.json $HASH
+  esy x sidecli sign-block 1/wallet.json $HASH
+  esy x sidecli sign-block 2/wallet.json $HASH
+  ```
+
+#### For Development
 The node binary is `http_server.exe` present in `bin/` folder. One way
 to execute it would be,
 
--- a/bin/sidecli.re
+++ b/bin/sidecli.re
@@ -323,6 +323,40 @@ let produce_block = {
   Term.(lwt_ret(const(produce_block) $ key_wallet $ state_bin));
 };
 
+let info_gen_identity = {
+  let doc = "Create a public-private key pair for a validator";
+  Term.info(
+    "generate-identity",
+    ~version="%‌%VERSION%%",
+    ~doc,
+    ~exits,
+    ~man,
+  );
+};
+
+let gen_identity = {
+  let f = () => {
+    open Mirage_crypto_ec;
+    let (sk, pk) = Ed25519.generate();
+    module Identity = {
+      [@deriving yojson]
+      type t = {
+        secret_key: string,
+        public_key: string,
+      };
+    };
+    open Tezos_interop;
+    print_endline(
+      Identity.to_yojson({
+        secret_key: Secret.(Ed25519(sk) |> to_string),
+        public_key: Key.(Ed25519(pk) |> to_string) //"edpk" ++ (pk |> Ed25519.pub_to_cstruct |> Cstruct.to_string |> Helpers.to_hex),
+      })
+      |> Yojson.Safe.to_string,
+    );
+  };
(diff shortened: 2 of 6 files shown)


callistonianembrace:
Draft status because faucet generated secret isn't compatible with `--tezos_secret` in the `setup-node` command.

.tmuxinator.yml
```diff
@@ -49,9 +49,8 @@ windows:
...
       panes:
-        - sleep 1 && $cur__target_dir/default/bin/http_server.exe 0
-        - sleep 1 && $cur__target_dir/default/bin/http_server.exe 1
-        - sleep 1 && $cur__target_dir/default/bin/http_server.exe 2
-        - sleep 1 && $cur__target_dir/default/bin/http_server.exe 3
-        - $cur__target_dir/default/bin/sidecli.exe make-credentials && sleep 3 && $cur__target_dir/default/bin/sidecli.exe inject-genesis
+        - sleep 1 && esy x http_server 0
+        - sleep 2 && esy x http_server 1
+        - sleep 3 && esy x http_server 2
+        - sleep 4 && esy x sidecli inject-genesis
```
anchpop, line 55: Does this no longer cause issues with esy?

anchpop (approved):

.tmuxinator.yml
anchpop: Actually it seems to work fine. Does that mean [this](https://github.com/marigold-dev/sidechain/blame/b31a21169c9efad588487268bf03bd4cd6b0941b/README.md#L170) should be changed, or does that still not work?

README.md
```diff
@@ -40,6 +40,90 @@ esy test
...
+    esy x sidecli setup-node "$data_directory" \
+	--secret=< \
+	--tezos_consensus_contract $contract_address_from_step_2 \
+	--tezos_rpc_node https://testnet-tezos.giganode.io \
+	--tezos_secret $test_net_wallet_secret \
+	--uri=$uri
+```
+
+**Obtaining the `tezos_secret` key
+This option takes the private key of the wallet to (TODO: do what?). While working with the test network, one needs to import the `secret` from the activate code JSON downloaded from [faucet.tzalpha.net](https://faucet.tzalpha.net) and activate the address.
```
anchpop, line 88: Should we add something about how to generate the key for `secret` too? That's the secret used by the validator that its address is derived from. 

  tezos-secret is one of the parameters in the json passed to [run_entrypoint.js](https://github.com/marigold-dev/sidechain/blob/0d41c92a9f9bec659c2eb7cbd9ecce703cf7e89c/tezos_interop/run_entrypoint.js#L14-L14) (specifically the parameter `secret`). The script uses it to sign a smart contract operation. 

In our use case, the script calls the entrypoint `update_root_hash` on the contract corresponding to the address passed in `tezos_consensus_contract`, and signs it with `tezos_secret`. 

@EduardoRFS what properties should the key passed for `tezos_secret` have? does it need to be configured with the mainchain smart contract in some way, or have money in its wallet to pay for gas, or something like that?

README.md
```diff
@@ -40,6 +40,90 @@ esy test
...
+2. `consensus.mligo` (found in `tezos_interop/`) needs to be deployed - a preloaded LIGO IDE with the contract can be found [here](https://ide.ligolang.org/p/-x6CdYJ5tIEaVzD9lGYsaA). Specify the list of validators' public addresses in `current_validators` field of the storage. Example,
+
+```ml
+{
+  root_hash = {
+    current_block_hash = 0x;
+    current_block_height = 0;
+    current_state_hash = 0x;
+    current_handles_hash = 0x;
+    current_validators = ([("edpk...": key)]: validators);
```
anchpop, line 67:
```suggestion
    // https://ligolang.org/docs/advanced/timestamps-addresses/#keys
    current_validators = ([("edpk...": key)]: validators);
```

setup.sh
```diff
@@ -0,0 +1,99 @@
...
+# @example join , "a" "b" -> "a,b"
+join () {
+  local IFS="$1"
+  shift
+  echo "$*"
+}
+
+
+data_directory="data"
+validators=(0 1 2)
```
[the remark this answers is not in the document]
anchpop: I had to invoke it with bash explicitly `bash ./setup.sh` to fix this, but I'm too much of a linux newb to understand why

setup.sh
```diff
@@ -0,0 +1,99 @@
...
+#    needs to import the `secret` from the activate code JSON downloaded
+#    from [faucet.tzalpha.net](https://faucet.tzalpha.net) and activate
+#    the address. 
+read -p "Download wallet from https://faucet.tzalpha.net and enter the path to the json file: " account_json_path
+
+# To import the keys obtained from faucet
+tezos-client -p "PsFLorenaUUuikDWvMDr6fGBRG8kt3e3D3fHoXK1j1BFRxeSH4i" --endpoint https://testnet-tezos.giganode.io import keys from mnemonic alice
+
+# To activate the address
+tezos-client -p "PsFLorenaUUuikDWvMDr6fGBRG8kt3e3D3fHoXK1j1BFRxeSH4i" --endpoint https://testnet-tezos.giganode.io activate account alice with "$account_json_path"
```
anchpop, line 86: Where does `PsFLorenaUUuikDWvMDr6fGBRG8kt3e3D3fHoXK1j1BFRxeSH4i` come from?

bin/sidecli.re
```diff
@@ -323,6 +323,45 @@ let produce_block = {
...
+let gen_identity = {
+  let f = uri => {
+    open Tezos_interop;
+    let (sk, pk) = Mirage_crypto_ec.Ed25519.generate();
+    module Identity = {
+      [@deriving yojson]
+      type t = {
+        secret_key: string,
+        public_key: string,
+        uri: Uri.t,
```
anchpop, line 352: why not use the [files api](https://github.com/marigold-dev/sidechain/blob/78866292ce5da54ed21d47483123e884adc8216c/bin/files.re#L19-L22) here?

.tmuxinator.yml
callistonianembrace: Yes, that entire section about esy not managing lock files can go away

setup.sh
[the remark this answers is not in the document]
callistonianembrace: What does `sh --version` say?

README.md
callistonianembrace: > Should we add something about how to generate the key for secret too? 
The `tezos-client` commands that follow - they're suppose to answer this. Would you like me to clarify? What's not clear right now?

setup.sh
callistonianembrace: Yes. That's the protocol hash of Florence

bin/sidecli.re
callistonianembrace: Because these are only meant to logged to the console - the way @EduardoRFS wanted it.

setup.sh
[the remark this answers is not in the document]
anchpop: It says 

```
$ sh --version
sh: 0: Illegal option --
```

I assume that isn't what you wanted - I can't figure out how to actually get the `sh` version

setup.sh
```diff
@@ -0,0 +1,99 @@
+#! /bin/sh
```
anchpop, line 1:
```suggestion
#! /bin/bash
```

This makes invoking `./setup.sh` work on my machine. Otherwise it tries to use `sh` which doesn't support some of the syntax used

setup.sh
anchpop: Are both of these commands necessary? if the first one imports the key from the mnemonic, what's the purpose of the second?

README.md
callistonianembrace: Can I resolve this conversation?

setup.sh
callistonianembrace: It converts the blinded pkh into an activated one - would you like me to point you to the docs?

callistonianembrace: https://tezos.gitlab.io/developer/proposal_testing.html?highlight=activate

setup.sh
callistonianembrace: Thank you!

setup.sh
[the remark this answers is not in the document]
callistonianembrace: What does the shell manual say? Which shell is it?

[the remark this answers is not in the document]
callistonianembrace: (Not a blocking thread)

setup.sh
callistonianembrace: Resolving for now, but we can reopen if necessary

callistonianembrace:
This PR has been rebased. We have new commands now - `setup-identity` and `setup-tezos`? @EduardoRFS Can you add the intended invocations here?

Issue #64, opened 2021-07-27
Better error handling and logging


callistonianembrace:
We're missing useful error logs.

Eg:

https://github.com/marigold-dev/sidechain/blob/bb551857634a0775fdcd7f8d07eba8cd15264e77/node/networking.re#L35

This issue could be addressed by adding a logging layer
1. API with multiple levels of logging. Perhaps, Lwt_log itself
2. Extensive usage throughout the codebase

callistonianembrace:
Desirable features of this logging. 

1. Easily filterable, making it easy to debug during crash events (or any anomaly)
2. Useful context (we could take inspiration from `RunAsync.t` in esy

callistonianembrace:
Related: https://github.com/marigold-dev/sidechain/pull/108

PR #113, opened 2021-08-04
Add logging

anchpop:
# Problem

We have no unified way of adding logs. See #64. Logs are often useful when debugging, but clutter up the console output otherwise, and it's annoying to have to remove and re-add them every time.

# Solutions

This PR adds the `Logs` OCaml library and sets it up for use within `sidecli` and `http_server`.  It also modifies all `sidecli` subcommands to take an option indicating that verbose logging should be enabled, and an option to control logging colors. (By default colors are on in sidecli, and vebosity is off.) Once I make `http_server` use `cmdliner` I will make the same change there.

### DIFF
--- a/bin/sidecli.re
+++ b/bin/sidecli.re
@@ -28,6 +28,26 @@ let man = [
   `P("Email bug reports to <contact@marigold.dev>."),
 ];
 
+// Log helpers
+// We take logging arguments on the command line (e.g. -v to enable verbosity)
+// So instead of writing e.g. `const(sign_block) $ folder_node $ block_hash`,
+// You would write `const_log(sign_block) $ folder_node $ block_hash` to enable
+// logging support.
+let setup_log = (style_renderer, level) => {
+  switch (style_renderer) {
+  | Some(style_renderer) => Fmt_tty.setup_std_outputs(~style_renderer, ())
+  | None => Fmt_tty.setup_std_outputs()
+  };
+
+  Logs.set_level(level);
+  Logs.set_reporter(Logs_fmt.reporter());
+  ();
+};
+let setup_log =
+  Term.(const(setup_log) $ Fmt_cli.style_renderer() $ Logs_cli.level());
+
+let const_log = p => Term.(const(() => p) $ setup_log);
+
 // Todo from Andre: we may have to do peer discovery?
 // I don't know anything about that, except for having played with hyperswarm.
 let validators_uris = node_folder => {
@@ -164,7 +184,7 @@ let create_wallet = () => {
   await(`Ok());
 };
 
-let create_wallet = Term.(lwt_ret(const(create_wallet) $ const()));
+let create_wallet = Term.(lwt_ret(const_log(create_wallet) $ const()));
 
 // create-transaction
 
@@ -350,7 +370,7 @@ let withdraw = {
 
   Term.(
     lwt_ret(
-      const(withdraw)
+      const_log(withdraw)
       $ folder_node
       $ address_from
       $ tezos_address
@@ -472,7 +492,7 @@ let sign_block_term = {
     Arg.(required & pos(1, some(hash), None) & info([], ~doc));
   };
 
-  Term.(lwt_ret(const(sign_block) $ folder_node $ block_hash));
+  Term.(lwt_ret(const_log(sign_block) $ folder_node $ block_hash));
 };
 
 // produce-block
@@ -513,7 +533,7 @@ let produce_block = {
     Arg.(required & pos(0, some(string), None) & info([], ~doc, ~docv));
   };
 
-  Term.(lwt_ret(const(produce_block) $ folder_node));
+  Term.(lwt_ret(const_log(produce_block) $ folder_node));
 };
 
 let ensure_folder = folder => {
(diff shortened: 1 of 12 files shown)


bin/http_server.re
```diff
@@ -15,6 +15,9 @@ open Protocol;
 open Node;
 open Networking;
 
+Logs.set_reporter(Logs_fmt.reporter());
+Logs.set_level(Logs.level_of_string("debug") |> Result.get_ok);
```
EduardoRFS, line 19: I think there is a variant under `Logs` maybe `Logs.Debug`?

package.json
```diff
@@ -56,4 +57,4 @@
     "@opam/opium": "github:EduardoRFS/opium:opium.opam#9fdbfbd0eedf238e1be0663ab95977be27d9a6e1",
     "@opam/rock": "github:EduardoRFS/opium:rock.opam#9fdbfbd0eedf238e1be0663ab95977be27d9a6e1"
   }
-}
+}
```
EduardoRFS, line 60: You need to fix your editor

EduardoRFS (commented):
Not fully reviewed just some comments.

Overall I like the goal but I think there is a requirement to having known keys on logs and we also need to check how costly is the `Logs_lwt` vs `Logs`, my gut feeling tells me that `Logs_lwt` will not be worth on performance to us but I may be wrong.

anchpop:
@EduardoRFS what do you mean by "having known keys on logs"?

And Logs_lwt should be basically a no-op if we disable logging, right? Or does the performance matter in other cases?

EduardoRFS:
> And Logs_lwt should be basically a no-op if we disable logging, right?

I'm not sure if it is worth it the complexity of having a monad, I feel like the normal Logs is good enough.

> What do you mean by "having known keys on logs"?

A good way to ensure that we index things in a per module basis without affecting the performance that much.

bin/http_server.re
```diff
@@ -15,6 +15,9 @@ open Protocol;
 open Node;
 open Networking;
 
+Logs.set_reporter(Logs_fmt.reporter());
+Logs.set_level(Some(Logs.Debug));
```
callistonianembrace, line 19: I'd prefer if this is Error by default, perhaps Warning. By making it Debug, we treat all levels of the logs with the same priority

callistonianembrace: Spoke to early: Line 19 shouldn't be necessary - Logs_cli should set the default level to warning already.

callistonianembrace (approved):
LGTM

Commented on setting of the default level as I didn't understand why it was done. I find value in the PR otherwise. LGTM

bin/http_server.re
callistonianembrace: Ignore me. http_server isn't using Logs_cli

callistonianembrace:
This is a useful PR, @EduardoRFS. Does it need more work?

IMO, the title needs to reflect that the logs are not just added to sidecli.

anchpop:
@callistonianembrace Edu mentioned that it would be nice if the logging included things like the file/line number the log was emitted at and the block height. The first requires I think a macro like the one @ulrikstrid wrote, and there's multiple ways to do the second but I thought the easiest would be to wait until we have an effect system

EduardoRFS:
@anchpop please rebase this to begin.

bin/sidecli.re
```diff
@@ -45,6 +59,8 @@ let exits =
 
 let lwt_ret = p => Term.(ret(const(Lwt_main.run) $ p));
 
+let const_log = p => Term.(const(() => p) $ setup_log);
```
d4hines, line 62: Can you explain the purpose of this function? It looks like it runs `setup_log`. Why do we have to run `setup_log` more than once?

anchpop: It should be only getting called once per sidecli invokation. If it's being called any more than that I've made a mistake

anchpop:
@callistonianembrace 

> Commented on setting of the default level as I didn't understand why it was done. I find value in the PR otherwise. LGTM

That's a temporary measure until #113 is merged and the level can be set by cmdliner

d4hines (approved):
LGTM

EduardoRFS (changes requested):
Two major problems, why `Logs_lwt`? What are we gaining on it? Seems like it should be removed.

At least we should have all logs message be prefixed by a key or something. Especially in a manner that can be easily filtered from the CLI, otherwise logging will become really annoying really fast.

Even if we don't have the line that is triggering the log this is a must have.

anchpop:
@callistonianembrace I don't actually understand what Logs_lwt is even doing. Logging is normally a synchronous action, right? In what situation would Logs_lwt ever be preferred? I'm actually even more confused because [the docs say this](https://erratique.ch/software/logs/doc/Logs/index.html#sync) about the cooperative reporter example you linked:

> However a purely synchronous reporter like Reporter acting on channels does not play well with Lwt's cooperative runtime system. It is possible to reuse Reporter to define a cooperative reporter, see [this example](https://erratique.ch/software/logs/doc/Logs_lwt/index.html#report_ex). However while this reporter makes Logs_lwt's log functions synchronous, those of Logs behave asynchronously. For now it seems it that this is unfortunately the best we can do if we want to preserve the ability to use Logs with or without cooperative concurency.

And it seems very weird that Logs would behave asynchronously while Logs_lwt would behave synchronously (I would expect it to be the reverse). I can take out all the lwt stuff since @EduardoRFS says there's no benefit, but also I'd like to understand for what use case logs_lwt even exists at all

PR #115, opened 2021-08-04
Tmuxinator: Build before starting tmux

anchpop:
# Problem

If you make a change, `esy x <command>` rebuilds before executing the command, which messes up the timing (the nodes must start before inject-genesis runs, for example). 

# Solutions

This adds a line in the `.tmuxinator.yml` that builds the project before opening tmux, so everything is always built and the timings are right.

### DIFF
--- a/.tmuxinator.yml
+++ b/.tmuxinator.yml
@@ -11,7 +11,7 @@ root: ./data
 
 # Project hooks
 # Runs on project start, always
-# on_project_start: command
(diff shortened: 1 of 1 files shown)


callistonianembrace (approved):

callistonianembrace:
Good one!

EduardoRFS:
another nice small dev experience improvement

PR #123, opened 2021-08-06
add setup-identity command

EduardoRFS:
## Problem

To `setup-node` you need to have a consensus contract address, but to setup the consensus address you need an identity which is provided by `setup-node`.

## Solution

This PR is a first step in separating `setup-node` to avoid problems the one above. It is based on the work at #111 by @callistonianembrace 

## Related

- #111

### DIFF
--- a/bin/sidecli.re
+++ b/bin/sidecli.re
@@ -451,6 +451,53 @@ let inject_genesis = {
   );
 };
 
+let ensure_folder = folder => {
+  let.await exists = Lwt_unix.file_exists(folder);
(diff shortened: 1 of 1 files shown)


anchpop (approved):

callistonianembrace:
> but to setup the consensus address you need an identity which is provided by setup-node.

This is not true - there was `generate-identity` in that PR (as suggested)

EduardoRFS:
@callistonianembrace yes, as mentioned

> It is based on the work at #111

PR #131, opened 2021-08-08
improve http_server error handling

EduardoRFS:
## Problem

Currently http_server logs only a few of the available error messages, this makes it quite hard to debug simple mistakes.

## Solution

This is just an improvement, it prints all the errors explicitly on the http_server module.

## Related

- #64

### DIFF
--- a/bin/http_server.re
+++ b/bin/http_server.re
@@ -19,6 +19,31 @@ let ignore_some_errors =
   fun
   | Error(#Flows.ignore) => Ok()
   | v => v;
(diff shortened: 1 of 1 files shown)


anchpop (approved):
Thank you! I wish there was a better way than manually copying each type into a string

anchpop (approved):
Thank you! I wish there was a better way than manually copying each type into a string

EduardoRFS:
@anchpop typedppxlib idea right here

PR #136, opened 2021-08-09
Readme: remove inaccurate section about lock files

anchpop:
## Problem

The readme contains information about esy lock files that is no longer up to date

## Solution

Remove that section


EduardoRFS (approved):
a cute star for a cute PR

PR #137, opened 2021-08-09
Prevent replay attack on main operations

EduardoRFS:
## Depends

- [x] #135 

## Problem

Currently an operation can be added two times in a block and it will be applied, this is not the desirable as it allows to multiply money.

## Solution

Do the same as for Operation.Side_chain, have a list of included main operations on the protocol state. The only difference is that this will never be cleaned as there is no such a thing as a maximum block_height on main operations.

### DIFF
--- a/tests/test_protocol.re
+++ b/tests/test_protocol.re
@@ -135,12 +135,17 @@ describe("protocol state", ({test, _}) => {
     expect.option(Validators.current(validators)).toBeNone();
     expect.list(Validators.to_list(validators)).toBeEmpty();
     let new_validator = Validators.{address: Address.make_pubkey()};
-    let main_op = kind =>
-      Operation.Main_chain.make(
-        ~tezos_hash=Helpers.BLAKE2B.hash(""),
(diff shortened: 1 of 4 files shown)


tests/test_protocol.re
```diff
@@ -135,12 +135,17 @@ describe("protocol state", ({test, _}) => {
...
+    let main_op = {
+      let accu = ref(0);
+      kind => {
+        incr(accu);
+        Operation.Main_chain.make(
+          ~tezos_hash=Helpers.BLAKE2B.hash(""),
+          ~tezos_index=accu^,
+          ~kind,
+        );
+      };
```
anchpop, line 147: Does the accumulator need to be stored in the state somewhere? Otherwise won't it be reset to zero whenever the node restarts?

anchpop (approved):

tests/test_protocol.re
EduardoRFS: this is a test

PR #139, opened 2021-08-10
Setup.sh: Add instructions for using mockup client

anchpop:
# The problem

Users of the docker client like me would have found it difficult to use the previous script as-is.

Also, the script used the account alias `alice` every time, which made it fail for repeat runs. (Also 'alice' is not a very informative name.)

Also, if you want to use a mockup client to avoid the hassle of running a full node, that was difficult in the previous version, but I added some instructions for how to do that.

# The solution

Fix all these things in one monster PR. My justification is that I'm not making a PR to merge into main so I don't need to make it good.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -1,5 +1,13 @@
-#! /bin/bash
+#! /bin/sh
+tezos_client_command="tezos_client"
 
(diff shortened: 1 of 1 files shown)


callistonianembrace (approved):

callistonianembrace:
Thanks!

Issue #118, opened 2021-08-05
clean top level driven experience


EduardoRFS:
I would like if we could upstream some of the work on #46 to utop or dune so that rtop can be integrated automatically on it.

A standard `dune utop` that detected that most of the project is in Reason or `dune utop` where you can setup an automated `#required "rtop";;` that changes the parser of utop would also do it.

## Related

#46

callistonianembrace:
This is what we want to achieve

```sh
#! /bin/sh

set -e

TMPDIR="$cur__target_dir/tmp";
mkdir -p $TMPDIR;
dune top | sed "s/\;\;/\;/g" > "$TMPDIR/$cur__name.re";
rtop -init "$TMPDIR/$cur__name.re" "$@"
```

This script also replaces #46

d4hines:
Dune utop works fine now.

PR #141, opened 2021-08-10
Use forked Dune to support reason compatible dune top

callistonianembrace:
**Problem**
We use a custom toplevel to load libraries from dune top

**Solution**
Patch Dune to emit Reason compatible output (from dune top)
Addresses #118


callistonianembrace:
With patched Dune we can simplify it to,

```sh
#! /bin/sh

set -e

rtop -init "$cur__root/tools/top/init" "$@"
```

With the init file containing
```sh
#use_output "dune top --reason-mode" 
```

anchpop (approved):

EduardoRFS:
I have a hard time with justifying using a forked version of dune only for this, I feel like this needs to be discussed further.

EduardoRFS:
We will be moving to OCaml soon

PR #150, opened 2021-08-12
sign-block expects only a node folder

EduardoRFS:
## Problem

Currently sign-block requires a `wallet.json` but this sign-block is needed to spawn a chain, but `wallet.json` should not be required to spawn a chain.

## Solution

Makes `sign-block` expects only a folder with `identity.json` like `produce-block`.

### DIFF
--- a/bin/sidecli.re
+++ b/bin/sidecli.re
@@ -245,9 +245,10 @@ let info_sign_block = {
   let doc = "Sign a block hash and broadcast to the network manually, useful when the chain is stale.";
   Term.info("sign-block", ~version="%‌%VERSION%%", ~doc, ~exits, ~man);
 };
-let sign_block = (wallet_file, block_hash) => {
-  let.await wallet = Files.Wallet.read(~file=wallet_file);
-  let signature = Signature.sign(~key=wallet.priv_key, block_hash);
+let sign_block = (folder, block_hash) => {
+  let file = folder ++ "/identity.json";
+  let.await identity = Files.Identity.read(~file);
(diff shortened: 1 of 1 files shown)


bin/sidecli.re
```diff
@@ -245,9 +245,10 @@ let info_sign_block = {
   let doc = "Sign a block hash and broadcast to the network manually, useful when the chain is stale.";
   Term.info("sign-block", ~version="%‌%VERSION%%", ~doc, ~exits, ~man);
 };
-let sign_block = (wallet_file, block_hash) => {
-  let.await wallet = Files.Wallet.read(~file=wallet_file);
-  let signature = Signature.sign(~key=wallet.priv_key, block_hash);
+let sign_block = (folder, block_hash) => {
+  let file = folder ++ "/identity.json";
```
anchpop, line 249: Are the standard names for files like this included in the files api? if so can we refer to that instead of hardcoding them in multiple places? not urgent

anchpop (approved):

anchpop:
Is there any information in identity.json besides the URI that can't be derived from wallet.json? If no, will identity.json eventually be superseded by wallet.json, once we get P2P and there's no need to hardcore the URIs?

bin/sidecli.re
EduardoRFS: That's the idea and why I'm making this changes, in a future PR to put them on the Files module

EduardoRFS:
@anchpop You're still gonna need to provide the URI in the future to ensure that it's easy to discover, so nope. The URIs should be used in the future only to do peer discovery.

PR #151, opened 2021-08-12
add script to easy setup

EduardoRFS:
## Problem

Currently you need to create a couple files by hand to setup a node, and also deploy the consensus contract, currently this is not automated at all.

## Solution

Here we make a script, that will prompt you to deploy the contract and create all the files needed after you deploy the contract.

## Related

This is based on @callistonianembrace work at:

- #111

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -1,0 +1,99 @@
+#! /bin/sh
+
+set -e 
+data_directory="data"
+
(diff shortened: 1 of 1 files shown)


anchpop (approved):
Just looked at this, haven't tested it locally, but it looks good to me

PR #152, opened 2021-08-16
always use validators from a node

EduardoRFS:
## Problem

Currently the validator list on the sidecli are loaded from a hardcoded path, this means that if the files are in a different folder everything stops working.

## Solution

Always receives a node folder as input and read the `validators.json` from this folder, in practice this means that now `create-transaction` also has an additional folder_node  argument.

### DIFF
--- a/bin/sidecli.re
+++ b/bin/sidecli.re
@@ -13,17 +13,11 @@ let man = [
 
 // Todo from Andre: we may have to do peer discovery?
 // I don't know anything about that, except for having played with hyperswarm.
-let validators = () =>
-  [
-    Files.Identity.read(~file="0/identity.json"),
-    Files.Identity.read(~file="1/identity.json"),
-    Files.Identity.read(~file="2/identity.json"),
-    Files.Identity.read(~file="3/identity.json"),
-  ]
-  |> Lwt.all;
-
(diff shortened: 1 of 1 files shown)


bin/sidecli.re
```diff
@@ -13,17 +13,11 @@ let man = [
 
 // Todo from Andre: we may have to do peer discovery?
```
callistonianembrace, line 14: This TODO can be removed now I guess?

EduardoRFS: Nope, we still don't have peer discovery

callistonianembrace (approved):

bin/sidecli.re
callistonianembrace: Validators will be sourced from the contract - not much to with p2p at the moment is it?

callistonianembrace:
Should the setup.sh be updated @EduardoRFS ?

bin/sidecli.re
callistonianembrace: (resolving as it is non blocking)

EduardoRFS: yes and validators shouldn't be required on the sidecli anyway

PR #155, opened 2021-08-16
README on how to run a sidechain

EduardoRFS:
## Depends

- [x] #154 

## Problem

Both the start and setup script are currently undocumented on how to use it.

## Solution

This is a basic guideline, the TLDR should be `./setup.sh; ./start.sh` but I'm bad with words.


callistonianembrace (approved):

README.md
```diff
@@ -38,49 +38,41 @@ tests. These can be run with,
...
-esy b dune exec ./bin/http_server.exe <args>
-```
+To run a sidechain you will need to have a Tezos node that you control and a Tezos Secret Key(edsk).
 
-(`http_server` because the node uses HTTP for p2p communication)
+You can run a Tezos node on a testnet and use the faucet to acquire a key with funds, as described in the following link:
 
-However, every node needs an identity before it can join or start a
-network. Identities can be created with the CLI tool, `sidecli`
+- https://tezos.b9lab.com/tezos-basics/testnet-1
```
callistonianembrace, line 49: We could link to the official docs here?
https://tezos.gitlab.io/introduction/test_networks.html#get-free-funds

EduardoRFS: The official docs seems quite crude for people that are not that technical

PR #157, opened 2021-08-16
Adds a Taquito script to fetch contract storage

callistonianembrace:
## Problem

We manually generate validators.json with sh scripts on the client, where as the deployed `consensus.mligo` should be the source of truth.

## Solution

This PR adds a Taquito script that inspects a contracts storage.  `module Fetch_storage` that fetches the storage and return it as michelson.

### DIFF
--- a/tezos_interop/fetch_storage.js
+++ b/tezos_interop/fetch_storage.js
@@ -1,0 +1,64 @@
+"use strict";
+
+const fs = require("fs");
+const taquito = require("@taquito/taquito");
+const { RpcClient } = require("@taquito/rpc");
+const { InMemorySigner } = require("@taquito/signer");
+
+const { TezosToolkit } = taquito;
+/**
+ * @typedef Input
+ * @type {object}
+ * @property {string} contract_address
+ * @property {number} confirmation
+ * @property {string} rpc_node
+ */
+/** @returns {Input} */
+const input = () => JSON.parse(fs.readFileSync(process.stdin.fd));
+
+/**
+ * @typedef OutputFinished
+ * @type {object}
+ * @property {"success"} status
+ * @property {rpc.MichelsonV1Expression} storage
+ */
+/**
+ * @typedef OutputError
+ * @type {object}
+ * @property {"error"} status
+ * @property {string} error
+ */
+/** @param {OutputFinished | OutputError} data */
+const output = (data) =>
+  fs.writeFileSync(process.stdout.fd, JSON.stringify(data, null, 2));
+
+const finished = (storage) => output({ status: "success", storage });
+const error = (error) =>
+  output({ status: "error", error: JSON.stringify(error) });
+
+(async () => {
+  const { rpc_node, contract_address, confirmation } = input();
+  const client = new RpcClient(rpc_node);
+  const Tezos = new TezosToolkit(rpc_node);
+  const block = await client.getBlock(); // fetches the head
+  const contract = await client.getContract(contract_address, {
+    block: block.hash,
+  });
+  const storage = contract.script.storage;
+  /* To make sure the storage state is finalized, we query the last
+     block and any one of it's operation, and wait till it receives
+     `n` confirmations (where n is the minimum blocks needs to
+     consider reorg highly unlikely. ie finality) */
+  const operationFromHead = block.operations.flat()[0];
+  if (operationFromHead) {
+    const operation = await Tezos.operation.createTransactionOperation(
+      operationFromHead.hash
+    );
+    const result = await operation.confirmation(confirmation);
+    if (await result.isInCurrentBranch()) {
+      finished(storage);
+    } else {
+      error({ message: "Not in current Branch" });
+    }
+  }
+})().catch(error);
--- a/tezos_interop/listen_transactions.js
+++ b/tezos_interop/listen_transactions.js
@@ -47,8 +47,10 @@ operationStream.on("error", error);
 operationStream.on("data", async (content) => {
   /** @type {rpc.OperationContentsAndResultMetadataTransaction} */
   const metadata = content.metadata;
-  if (content.kind !== taquito.OpKind.TRANSACTION
-    || metadata.operation_result.status !== "applied") {
+  if (
+    content.kind !== taquito.OpKind.TRANSACTION ||
+    metadata.operation_result.status !== "applied"
+  ) {
     return;
   }
   try {
--- a/tezos_interop/tezos_interop.re
+++ b/tezos_interop/tezos_interop.re
@@ -506,24 +506,82 @@ module Run_contract = {
     };
   };
 };
-module Listen_transactions = {
-  open Tezos_micheline;
-  type michelson = Micheline.node(int, Michelson_v1_primitives.prim);
-  let michelson_of_yojson = json => {
-    // TODO: do this without serializing
-    let.ok json =
-      Yojson.Safe.to_string(json) |> Data_encoding.Json.from_string;
-    try(
-      Ok(
-        Micheline.root(
-          Data_encoding.Json.destruct(Pack.expr_encoding, json),
-        ),
+
+type michelson =
+  Tezos_micheline.Micheline.node(int, Michelson_v1_primitives.prim);
+let michelson_of_yojson = json => {
+  // TODO: do this without serializing
+  let.ok json = Yojson.Safe.to_string(json) |> Data_encoding.Json.from_string;
+  try(
+    Ok(
+      Tezos_micheline.Micheline.root(
+        Data_encoding.Json.destruct(Pack.expr_encoding, json),
+      ),
+    )
+  ) {
+  | _ => Error("invalid json")
+  };
+};
+
+module Fetch_storage: {
+  let _run:
+    (~rpc_node: Uri.t, ~confirmation: int, ~contract_address: Address.t) =>
+    Lwt.t(result(michelson, string));
+} = {
+  [@deriving to_yojson]
+  type input = {
+    rpc_node: string,
+    confirmation: int,
+    contract_address: string,
+  };
+  let output_of_yojson = json => {
+    module T = {
+      [@deriving of_yojson({strict: false})]
+      type t = {status: string}
+      and finished = {storage: michelson}
+      and error = {error: string};
+    };
+    let.ok {status} = T.of_yojson(json);
+    switch (status) {
+    | "success" =>
+      let.ok {storage} = T.finished_of_yojson(json);
+      Ok(storage);
+    | "error" =>
+      let.ok T.{error: errorMessage} = T.error_of_yojson(json);
+      Error(errorMessage);
+    | _ =>
+      Error(
+        "JSON output %s did not contain 'success' or 'error' for field `status`",
       )
-    ) {
-    | _ => Error("invalid json")
     };
   };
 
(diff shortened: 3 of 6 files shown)


anchpop (approved):

tezos_interop/webpack.config.js
```diff
@@ -1,12 +1,13 @@
...
+    listen_transactions: "./listen_transactions.js",
   },
   output: {
-    filename: '[name].bundle.js',
+    filename: "[name].bundle.js",
     path: __dirname,
   },
-  mode: env.dev ? 'development' : 'production',
-  target: 'node'
+  mode: env.dev ? "development" : "production",
```
callistonianembrace, line 11: I have opened https://github.com/marigold-dev/sidechain/pull/159 to prevent noise like this in future

callistonianembrace:
**How to test this MR?**

1. Make sure you a URI to a node in the test network where the contract was deployed. If you used LIGO to deploy the consensus.mligo contract, you'll need florencenet - this [hackmd](https://hackmd.io/wJyGfigMQOesxD_qN-pDwQ) might help.

2. Run `esy webpack` inside `tezos_interop/`

3. Run `node fetch_validators.bundle.js` and input the json. For instance,  `echo "{"rpc_node" : "http://localhost:8732", "contract_address" : "KT1THt7BQNmovken76RkErjFWaCGU2AUxVvA"}
" | node fetch_validators.bundle.js`

tezos_interop/fetch_validators.js
```diff
@@ -0,0 +1,41 @@
...
+const finished = (current_validators) => output({ current_validators });
+const error = (error) =>
+  output({ status: "error", error: JSON.stringify(error) });
+
+(async () => {
+  const { rpc_node, contract_address } = input();
+  const Tezos = new TezosToolkit(rpc_node);
+  const contract = await Tezos.contract.at(contract_address);
+  const storage = await contract.storage();
+  finished(storage.root_hash.current_validators);
```
EduardoRFS, line 40: The JS scripts should just provide us with a simple interface, and never interact directly with a contract like this. My suggestion would be a `fetch_storage.js`.

This is also useful for preventing nodes trying to commit the block hash two times and also for the validators URI discovery where we will need to query the discovery contract.

EduardoRFS (changes requested):

tezos_interop/fetch_storage.js
```diff
@@ -0,0 +1,70 @@
...
+ */
+/** @param {OutputFinished | OutputError} data */
+const output = (data) =>
+  fs.writeFileSync(process.stdout.fd, JSON.stringify(data, null, 2));
+
+const finished = (storage) => output(storage);
+const error = (error) =>
+  output({ status: "error", error: JSON.stringify(error) });
+
+const finality = 10;
```
EduardoRFS, line 44: this should not be hard coded

tezos_interop/fetch_storage.js
```diff
@@ -0,0 +1,70 @@
+"use strict";
```
EduardoRFS, line 1: The name of this file is not right, it doesn't just fetch_storage

tezos_interop/fetch_storage.js
```diff
@@ -0,0 +1,70 @@
...
+     consider reorg highly unlikely. ie finality) */
+  const block = await client.getBlock();
+  let operationFromHead = block.operations.flat()[0];
+  if (operationFromHead) {
+    const operation = await Tezos.operation.createTransactionOperation(
+      operationFromHead.hash
+    );
+    const result = await operation.confirmation(finality);
+    if (await result.isInCurrentBranch()) {
+    }
```
EduardoRFS, line 63: this is not right, this if should not just fall, otherwise because this operation is not the in the current branch it will always return the storage

tezos_interop/fetch_storage.js
```diff
@@ -0,0 +1,70 @@
...
+  const client = new RpcClient(rpc_node);
+  const Tezos = new TezosToolkit(rpc_node);
+  const contract = await Tezos.contract.at(contract_address);
+  const storage = await contract.storage();
+  /* To make sure the storage state is finalized, we query the last
+     block and any one of it's operation, and wait till it receives
+     `n` confirmations (where n is the minimum blocks needs to
+     consider reorg highly unlikely. ie finality) */
+  const block = await client.getBlock();
+  let operationFromHead = block.operations.flat()[0];
```
EduardoRFS, line 56: why let?

EduardoRFS: Also this is not what we discussed, we discussed about using the last known operation for the consensus contract, not the last known operation of a block.

tezos_interop/fetch_storage.js
```diff
@@ -0,0 +1,70 @@
...
+ * @property {string} contract_address
+ * @property {string} rpc_node
+ */
+/** @returns {Input} */
+const input = () => JSON.parse(fs.readFileSync(process.stdin.fd));
+
+/**
+ * @typedef OutputFinished
+ * @type {object}
+ * @property {string} root_hash.current_block_hash
```
EduardoRFS, line 21: This output is only the case if the storage is for consensus, if I pass another contract_address here it will not work.

tezos_interop/fetch_storage.js
```diff
@@ -0,0 +1,70 @@
...
+    );
+    const result = await operation.confirmation(finality);
+    if (await result.isInCurrentBranch()) {
+    }
+  }
+  /* However, if operations were empty, reasons being
+     1. Stale chain
+     2. Node not in archive mode
+     ... we go ahead with the storage state */
+  finished(storage);
```
EduardoRFS, line 69: You're returning the storage even if the operation is not in the current branch.

tezos_interop/listen_transactions.js
```diff
@@ -47,8 +47,10 @@ operationStream.on("error", error);
 operationStream.on("data", async (content) => {
   /** @type {rpc.OperationContentsAndResultMetadataTransaction} */
   const metadata = content.metadata;
-  if (content.kind !== taquito.OpKind.TRANSACTION
-    || metadata.operation_result.status !== "applied") {
+  if (
+    content.kind !== taquito.OpKind.TRANSACTION ||
+    metadata.operation_result.status !== "applied"
+  ) {
```
EduardoRFS, line 53: Also, please don't format code if it's not related to the PR

tezos_interop/tezos_interop.rei
```diff
@@ -154,3 +154,9 @@ module Consensus: {
 };
 
 module Discovery: {let sign: (Secret.t, ~nonce: int64, Uri.t) => Signature.t;};
+
+module Fetch_validators: {
+  let run:
+    (~rpc_node: string, ~contract_address: string) =>
+    Lwt.t(result(list(string), string));
```
EduardoRFS, line 161: This interface should not be like this, this, look how we're doing on the other taquito scripts.

EduardoRFS (changes requested):

tezos_interop/fetch_storage.js
```diff
@@ -0,0 +1,62 @@
...
+ * @property {string} contract_address
+ * @property {number} confirmation
+ * @property {string} rpc_node
+ */
+/** @returns {Input} */
+const input = () => JSON.parse(fs.readFileSync(process.stdin.fd));
+
+/**
+ * @typedef OutputFinished
+ * @type {rpc.MichelsonV1Expression}
```
EduardoRFS, line 21: You need a tag to distinguish the outputs, just add a `status` here.

tezos_interop/tezos_interop.re
```diff
@@ -506,24 +506,60 @@ module Run_contract = {
...
+    Lwt.t(result(michelson, string));
+} = {
+  [@deriving to_yojson]
+  type input = {
+    rpc_node: string,
+    confirmation: int,
+    contract_address: string,
+  };
+  [@deriving of_yojson]
+  type output = {storage: michelson};
```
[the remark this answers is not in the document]
EduardoRFS: Also the output of your script doesn't contain the `storage` key in the JSON.

EduardoRFS (changes requested):

tezos_interop/tezos_interop.re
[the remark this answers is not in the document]
callistonianembrace: You wanted the fetch storage JS script to be agnostic of it's application - if cant make it validators if that is so.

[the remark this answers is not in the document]
callistonianembrace: The output is now of type `{ status: string, storage: michelson }`

anchpop (approved):

tezos_interop/fetch_storage.js
```diff
@@ -0,0 +1,67 @@
...
+ * @typedef OutputError
+ * @type {object}
+ * @property {"error"} status
+ * @property {string} error
+ */
+/** @param {OutputFinished | OutputError} data */
+const output = (data) =>
+  fs.writeFileSync(
+    process.stdout.fd,
+    JSON.stringify({ status: "success", storage: data }, null, 2)
```
EduardoRFS, line 35: This is wrong, `error` will call output and the json generated will be `{ status: "success", storage: { status: "error" } }`

tezos_interop/tezos_interop.re
```diff
@@ -506,24 +506,65 @@ module Run_contract = {
...
+  [@deriving to_yojson]
+  type input = {
+    rpc_node: string,
+    confirmation: int,
+    contract_address: string,
+  };
+  [@deriving of_yojson]
+  type output = {
+    status: string,
+    storage: michelson,
```
EduardoRFS, line 540: This is still not addressed, the output of this function should not be a michelson, the name of the module is `Fetch_validators`.

EduardoRFS: Also the _run function is not using this type

EduardoRFS (changes requested):

tezos_interop/tezos_interop.re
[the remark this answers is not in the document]
EduardoRFS: yes but your module is not agnostic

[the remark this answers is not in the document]
callistonianembrace: How can we resolve this @EduardoRFS ?

tezos_interop/fetch_storage.js
callistonianembrace: Sorry about the oversight. Fixed in the next commit.

tezos_interop/tezos_interop.re
EduardoRFS: The name of the module is `Fetch_validators` but it's not fetching validators, it's fetching michelson. So the name of the module needs to be different or the implementation should be different.

The module has an output type, which is not used as output(actually I don't think it's even used at all). So the name of the type needs to change or it should be the output.

tezos_interop/tezos_interop.re
```diff
@@ -506,24 +506,65 @@ module Run_contract = {
...
+  };
```
EduardoRFS, line 541: That's gonna fail to parse errors as they have the type

```reasonml
type output = {
  status: string,
  error: string
}
```

My suggestion, do a `| Success(michelson) | Error(string)` variant and decode it by hand

EduardoRFS (changes requested):

EduardoRFS (approved):
anime girl smiling of happiness after merging PR

PR #158, opened 2021-08-16
Fetch validators from consensus contract

callistonianembrace:
## Depends
- [x] #157 

## Problem
We manually generate validators.json with sh scripts on the client, where as the deployed `consensus.mligo` should be the source of truth.

## Solution
This PR uses the Taquito script from #157 to fetch the validators set

### DIFF
--- a/bin/http_server.re
+++ b/bin/http_server.re
@@ -179,10 +179,24 @@ let handle_ticket_balance =
 let node = {
   let folder = Sys.argv[1];
   let.await identity = Files.Identity.read(~file=folder ++ "/identity.json");
-  let.await validators =
-    Files.Validators.read(~file=folder ++ "/validators.json");
   let.await interop_context =
     Files.Interop_context.read(~file=folder ++ "/tezos.json");
+  let.await validator_res =
+    Tezos_interop.Consensus.fetch_validators(~context=interop_context);
+  let validators =
+    switch (validator_res) {
+    | Ok(current_validators) =>
+      current_validators
+      |> List.mapi((i, validator) => {
+           (
+             switch (validator) {
+             | Tezos_interop.Key.Ed25519(k) => k
+             },
+             Printf.sprintf("https://localhost:444%d", i) |> Uri.of_string,
+           )
+         })
+    | Error(err) => failwith(err)
+    };
   let initial_validators_uri =
     List.fold_left(
       (validators_uri, (address, uri)) =>
--- a/tezos_interop/fetch_storage.js
+++ b/tezos_interop/fetch_storage.js
@@ -60,5 +60,7 @@ const error = (error) =>
     } else {
       error({ message: "Not in current Branch" });
     }
+  } else {
+    error({ message: "Internal error: operationFromHead was undefined" });
   }
 })().catch(error);
--- a/tezos_interop/tezos_interop.re
+++ b/tezos_interop/tezos_interop.re
@@ -507,8 +507,6 @@ module Run_contract = {
   };
 };
 
-type michelson =
-  Tezos_micheline.Micheline.node(int, Michelson_v1_primitives.prim);
 let michelson_of_yojson = json => {
   // TODO: do this without serializing
   let.ok json = Yojson.Safe.to_string(json) |> Data_encoding.Json.from_string;
@@ -523,8 +521,10 @@ let michelson_of_yojson = json => {
   };
 };
 
+type michelson =
+  Tezos_micheline.Micheline.node(int, Michelson_v1_primitives.prim);
 module Fetch_storage: {
-  let _run:
+  let run:
     (~rpc_node: Uri.t, ~confirmation: int, ~contract_address: Address.t) =>
     Lwt.t(result(michelson, string));
 } = {
@@ -565,7 +565,7 @@ module Fetch_storage: {
   };
   let file = Lwt_main.run(file);
 
-  let _run = (~rpc_node, ~confirmation, ~contract_address) => {
+  let run = (~rpc_node, ~confirmation, ~contract_address) => {
     let input = {
       rpc_node: Uri.to_string(rpc_node),
       confirmation,
@@ -660,6 +660,7 @@ module Listen_transactions = {
 };
 module Consensus = {
   open Pack;
+  open Tezos_micheline;
 
   let hash_packed_data = data =>
     data |> to_bytes |> Bytes.to_string |> BLAKE2B.hash;
@@ -764,7 +765,7 @@ module Consensus = {
     switch (entrypoint, micheline) {
     | (
         "deposit",
-        Tezos_micheline.Micheline.Prim(
+        Micheline.Prim(
           _,
           Michelson_v1_primitives.D_Pair,
           [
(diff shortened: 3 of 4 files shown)


anchpop (approved):

bin/http_server.re
```diff
@@ -177,10 +177,36 @@ let handle_ticket_balance =
...
+             | Some(address) => address
+             | None =>
+               failwith(
+                 Printf.sprintf(
+                   "Could not parse %s received from the contract into an address",
+                   validator,
+                 ),
+               )
+             },
+             Printf.sprintf("https://localhost:444%d", i) |> Uri.of_string,
```
callistonianembrace, line 204: Opened a follow up issue to keep this PR self contained
https://github.com/marigold-dev/sidechain/issues/160

tezos_interop/tezos_interop.rei
```diff
@@ -154,3 +154,9 @@ module Consensus: {
 };
 
 module Discovery: {let sign: (Secret.t, ~nonce: int64, Uri.t) => Signature.t;};
+
+module Fetch_validators: {
+  let run:
+    (~rpc_node: string, ~contract_address: string) =>
+    Lwt.t(result(list(string), string));
+};
```
EduardoRFS, line 162: This should not exported, look how the other interops with taquito do. TLDR, it should be exported as part of the Consensus module and not use strings for types.

tezos_interop/tezos_interop.re
```diff
@@ -539,10 +539,10 @@ module Fetch_validators: {
       rpc_node: Uri.to_string(rpc_node),
       confirmation,
       contract_address: Address.to_string(contract_address),
-    };
+      };
```
EduardoRFS, line 542: Not needed spaces

tezos_interop/tezos_interop.re
```diff
@@ -539,10 +539,10 @@ module Fetch_validators: {
...
     // TODO: stop hard coding this
     let command = "node";
-    let.await file =
+      let.await file =
```
EduardoRFS, line 545: Not needed spaces

tezos_interop/tezos_interop.rei
```diff
@@ -151,6 +153,8 @@ module Consensus: {
   };
   let listen_operations:
     (~context: Context.t, ~on_operation: operation => unit) => unit;
+  let fetch_validators:
+    (~context: Context.t) => Lwt.t(result(list(Secret.t), string));
```
EduardoRFS, line 157:
```suggestion
    (~context: Context.t) => Lwt.t(result(list(Key.t), string));
```

There is no way the storage can return `Secret.t`, as they're never stored on Tezos to begin with, probably this should be Key.t

tezos_interop/tezos_interop.re
```diff
@@ -748,6 +790,52 @@ module Consensus = {
...
+        }
+      | Ok(_) => Error("Failed to parse storage micheline expression")
+      | Error(msg) => Error(msg);
+    let.await micheline_storage =
+      Fetch_validators.run(
+        ~confirmation=required_confirmations,
+        ~rpc_node,
+        ~contract_address=consensus_contract,
+      );
+    Lwt.return @@ micheline_to_validators(micheline_storage);
```
EduardoRFS, line 837:
```suggestion
    Lwt.return(micheline_to_validators(micheline_storage));
```

tezos_interop/tezos_interop.re
```diff
@@ -748,6 +790,52 @@ module Consensus = {
...
+        ) => {
+          Ok(
+            List.map(
+              fun
+              | Micheline.String(_, sk) =>
+                switch (Secret.of_string(sk)) {
+                | Some(sk) => sk
+                | None => failwith("sk isn't a valid Secret")
+                }
+              | _ => failwith("Should have been String"),
```
EduardoRFS, line 824: This should not be a failwith, I think there is a `fold_left_ok` if you want you can do a make `map_ok`

EduardoRFS (changes requested):

tezos_interop/tezos_interop.re
callistonianembrace: Didn't get you - what's to be done with keys that dont parse?

EduardoRFS: return an Error()

callistonianembrace: We could still use the keys that parse - what about a filter_map?

EduardoRFS: If a single key failed then everything needs to fail.

tezos_interop/tezos_interop.re
```diff
@@ -806,6 +806,51 @@ module Consensus = {
...
+                _,
+                D_Pair,
+                [_, Prim(_, D_Pair, [_, Prim(_, D_Pair, keys, _)], _)],
+                _,
+              ),
+            ],
+            _,
+          ),
+        ) => {
+          List.fold_left(
```
EduardoRFS, line 829: You should use `fold_left_ok` here

EduardoRFS (changes requested):

tezos_interop/fetch_storage.js
```diff
@@ -60,5 +60,7 @@ const error = (error) =>
     } else {
       error({ message: "Not in current Branch" });
     }
+  } else {
+    error({ message: "Internal error: operationFromHead was undefined" });
```
EduardoRFS, line 64: That's a nice catch

tezos_interop/tezos_interop.re
```diff
@@ -555,29 +555,26 @@ module Fetch_storage: {
...
+      let.await output =
+        Lwt_process.pmap(
+          (command, [|command, file|]),
+          Yojson.Safe.to_string(input_to_yojson(input)),
+        );
+      switch (Yojson.Safe.from_string(output) |> output_of_yojson) {
+      | Ok(storage) => await(Ok(storage))
+      | Error(error) => await(Error(error))
+      };
+    });
```
EduardoRFS, line 577: I don't think this change should be here, also I noticed a problem that this file is being created on every execution and not in a global level.

EduardoRFS (changes requested):

tezos_interop/tezos_interop.re
callistonianembrace: Addressed by rebasing in the next commit

tezos_interop/tezos_interop.rei
```diff
@@ -102,6 +102,8 @@ module Context: {
     required_confirmations: int,
   };
 };
+
+type michelson = Micheline.node(int, Michelson_v1_primitives.prim);
```
EduardoRFS, line 106:
```suggestion
```

tezos_interop/tezos_interop.re
```diff
@@ -808,6 +809,42 @@ module Consensus = {
...
+            [Prim(_, D_Pair, [_, Seq(_, keys)], _), _, _],
+            _,
+          ),
+        ) => {
+          fold_left_ok(
+            (acc, k) =>
+              switch (k) {
+              | Micheline.String(_, k) =>
+                switch (Key.of_string(k)) {
+                | Some(k) => Ok([k, ...acc])
```
EduardoRFS, line 829: This reverts the order of validators, and it shouldn't, the order of validators is quite important for the consensus, as it's how we decide on the next validator.

EduardoRFS (changes requested):

tezos_interop/tezos_interop.rei
callistonianembrace: What do you want here exactly?

callistonianembrace: Attempted addressing this in the incoming commit

tezos_interop/tezos_interop.re
callistonianembrace: Added a List.rev

tezos_interop/tezos_interop.re
```diff
@@ -507,8 +507,6 @@ module Run_contract = {
   };
 };
 
-type michelson =
-  Tezos_micheline.Micheline.node(int, Michelson_v1_primitives.prim);
```
EduardoRFS, line 511: Why did you remove this? Now the type is being duplicated in 3 places, please revert this change

EduardoRFS (changes requested):

tezos_interop/tezos_interop.re
callistonianembrace: https://github.com/marigold-dev/sidechain/pull/158#discussion_r705558502

EduardoRFS (approved):
anime girl happy with the PR getting merged

PR #162, opened 2021-08-17
Adds *.bundle.js.LICENSE.txt to .gitignore

callistonianembrace:
**Problem**
*.bundle.LICENSE.txt files keep showing up as untracked files in the VCS

**Solution**
This PR adds them to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules
 data
 *.install
 *.bundle.js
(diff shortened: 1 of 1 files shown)


EduardoRFS (approved):
Not sure if we should ignore this problem, but I agree it's annoying so I opened a follow up. #166

PR #173, opened 2021-08-23
Docs: Extract developer documentation as a separate wiki

callistonianembrace:
**Problem**
We include developer documentation in the project Readme. Readme as often distributed along with the source and contain minimal necessary information to run the project. Editing Readme have to go through PR workflows, which is can be slow, esp. documenting sub topics a like working with test-networks or tip and tricks that are optional and help others but not necessarily the official word on the topic

**Solution**
Maintain comprehensive developer documentation that can help onboard new developers and interested contributors, as a separate wiki. To de-deduplicate, this PR removes non-essential content from the Readme.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,23 +76,4 @@ It will start all your nodes, produce a block and sign it, this will start the c
 
 ## Contributing
 
-### Guidelines
-
-Please consider the following,
-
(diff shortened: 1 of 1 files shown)


anchpop (approved):
Hmm, I wish there was a way to have the wiki automatically copy some .md files, but this is good too

PR #175, opened 2021-08-27
introduce past_state_root_hashes to track if node is in sync

callistonianembrace:
**Problem**
We don't have a way to know if the node is in sync with the mainchain

**Solution**
This PR (along with follow up PRs), introduce routine checks that make sure newly published state_root_hash on the mainchain are a part of the node's history of computed state_root_hashes. Specifically, this PR introduces a rolling list of such hashes that are looked into.

### DIFF
--- a/node/state.re
+++ b/node/state.re
@@ -235,6 +235,7 @@ let load_snapshot =
 
   // TODO: this is clearly an abstraction leak
 
+  let past_state_root_hashes = [];
   let protocol =
     Protocol.{
       ledger,
@@ -244,6 +245,7 @@ let load_snapshot =
       validators_hash,
       block_height,
       last_block_hash,
+      past_state_root_hashes,
       state_root_hash,
       last_state_root_update: 0.0,
       last_applied_block_timestamp: 0.0,
--- a/protocol/protocol.re
+++ b/protocol/protocol.re
@@ -147,6 +147,12 @@ let apply_block = (state, block) => {
         block.state_root_hash != state.state_root_hash
           ? Unix.time() : state.last_state_root_update,
       last_applied_block_timestamp: Unix.time(),
+      past_state_root_hashes:
+        if (state.state_root_hash != block.state_root_hash) {
+          [state.state_root_hash, ...state.past_state_root_hashes];
+        } else {
+          state.past_state_root_hashes;
+        },
       state_root_hash: block.state_root_hash,
       validators_hash: block.validators_hash,
     },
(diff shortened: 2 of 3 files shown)


protocol/protocol.re
```diff
@@ -147,6 +147,10 @@ let apply_block = (state, block) => {
         block.state_root_hash != state.state_root_hash
           ? Unix.time() : state.last_state_root_update,
       last_applied_block_timestamp: Unix.time(),
+      past_state_root_hashes: [
+        state.state_root_hash,
```
EduardoRFS, line 151: This should only be added if the state_root_hash is different than the previous one

protocol/state.re
```diff
@@ -16,6 +16,9 @@ type t = {
   // shared consensus
   block_height: int64,
   last_block_hash: BLAKE2B.t,
+  // List of state_root_hash'es computed by this node from a last known good state
+  // This will later be validated against the main chain
```
EduardoRFS, line 20: Nonblocking: not sure the wording here is right, 

> validated against the main chain

EduardoRFS (changes requested):

protocol/state.re
callistonianembrace: Is `validated` the problematic word? I'm not sure what's better here

EduardoRFS:
@d4hines I think your change will inevitably lead to this being fixed right? I believe we can close this PR then.

d4hines:
Correct - when #195 is done this should be covered.

d4hines:
After talking it through with @callistonianembrace, we realized that there are two ways a node can be out of sync:
- If a node is starting out from scratch and doesn't know for certain the current state. In this case it needs to follow the procedure outlined in https://hackmd.io/5FN2ukMYRGWEvlSC9XFmWQ.
- If a node has fallen behind in async state hashing (not yet implemented, will be in #195 or child PR's). In this case, it needs to follow the procedures outlined in that PR description.

My work will only address the second issue for now. It might make sense to coordinate the changes, or it might make sense to keep them independent - not sure.

EduardoRFS:
Can you explain again what does this PR achieves?

d4hines:
#234 might be relevant to the implementation of this PR. Specifically see https://github.com/marigold-dev/deku/pull/234/commits/749d273bb704ff3c718f83c02a8acf52d28f64b5

node/state.re
```diff
@@ -235,6 +235,7 @@ let load_snapshot =
 
   // TODO: this is clearly an abstraction leak
 
+  let past_state_root_hashes = [];
```
d4hines, line 238: It seems this list will only ever grow, which is problematic. Every data structure should have some kind of upper limit.

d4hines (changes requested):
See my comment. Also, let's talk about what #234 adds to the mix.

d4hines:
I believe we should close this PR. We already download snapshots and correctly stop the chain if they provide bad blocks on top of the snapshot. The thing we're missing is fetching the most recently published state root hash from chain storage and confirming the snapshot sent corresponds to this hash. This PR doesn't address this - I think the concerns this PR would address are already implemented in our current syncing mechanism.

PR #176, opened 2021-08-31
Parse `update_root_hash` operation  from the mainchain

callistonianembrace:
## Problem

We don't have a way to know if the node is in sync with the mainchain

## Solution

This PR (along with #175 ), adds the ability to parse `update_root_hash` operations, so that updates to state root hash published to the main chain can be listened in, which would help a node know if it is in sync with main chain or not.

### DIFF
--- a/tezos_interop/tezos_interop.re
+++ b/tezos_interop/tezos_interop.re
@@ -754,7 +754,8 @@ module Consensus = {
         // TODO: proper type for amounts
         amount: Z.t,
         destination: Address.t,
-      });
+      })
+    | Update_root_hash(BLAKE2B.t);
   type operation = {
     hash: BLAKE2B.t,
     index: int,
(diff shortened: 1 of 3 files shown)


tezos_interop/tezos_interop.re
```diff
@@ -693,43 +693,100 @@ module Consensus = {
...
-        Data_encoding.Binary.of_bytes_opt(Address.encoding, ticketer);
-      let ticket = Ticket.{ticketer, data};
-      Some(Deposit({ticket, destination, amount}));
-    | _ => None
-    };
+            )
+          ),
+        ) =>
+        switch (state_root_hash |> Bytes.to_string |> BLAKE2B.of_raw_string) {
+        | Some(state_root_hash) => Some(Update_root_hash(state_root_hash))
```
EduardoRFS, line 756: As you already extracted all the data, I would ask you to make a record like it's doing on deposit

EduardoRFS (changes requested):

anchpop (approved):

tezos_interop/tezos_interop.re
callistonianembrace: Not sure I understand - you mean this `Some(Update_root_hash({ state_root_hash } ))` ? What do we gain?

EduardoRFS: `Update_root_hash` is not only `state_root_hash`, I'm talking about all the extracted data, not only state_root_hash, again, look how it's doing on deposits.

callistonianembrace: As the other properties are not relevant to the scope of the PR, which is to merely extract out `state_root_hash` so that in a future PR, a lookup can be made to know if the node is in a sync state or not, can we extract the other properties in a future PR?

Adding new properties will not add more things to test in this PR, thereby increasing its scope.

EduardoRFS: @callistonianembrace feel free to open a follow up issue, the idea is that this variant should be 1 : 1 with the variant encoded in the consensus.mligo.

callistonianembrace: Created a follow up issue here - https://github.com/marigold-dev/sidechain/issues/181

EduardoRFS (approved):

EduardoRFS (approved):
anime girl surprised with the PR finished

PR #178, opened 2021-09-06
Avoids broadcasting operations from sidecli

### LINKED ISSUE #70: ensure that client talks to a single node
The client should never talk to multiple nodes unless something hackish like `inject-genesis` is happening.

callistonianembrace:
## Problem

Clients like sidecli broadcast the network operations themselves. 

## Solution

This PR directly posts operations to `/operation-gossip` endpoint as using http endpoints meant inter-node communication can also be used by sidecli

## Related

- closes #70

### DIFF
--- a/bin/sidecli.re
+++ b/bin/sidecli.re
@@ -184,12 +184,14 @@ let create_transaction =
       ~ticket,
       ~kind=Transaction({destination: received_address}),
     );
+  let file = folder_node ++ "/identity.json";
+  let.await identity = Files.Identity.read(~file);
 
   // Broadcast transaction
   let.await () =
-    Networking.broadcast_operation_gossip_to_list(
-      validators_uris,
+    Networking.request_operation_gossip(
       Networking.Operation_gossip.{operation: transaction},
+      identity.uri,
     );
   Format.printf(
     "operation.hash: %s\n%!",
@@ -265,9 +267,9 @@ let info_withdraw = {
 let withdraw =
     (folder_node, sender_wallet_file, tezos_address, amount, ticket) => {
   open Networking;
-  let.await validators_uris = validators_uris(folder_node);
-  let validator_uri = List.hd(validators_uris);
-  let.await block_level_response = request_block_level((), validator_uri);
+  let file = folder_node ++ "/identity.json";
+  let.await identity = Files.Identity.read(~file);
+  let.await block_level_response = request_block_level((), identity.uri);
   let block_level = block_level_response.level;
   let.await wallet = Files.Wallet.read(~file=sender_wallet_file);
   let operation =
@@ -283,9 +285,9 @@ let withdraw =
 
   // Broadcast transaction
   let.await () =
-    Networking.broadcast_operation_gossip_to_list(
-      validators_uris,
+    Networking.request_operation_gossip(
       Networking.Operation_gossip.{operation: operation},
+      identity.uri,
     );
 
   Format.printf("operation.hash: %s\n%!", BLAKE2B.to_string(operation.hash));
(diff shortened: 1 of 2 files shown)


EduardoRFS:
This endpoint is already implemented https://github.com/marigold-dev/sidechain/blob/b1de8a6725c1cd3955a26acaa9cab726c1e5f6f6/bin/http_server.re#L167

callistonianembrace:
@EduardoRFS Fixed to re-use `/operation-gossip` endpoint :+1:

EduardoRFS:
@callistonianembrace this is not a valid change, what if the validator 0 is offline?

callistonianembrace:
> @callistonianembrace this is not a valid change, what if the validator 0 is offline?

Earlier, I was expecting a more noisy PR. The solution I had in mind was to explicity accept `--endpoint` cmdline arg. I'll add a commit to address this right away in this PR as it's a much simpler PR now

bin/sidecli.re
```diff
@@ -167,7 +167,14 @@ let info_create_transaction = {
 };
 
 let create_transaction =
-    (folder_node, sender_wallet_file, received_address, amount, ticket) => {
+    (
+      folder_node,
+      endpoint,
```
EduardoRFS, line 172: This field is not needed, you're already passing `folder_node` you can access the `identity.json` and get the URI there.

bin/sidecli.re
```diff
@@ -262,12 +276,9 @@ let info_withdraw = {
   Term.info("withdraw", ~version="%‌%VERSION%%", ~doc, ~exits, ~man);
 };
 
-let withdraw =
-    (folder_node, sender_wallet_file, tezos_address, amount, ticket) => {
+let withdraw = (endpoint, sender_wallet_file, tezos_address, amount, ticket) => {
```
EduardoRFS, line 279: I would say to add a `folder_node` like the other commands, especially as the sidecli is mostly target as development.

EduardoRFS (changes requested):

bin/sidecli.re
```diff
@@ -199,6 +206,12 @@ let create_transaction =
   Lwt.return(`Ok());
 };
 
+let endpoint_uri = {
```
anchpop, line 209: should be changed to `transaction_endpoint_uri` or `doc` should be changed not to specify that it's used for a transaction (possibly make `endpoint_uri` take a parameter?)

anchpop (approved):

bin/sidecli.re
callistonianembrace: Addressed in the incoming commit

bin/sidecli.re
callistonianembrace: Addressed in the incoming commit

bin/sidecli.re
callistonianembrace: I have incorporated Eduardo's suggestion ie using identity.json

bin/sidecli.re
```diff
@@ -265,9 +267,9 @@ let info_withdraw = {
 let withdraw =
     (folder_node, sender_wallet_file, tezos_address, amount, ticket) => {
   open Networking;
-  let.await validators_uris = validators_uris(folder_node);
-  let validator_uri = List.hd(validators_uris);
-  let.await block_level_response = request_block_level((), validator_uri);
+  let file = folder_node ++ "/identity.json";
+  let.await identity = Files.Identity.read(~file);
```
anchpop, line 271: it's not really in the scope of the PR but at some point we should write a function that combines these (and also uses ` Filename.concat`)

anchpop (approved):

EduardoRFS (approved):
anime girl approves this PR

PR #179, opened 2021-09-07
create fetch_storage script on boot only

EduardoRFS:
## Problem

Currently the `Fetch_storage.run` function is creating a temporary file on every call, this file is identical for every execution so it should be created only once.

## Solution

Move it to the global scope of the module, so that it's created during the boot of the application

## Related

- #157

### DIFF
--- a/tezos_interop/tezos_interop.re
+++ b/tezos_interop/tezos_interop.re
@@ -556,19 +556,21 @@ module Fetch_storage: {
     };
   };
 
+  // TODO: stop hard coding this
+  let command = "node";
+  let file = {
+    let.await (file, oc) = Lwt_io.open_temp_file(~suffix=".js", ());
+    let.await () = Lwt_io.write(oc, [%blob "fetch_storage.bundle.js"]);
+    await(file);
+  };
+  let file = Lwt_main.run(file);
+
   let _run = (~rpc_node, ~confirmation, ~contract_address) => {
     let input = {
       rpc_node: Uri.to_string(rpc_node),
       confirmation,
       contract_address: Address.to_string(contract_address),
     };
-    // TODO: stop hard coding this
-    let command = "node";
-    let.await file =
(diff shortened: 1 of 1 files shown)


tezos_interop/tezos_interop.re
```diff
@@ -556,19 +556,21 @@ module Fetch_storage: {
     };
   };
 
+  // TODO: stop hard coding this
+  let command = "node";
```
callistonianembrace, line 560: This  doesn't look right -  the temp file will disappear after the `with_temp_file` finishes . Does this PR intend to keep it for entire life time of the node?

EduardoRFS: No it doesn't, the temp file is kept there, the `with_temp_file` here is above keeping the temp_file open, not about keeping it present. This is already how it's done for other scripts, otherwise the scripts wouldn't work.

callistonianembrace: https://ocsigen.org/lwt/5.2.0/api/Lwt_io#VALwith_temp_file

> When the promise returned by f is resolved, with_temp_file closes the channel and deletes the temporary file by calling Lwt_unix.unlink. Since 3.2.0

callistonianembrace: This was fixed today.

EduardoRFS: Oh then we're abusing of a bug, that's nice, will fix both branches

EduardoRFS: Nope, I was wrong, I'm not using with_temp_file on the other code, mb

EduardoRFS: thank you fixed it

callistonianembrace (approved):

anchpop (approved):

EduardoRFS:
Oh, I love a fast review
anime girl proud of fast review

d4hines (commented):
LGTM

PR #182, opened 2021-09-07
Proto: Adds support for add/remove validators in the sidechain

callistonianembrace:
## Depends

- [x] #184 
- [x] #185 
- [x] #187 

## Problem

Adding/removing validators is missing right now

## Solution

This PR adds protocol level support for submit addition and removal of validators

### DIFF
--- a/node/building_blocks.re
+++ b/node/building_blocks.re
@@ -63,11 +63,40 @@ let is_signable = (state, block) => {
     state.State.pending_main_ops |> List.exists(op => op == main_op);
   let all_main_ops_are_known =
     List.for_all(is_known_main, block.Block.main_chain_ops);
+  let {
+    Node.trusted_validator_membership_change,
+    protocol: {last_seen_membership_change_timestamp, _},
+    _,
+  } = state;
+  let current_time = Unix.time();
+  // TODO: this should not be hard coded here
+  let next_allowed_membership_change_timestamp =
+    last_seen_membership_change_timestamp +. 24. *. 60. *. 60.;
+
+  let contains_only_trusted_add_validator_op =
+    List.for_all(h => {
+      switch (h.Operation.Side_chain.kind) {
+      | Add_validator(validator) =>
+        Trusted_validators_membership_change.Set.mem(
+          {address: validator.address, action: Add},
+          trusted_validator_membership_change,
+        )
+        && current_time > next_allowed_membership_change_timestamp
+      | Remove_validator(validator) =>
+        Trusted_validators_membership_change.Set.mem(
+          {address: validator.address, action: Remove},
+          trusted_validator_membership_change,
+        )
+        && current_time > next_allowed_membership_change_timestamp
+      | _ => true
+      }
+    });
   is_next(state, block)
   && !is_signed_by_self(state, ~hash=block.hash)
   && is_current_producer(state, ~key=block.author)
   && !has_next_block_to_apply(state, ~hash=block.hash)
-  && all_main_ops_are_known;
+  && all_main_ops_are_known
+  && contains_only_trusted_add_validator_op(block.side_chain_ops);
 };
 
 let sign = (~key, block) => Block.sign(~key, block);
--- a/node/state.re
+++ b/node/state.re
@@ -31,7 +31,14 @@ type t = {
   // TODO: use proper variants in the future
   // TODO: this also needs to be cleaned in the future
   recent_operation_results:
-    BLAKE2B.Map.t([ | `Transaction | `Withdraw(Ledger.Handle.t)]),
+    BLAKE2B.Map.t(
+      [
+        | `Add_validator
+        | `Remove_validator
+        | `Transaction
+        | `Withdraw(Ledger.Handle.t)
+      ],
+    ),
 };
 
 let make =
@@ -255,6 +262,7 @@ let load_snapshot =
       state_root_hash,
       last_state_root_update: 0.0,
       last_applied_block_timestamp: 0.0,
+      last_seen_membership_change_timestamp: 0.0,
     };
   let.ok protocol =
     fold_left_ok(
--- a/protocol/operation.re
+++ b/protocol/operation.re
@@ -55,7 +55,10 @@ module Side_chain = {
   [@deriving yojson]
   type kind =
     | Transaction({destination: Wallet.t})
-    | Withdraw({owner: Tezos_interop.Address.t});
+    | Withdraw({owner: Tezos_interop.Address.t})
+    | Add_validator(Validators.validator)
+    | Remove_validator(Validators.validator);
+
   [@deriving yojson]
   type t = {
     hash: BLAKE2B.t,
--- a/protocol/protocol.re
+++ b/protocol/protocol.re
@@ -58,28 +58,36 @@ let apply_side_chain = (state: t, operation) => {
 
   // apply operation
   let included_operations = Set.add(operation, state.included_operations);
+  let state = {...state, included_operations};
 
   let {source, amount, ticket, _} = operation;
-  let.ok (ledger, result) =
-    switch (operation.kind) {
-    | Transaction({destination}) =>
-      let.ok ledger =
-        Ledger.transfer(~source, ~destination, amount, ticket, state.ledger);
-      Ok((ledger, `Transaction));
-    | Withdraw({owner}) =>
-      let.ok (ledger, handle) =
-        Ledger.withdraw(
-          ~source,
-          ~destination=owner,
-          amount,
-          ticket,
-          state.ledger,
-        );
-      // TODO: publish the handle somewhere
-      Ok((ledger, `Withdraw(handle)));
-    };
-
-  Ok(({...state, ledger, included_operations}, result));
+  let update_validators = validators => {
+    let last_seen_membership_change_timestamp = Unix.time();
+    {...state, validators, last_seen_membership_change_timestamp};
+  };
+  switch (operation.kind) {
+  | Transaction({destination}) =>
+    let.ok ledger =
+      Ledger.transfer(~source, ~destination, amount, ticket, state.ledger);
+    Ok(({...state, ledger}, `Transaction));
+  | Withdraw({owner}) =>
+    let.ok (ledger, handle) =
+      Ledger.withdraw(
+        ~source,
+        ~destination=owner,
+        amount,
+        ticket,
+        state.ledger,
+      );
+    // TODO: publish the handle somewhere
+    Ok(({...state, ledger}, `Withdraw(handle)));
+  | Add_validator(validator) =>
+    let validators = Validators.add(validator, state.validators);
+    Ok((update_validators(validators), `Add_validator));
+  | Remove_validator(validator) =>
+    let validators = Validators.remove(validator, state.validators);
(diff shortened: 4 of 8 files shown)


anchpop (approved):

callistonianembrace:
@EduardoRFS Commits re-ordered

EduardoRFS:
Fix the list of dependencies on the PR, and rebase this, the patch seems good tho.

d4hines:
Quick scan of the PR looks fine to me.

However, I'm a bit concerned what this new property will do to consensus. In principle if we can verify the consensus of N validators, then we should be able to verify it when N changes (perhaps with certain constraints around when it's allowed to change). However, in practice I'm worried this will introduce more places for us to forget to check. The project is in a vulnerable state right now - not much unit or integration tests, let alone regression tests or formal verification. My two cents would be to hold off on this feature until we at least have a better specification for consensus (right now we don't even have a document that says what properties need to be checked, let alone a way of checking them).

callistonianembrace:
This PR could definitely use more unit tests as @EduardoRFS has always been writing.

d4hines:
> This PR could definitely use more unit tests as @EduardoRFS has always been writing.

That's not my issue. While unit tests always help to add confidence, I'm not doubting the correctness of your implementation of the feature. I'm doubting the safety of the feature in combination with everything else. This isn't something we can test because we haven't specified the global properties that the chain must have. For example, what constitute safety and liveness in our case? Different consensus algorithms guarantee different things, and I think we should spend some time specifying ours. Once we know what we're trying to guarantee, it will be easier to convince ourselves we indeed meet those guarantees.

protocol/protocol.re
```diff
@@ -60,26 +60,35 @@ let apply_side_chain = (state: t, operation) => {
...
+        ticket,
+        state.ledger,
+      );
+    // TODO: publish the handle somewhere
+    Ok(({...state, ledger, included_operations}, `Withdraw(handle)));
+  | Add_validator(validator) =>
+    let validators = Validators.add(validator, state.validators);
+    Ok((
+      {...state, validators, included_operations},
+      `Add_validator(validator),
```
callistonianembrace, line 83: Having `validator` parameter in the polyvar `Add_validator` doesn't seem useful. I think we can get rid of it

node/building_blocks.re
```diff
@@ -59,30 +59,45 @@ let is_current_producer = (state, ~key) => {
...
+      | Add_validator(validator, timestamp) =>
         if (Trusted_validators_membership_change.(
               Set.mem(
                 {address: validator.address, action: Add},
-                trusted_validators,
+                trusted_validator_membership_change,
               )
-            )) {
+            )
+            && timestamp > current_time
```
callistonianembrace, line 84: Should this be `>=`? Like when a node cli sends a Add/Remove operation to itself

node/building_blocks.re
```diff
@@ -58,6 +58,53 @@ let is_current_producer = (state, ~key) => {
 // TODO: bad naming
 // TODO: check if block must have published a new snapshot
 let is_signable = (state, block) => {
+  open Protocol.Operation.Side_chain;
+  let Node.{
+        trusted_validator_membership_change,
+        protocol: {last_seen_membership_change_timestamp, _},
+        _,
+      } = state;
+  let allowed_op_reception_delay = 0.200; // Assumed max network latency
```
EduardoRFS, line 67: The max network latency is closer to a second

node/building_blocks.re
```diff
@@ -58,6 +58,53 @@ let is_current_producer = (state, ~key) => {
...
+  let current_time = Unix.time();
+  let allowed_op_window = current_time +. allowed_op_reception_delay;
+  let next_allowed_membership_change_timestamp =
+    last_seen_membership_change_timestamp +. 24. *. 60. *. 60.;
+  let rec contains_only_trusted_add_validator_op = result_so_far =>
+    fun
+    | [] => result_so_far
```
EduardoRFS, line 74: Use `List.for_all`

node/building_blocks.re
```diff
@@ -58,6 +58,53 @@ let is_current_producer = (state, ~key) => {
...
+        if (Trusted_validators_membership_change.(
+              Set.mem(
+                {address: validator.address, action: Remove},
+                trusted_validator_membership_change,
+              )
+            )
+            && timestamp >= current_time
+            && timestamp <= allowed_op_window
+            && timestamp > next_allowed_membership_change_timestamp) {
+          contains_only_trusted_add_validator_op(result_so_far, t);
```
EduardoRFS, line 101: This logic is duplicated from the case above and it's a really  important piece of domain logic, please abstract it in some meaningful function

node/building_blocks.re
```diff
@@ -67,7 +114,11 @@ let is_signable = (state, block) => {
   && !is_signed_by_self(state, ~hash=block.hash)
   && is_current_producer(state, ~key=block.author)
   && !has_next_block_to_apply(state, ~hash=block.hash)
-  && all_main_ops_are_known;
+  && all_main_ops_are_known
+  && contains_only_trusted_add_validator_op(
+       true,
```
EduardoRFS, line 119: This doesn't make sense here.

node/snapshots.re
```diff
@@ -11,16 +11,19 @@ type t = {
   last_block: Block.t,
   last_block_signatures: Signatures.t,
   additional_blocks: list(Block.t),
+  last_seen_membership_change_timestamp: float,
```
EduardoRFS, line 14: Why was this added to the snapshots?

bin/deku_node.re
```diff
@@ -140,6 +140,8 @@ let handle_protocol_snapshot =
         last_block: snapshots.last_block,
         last_block_signatures:
           Signatures.to_list(snapshots.last_block_signatures),
+        last_seen_membership_change_timestamp:
+          snapshots.last_seen_membership_change_timestamp,
```
EduardoRFS, line 144: This is not okay, you cannot send a timestamp from one validator to another.

protocol/operation.re
```diff
@@ -52,10 +52,16 @@ module Main_chain = {
...
   // TODO: I don't like this structure model
+  [@deriving yojson]
+  type timestamp = float;
+
   [@deriving yojson]
   type kind =
     | Transaction({destination: Wallet.t})
-    | Withdraw({owner: Tezos_interop.Address.t});
+    | Withdraw({owner: Tezos_interop.Address.t})
+    | Add_validator(Validators.validator, timestamp)
```
EduardoRFS, line 62: This should not have a timestamp

protocol/operation.re
```diff
@@ -52,10 +52,16 @@ module Main_chain = {
...
+    | Remove_validator(Validators.validator, timestamp);
```
EduardoRFS, line 63: Also I would say those two should be a single operation as every comment will always apply to both of them.

protocol/protocol.re
EduardoRFS: Agreed, if it is useful in the feature we can add it back.

protocol/protocol.re
```diff
@@ -60,26 +60,47 @@ let apply_side_chain = (state: t, operation) => {
...
+    Ok((
+      {
+        ...state,
+        validators,
+        last_seen_membership_change_timestamp,
+        included_operations,
+      },
+      `Remove_validator(validator),
+    ));
+  };
```
EduardoRFS, line 103: This logic is duplicated from Add_validator

protocol/state.re
```diff
@@ -13,6 +13,7 @@ type t = {
   // consensus
   validators: Validators.t,
   validators_hash: BLAKE2B.t,
+  last_seen_membership_change_timestamp: float,
```
EduardoRFS, line 16: That's not right, this is part of the local consensus data, together with `last_state_root_update` and `last_applied_block_timestamp`

EduardoRFS (changes requested):

node/building_blocks.re
callistonianembrace: Can you please elaborate?

protocol/protocol.re
callistonianembrace: Addressed in the next commit

node/building_blocks.re
callistonianembrace: Addressed by getting rid of it - timestamps are no longer a part of the operation design now.

node/building_blocks.re
callistonianembrace: Addressed in the next commit

node/building_blocks.re
callistonianembrace: Logic now simpler in the next commit now that timestamp is no longer  a part of operation and doesn't have to travel over the wire.

node/building_blocks.re
callistonianembrace: Moved to `Node.apply_block`, before `Protocol.apply_block`

node/snapshots.re
callistonianembrace: So that all the validators have a common notion of `last_seen_membership_change_timestamp`

protocol/operation.re
callistonianembrace: Addressed in the next commit by removing the timestamp from the operation.

protocol/protocol.re
callistonianembrace: abstracted into a single function. For now, Add/Remove are separate operations.

protocol/operation.re
callistonianembrace: Let's revisit this again at the end.

protocol/state.re
callistonianembrace: I don't understand this comment - are you saying `last_seen_membership_change_timestamp` should be moved to node local state? `last_state_root_update` and `last_applied_block_timestamp` are currently a part of protocol state and not node - are you saying they should be moved too?

callistonianembrace: Thanks for explaining offline. Making the changes in the next commit

node/snapshots.re
callistonianembrace: As explained offline, since this property is local consensus, ie values other nodes need not agree, I will be removing them from the snapshots.

node/state.re
```diff
@@ -117,8 +124,46 @@ let try_to_commit_state_hash = (~old_state, state, block, signatures) => {
...
+          && current_time > next_allowed_membership_change_timestamp
+        | _ => true
+        }
+      })
+    );
+
+  let.assert () = (
+    `Invalid_block_when_applying,
+    contains_only_trusted_add_validator_op(block.side_chain_ops),
+  );
```
EduardoRFS, line 166: This entire checking should not be here, this is a property of a signable block, you're not allowed to sign a block that contains a trusted validator change that you don't know, but you should still apply the block if 2/3 of the validators sign it.

EduardoRFS (changes requested):

node/state.re
callistonianembrace: Fair point! How about moving this to `is_signable` ? Attempted this in the next incoming commit.

EduardoRFS:
Looks okay to me, made some changes at https://github.com/marigold-dev/deku/pull/175, as soon as the CI is green I merge

EduardoRFS (approved):
deku vibing hard with the new feature

PR #184, opened 2021-09-07
Adds trusted_validators in the node-local state

callistonianembrace:
## Problem

Add_validator operation needs to be verified against a list of trusted validators that a node operator maintains/obtains off-chain.

## Solution

This PR adds a new entry in the node state - `trusted_validators` - which a node sources from `trusted-validators.json`. This files is expected to contain keys and URIs of validators that a node operation doesn't mind being added to the chain as a validator in the future

### DIFF
--- a/bin/deku_node.re
+++ b/bin/deku_node.re
@@ -179,6 +179,14 @@ let handle_ticket_balance =
 let node = {
   let folder = Sys.argv[1];
   let.await identity = Files.Identity.read(~file=folder ++ "/identity.json");
+  let.await trusted_validator_membership_change_list =
+    Files.Trusted_validators_membership_change.read(
+      ~file=folder ++ "/trusted-validator-membership-change.json",
+    );
+  let trusted_validator_membership_change =
+    Trusted_validators_membership_change.Set.of_list(
+      trusted_validator_membership_change_list,
+    );
   let.await interop_context =
     Files.Interop_context.read(~file=folder ++ "/tezos.json");
   let.await validator_res =
@@ -207,6 +215,7 @@ let node = {
   let node =
     State.make(
       ~identity,
+      ~trusted_validator_membership_change,
       ~interop_context,
       ~data_folder=folder,
       ~initial_validators_uri,
--- a/bin/files.re
+++ b/bin/files.re
@@ -12,14 +12,17 @@ let read_json = (of_yojson, ~file) => {
   | Error(error) => raise(Invalid_json(error))
   };
 };
+
 let write_json = (to_yojson, data, ~file) =>
   Lwt_io.with_file(~mode=Output, file, oc =>
     Lwt_io.write(oc, Yojson.Safe.pretty_to_string(to_yojson(data)))
   );
+
 module Identity = {
   let read = read_json(identity_of_yojson);
   let write = write_json(identity_to_yojson);
 };
+
 module Wallet = {
   [@deriving yojson]
   type t = {
@@ -77,4 +80,12 @@ module State_bin = {
     Lwt_io.with_file(~mode=Input, file, Lwt_io.read_value);
   let write = (protocol, ~file) =>
     Lwt_io.with_file(~mode=Output, file, Lwt_io.write_value(_, protocol));
+};
+
+module Trusted_validators_membership_change = {
+  [@deriving yojson]
+  type t = Node.Trusted_validators_membership_change.t;
+
+  let read = read_json([%of_yojson: list(t)]);
+  let write = write_json([%to_yojson: list(t)]);
 };
--- a/bin/files.rei
+++ b/bin/files.rei
@@ -8,6 +8,7 @@ module Identity: {
   let read: (~file: string) => Lwt.t(identity);
   let write: (identity, ~file: string) => Lwt.t(unit);
 };
+
 module Wallet: {
   type t = {
     address: Wallet.t,
@@ -16,10 +17,12 @@ module Wallet: {
   let read: (~file: string) => Lwt.t(t);
   let write: (t, ~file: string) => Lwt.t(unit);
 };
+
 module Validators: {
   let read: (~file: string) => Lwt.t(list((Address.t, Uri.t)));
   let write: (list((Address.t, Uri.t)), ~file: string) => Lwt.t(unit);
 };
+
 module Interop_context: {
   let read: (~file: string) => Lwt.t(Tezos_interop.Context.t);
   let write: (Tezos_interop.Context.t, ~file: string) => Lwt.t(unit);
@@ -28,4 +31,12 @@ module Interop_context: {
 module State_bin: {
   let read: (~file: string) => Lwt.t(Protocol.t);
   let write: (Protocol.t, ~file: string) => Lwt.t(unit);
+};
+
+module Trusted_validators_membership_change: {
+  [@deriving yojson]
+  type t = Node.Trusted_validators_membership_change.t;
+
+  let read: (~file: string) => Lwt.t(list(t));
+  let write: (list(t), ~file: string) => Lwt.t(unit);
 };
--- a/node/state.rei
+++ b/node/state.rei
@@ -12,6 +12,7 @@ module Address_map: Map.S with type key = Address.t;
 module Uri_map: Map.S with type key = Uri.t;
 type t = {
   identity,
+  trusted_validator_membership_change: Trusted_validators_membership_change.Set.t,
   interop_context: Tezos_interop.Context.t,
   data_folder: string,
   pending_side_ops: list(Operation.Side_chain.t),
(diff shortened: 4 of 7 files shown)


bin/http_server.re
```diff
@@ -179,6 +179,15 @@ let handle_ticket_balance =
 let node = {
   let folder = Sys.argv[1];
   let.await identity = Files.Identity.read(~file=folder ++ "/identity.json");
+  let.await trusted_validators =
+    Files.Validators.read(~file=folder ++ "/trusted-validators.json")
```
EduardoRFS, line 183: It's not trusted validators, is trusted-new-validators, this is a set of validators who are authorized to join the network once.

EduardoRFS: Also I would ask you to make another file on `Files`

EduardoRFS (changes requested):

bin/http_server.re
anchpop: what does it mean for a validator to be authorized to join the network once? do validators join the network multiple times?

callistonianembrace: @anchpop Before adding/removing a validator, existing validators decided off-chain who is getting booted/added. Each one of them instruct their respective nodes, "Hey, if you see on-chain operation saying add this validator, and it happens to be present in `trusted_validators` accept it. Else reject"

(apologies for anthromorphising things :D )

callistonianembrace:
@EduardoRFS Requested changes have been addressed

bin/files.re
```diff
@@ -48,6 +48,22 @@ module Validators = {
     );
 };
 
+module Trusted_validators = {
+  [@deriving yojson]
+  type t = {address: Address.t};
```
EduardoRFS, line 53: As discussed we also need to store if it's trusted to be added or removed.

EduardoRFS (changes requested):

bin/files.re
callistonianembrace: I think you've addresses the wrong PR. Update actions (adding/removing) happen on PR #185

callistonianembrace:
@EduardoRFS As far as persistence if concerned, this PR is complete. It's only add/remove actions that are not persisted. This can be addressed in PR #185 

Is there anything else to address here?

bin/files.re
EduardoRFS: Yes but this file is designed to be used at #185 so on the current state of the PR this file doesn't move to any meaningful goal.

There is no list of `Trusted_validators`, there is a list of `Trusted_validators_membership_change`

EduardoRFS (changes requested):

bin/files.re
callistonianembrace: > this file doesn't move to any meaningful goal.
It seeds the node-local state

EduardoRFS: Yes but the file is in the wrong format, and it's not created by the setup scripts. 

Add the field containing the kind of membership change.

node/state.rei
```diff
@@ -12,6 +12,7 @@ module Address_map: Map.S with type key = Address.t;
 module Uri_map: Map.S with type key = Uri.t;
 type t = {
   identity,
+  trusted_validators: Validators.t,
```
EduardoRFS, line 15: This should definitely not be a `Validators.t`

EduardoRFS (changes requested):

node/state.rei
callistonianembrace: Addressed in the incoming commit

anchpop (approved):

node/trusted_validators_membership_change.re
```diff
@@ -0,0 +1,45 @@
...
+    let.ok validators = [%of_yojson: list(t)](json);
+    Ok(validators);
+  });
+let write = write_json(validators => validators |> [%to_yojson: list(t)]);
+
+let compare = (x, y) =>
+  Cstruct.compare(
+    Ed25519.pub_to_cstruct(x.address),
+    Ed25519.pub_to_cstruct(y.address),
+  );
```
EduardoRFS, line 39: This compare is not right, what if both have the same address and different action?

node/trusted_validators_membership_change.re
```diff
@@ -0,0 +1,45 @@
+open Helpers;
+open Protocol;
+open Mirage_crypto_ec;
+
+module type T = {
```
EduardoRFS, line 5: Why this module type here?

node/trusted_validators_membership_change.re
```diff
@@ -0,0 +1,45 @@
+open Helpers;
```
EduardoRFS, line 1: Overall I would ask that you keep this module inside of `Files` it's not a big deal also files interfaces should not be used for internal interfaces.

That will lead to some duplication, but the interface generated by yojson is really not ideal.

node/trusted_validators_membership_change.re
```diff
@@ -0,0 +1,45 @@
...
+[@deriving yojson]
+type t = {
+  action,
+  address: Address.t,
+};
+let read =
+  read_json(json => {
+    let.ok validators = [%of_yojson: list(t)](json);
+    Ok(validators);
+  });
```
EduardoRFS, line 32:
```suggestion
  read_json([%of_yojson: list(t)]);
```

node/trusted_validators_membership_change.re
```diff
@@ -0,0 +1,45 @@
...
+let write = write_json(validators => validators |> [%to_yojson: list(t)]);
```
EduardoRFS, line 33:
```suggestion
let write = write_json([%to_yojson: list(t)]);
```

EduardoRFS (changes requested):

node/trusted_validators_membership_change.re
callistonianembrace: Keep one copy in Files and another here

node/trusted_validators_membership_change.re
callistonianembrace: At one point I was using first class modules somewhere, but figured a way without it. The type is irrelevant now and can be removed. Removing..

node/trusted_validators_membership_change.re
callistonianembrace: Doing in fieldwise compare in the next incoming commit

bin/deku_node.re
```diff
@@ -179,6 +179,15 @@ let handle_ticket_balance =
 let node = {
   let folder = Sys.argv[1];
   let.await identity = Files.Identity.read(~file=folder ++ "/identity.json");
+  let.await trusted_validator_membership_change = {
+    Lwt.Infix.
+      (
+        Files.Trusted_validators_membership_change.read(
+          ~file=folder ++ "/trusted-validator-membership-change.json",
+        )
+        >|= Trusted_validators_membership_change.Set.of_list
```
EduardoRFS, line 188: please don't use infix operators

bin/files.re
```diff
@@ -78,3 +81,14 @@ module State_bin = {
   let write = (protocol, ~file) =>
     Lwt_io.with_file(~mode=Output, file, Lwt_io.write_value(_, protocol));
 };
+
+module Trusted_validators_membership_change = {
+  [@deriving yojson]
+  type action = Node.Trusted_validators_membership_change.action;
+
```
EduardoRFS, line 88: This can be removed
```suggestion
```

bin/files.rei
```diff
@@ -29,3 +30,14 @@ module State_bin: {
   let read: (~file: string) => Lwt.t(Protocol.t);
   let write: (Protocol.t, ~file: string) => Lwt.t(unit);
 };
+
+module Trusted_validators_membership_change: {
+  [@deriving yojson]
+  type action = Node.Trusted_validators_membership_change.action;
+
```
EduardoRFS, line 37:
```suggestion
```

node/trusted_validators_membership_change.re
```diff
@@ -0,0 +1,34 @@
...
+  | Add
+  | Remove;
+
+[@deriving yojson]
+type t = {
+  action,
+  address: Address.t,
+};
+
+let compare = (x, y) =>
```
EduardoRFS, line 15: Use `ppx_deriving.ord`

helpers/dune
```diff
@@ -1,5 +1,5 @@
 (library
  (name helpers)
- (libraries lwt uri zarith hex digestif)
+ (libraries lwt lwt.unix uri zarith hex digestif)
```
EduardoRFS, line 3: This change is not needed

EduardoRFS (changes requested):
The design seems to be mostly okay

helpers/dune
callistonianembrace: Indeed. Left over from the previous change that moved the file io to helpers.

bin/deku_node.re
callistonianembrace: Addressing in the incoming commit.

node/trusted_validators_membership_change.re
callistonianembrace: Addressing in the incoming commit

EduardoRFS (approved):
konata from lucky star saying good job

PR #185, opened 2021-09-07
Endpoint: To add/remove trusted_validators

callistonianembrace:
**Depends on**
#184 

**Problem**
Missing endpoint to add/remove trusted_validators from node's local state

**Solution**
This PR adds an http endpoint so that a CLI can written to help a node operation accept an off-chain verified `Add_validator` operation.


node/flows.re
```diff
@@ -411,3 +411,16 @@ let request_withdraw_proof = (state, ~hash) =>
   };
 let request_ticket_balance = (state, ~ticket, ~address) =>
   state.Node.protocol.ledger |> Ledger.balance(address, ticket);
+
+let trusted_validators_membership = (state, update_state, request) => {
```
EduardoRFS, line 415: This operation needs to be signed by the validator key, otherwise anyone could register a key as trusted validator

bin/http_server.re
```diff
@@ -161,6 +161,20 @@ let handle_receive_operation_gossip =
       Ok();
     },
   );
+
+let handle_trusted_validators_membership =
+  handle_request(
+    (module Trusted_validators_membership),
```
EduardoRFS, line 167: I'm not sure if this name makes sense

EduardoRFS (changes requested):

bin/http_server.re
callistonianembrace: Happy to use a different one if you have a suggestion. It seems fine to me.

node/flows.re
```diff
@@ -411,3 +411,16 @@ let request_withdraw_proof = (state, ~hash) =>
...
+  open Networking.Trusted_validators_membership;
+  let {address, action} = request;
+  let new_validators =
+    switch (action) {
+    | Add =>
+      Validators.(add({address: address}, state.Node.trusted_validators))
+    | Remove =>
+      Validators.(remove({address: address}, state.Node.trusted_validators))
+    };
+  ignore @@ update_state({...state, trusted_validators: new_validators});
```
EduardoRFS, line 425:
```suggestion
  let (_: State.t) = update_state({...state, trusted_validators: new_validators});
  ()
```

Big problem with ignore + partial application

EduardoRFS (changes requested):

anchpop (approved):
Whoops, meant to leave [this comment](https://github.com/marigold-dev/sidechain/pull/187#pullrequestreview-749433578) here.

bin/http_server.re
callistonianembrace: Trusted_validators_membership -> rusted_validators_membership_change

node/flows.re
callistonianembrace: Addressed in the incoming commit

callistonianembrace:
Adding persistence felt like jumping a few hoops. Added a separate commit which can be squashed later.

node/networking.re
```diff
@@ -184,6 +185,17 @@ module Ticket_balance = {
   let path = "/ticket-balance";
 };
 
+module Trusted_validators_membership_change = {
+  [@deriving yojson]
+  type request = {
+    signature: Signature.t,
+    payload: Trusted_validators_membership_change.t,
```
EduardoRFS, line 192: This is not okay, this type is internal to the node, networking should not rely on it as changing the internal type will break the networking layer.

I understand this is not the case for some endpoints already, but let's try to not make it worse.

My suggestion is similar to the following

```reason
type action = | Add | Remove
type request = {
  signature: Signature.t,
  action: action,
  address: Address.t
}
```

node/flows.re
```diff
@@ -412,3 +414,42 @@ let request_withdraw_proof = (state, ~hash) =>
...
+            {action, address},
+            state.Node.trusted_validator_membership_change,
+          )
+        | Remove =>
+          TVM_Set.remove(
+            {address, action},
+            state.Node.trusted_validator_membership_change,
+          )
+        };
+      let _: State.t =
```
EduardoRFS, line 446: nice

node/flows.re
```diff
@@ -412,3 +414,42 @@ let request_withdraw_proof = (state, ~hash) =>
...
+        update_state({
+          ...state,
+          trusted_validator_membership_change: new_validators,
+        });
+      persist(~file, new_validators |> TVM_Set.elements) |> Lwt_main.run;
```
EduardoRFS, line 451: Never uses Lwt_main.run outside of the toplevel. Also here it will crash the server.

EduardoRFS (changes requested):

bin/deku_node.re
```diff
@@ -9,6 +9,8 @@ open Opium;
 open Helpers;
 open Protocol;
 open Node;
+module TVM = Trusted_validators_membership_change;
+module TVM_Set = TVM.Set;
```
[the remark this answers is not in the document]
callistonianembrace: The `open Networking` causes namespace clashes.

node/networking.re
callistonianembrace: Addressed in the next commit

node/flows.re
callistonianembrace: Addressed in the next commit. It was no longer necessary to have Lwt_main in this PR. That being said, there are multiple `Lwt_main.run`s in `tezos_interop.re`

callistonianembrace:
@EduardoRFS Addressed everything except the TVM abbreviation.

bin/deku_node.re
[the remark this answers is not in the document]
callistonianembrace: Got rid of the toplevel open

node/flows.re
```diff
@@ -14,6 +14,10 @@ type ignore = [
   | `Not_current_block_producer
   | `Pending_blocks
 ];
+type endpoint_errors = [
+  | `Failed_to_create_payload_blake2b
+  | `Failed_to_verify_payload
+];
```
EduardoRFS, line 20: this type is not being used

node/flows.re
```diff
@@ -412,3 +414,51 @@ let request_withdraw_proof = (state, ~hash) =>
   };
 let request_ticket_balance = (state, ~ticket, ~address) =>
   state.Node.protocol.ledger |> Ledger.balance(address, ticket);
+
+let trusted_validators_membership =
+    (
+      ~file,
+      ~persist:
+         (list(Trusted_validators_membership_change.t), ~file: string) =>
+         Lwt.t(unit),
```
EduardoRFS, line 423:
```suggestion
      ~persist,
```

EduardoRFS (changes requested):

node/flows.re
EduardoRFS: Yes, but those on `tezos_interop` are on the toplevel. They're not called during the execution of the server only during bootstrapping.

node/flows.re
```diff
@@ -412,3 +414,51 @@ let request_withdraw_proof = (state, ~hash) =>
...
+      state,
+      update_state,
+      request,
+    ) => {
+  open Networking.Trusted_validators_membership_change;
+  let {signature, payload: {address, action} as payload_} = request;
+  switch (
+    payload_ |> payload_to_yojson |> Yojson.Safe.to_string |> BLAKE2B.of_string
```
EduardoRFS, line 431: `BLAKE2B.of_string` doesn't do what you want, it's not hashing it's just casting it. Probably you want to do `BLAKE2B.hash`.

node/flows.re
```diff
@@ -412,3 +414,51 @@ let request_withdraw_proof = (state, ~hash) =>
...
+      let new_validators =
+        switch (action) {
+        | Add =>
+          Trusted_validators_membership_change.Set.add(
+            {action: Add, address},
+            state.Node.trusted_validator_membership_change,
+          )
+        | Remove =>
+          Trusted_validators_membership_change.Set.remove(
+            {address, action: Remove},
```
EduardoRFS, line 444:
```suggestion
            {action: Remove, address},
```

node/flows.re
```diff
@@ -412,3 +414,51 @@ let request_withdraw_proof = (state, ~hash) =>
...
+        update_state({
+          ...state,
+          trusted_validator_membership_change: new_validators,
+        });
+      let.await () =
+        persist(
+          ~file,
+          new_validators |> Trusted_validators_membership_change.Set.elements,
+        );
+      Result.ok() |> Lwt.return;
```
EduardoRFS, line 458:
```suggestion
      await(Ok());
```

node/flows.re
```diff
@@ -412,3 +414,51 @@ let request_withdraw_proof = (state, ~hash) =>
...
+    } else {
+      Result.error(`Failed_to_verify_payload) |> Lwt.return;
```
EduardoRFS, line 460:
```suggestion
      await(Error(`Failed_to_verify_payload));
```

EduardoRFS (changes requested):

node/flows.re
callistonianembrace: They're used here: https://github.com/marigold-dev/sidechain/blob/callistonianembrace%2Fendpoint-add-offchain-trusted-validators/node/flows.re#L460-L462

node/flows.re
callistonianembrace: I'm committing this. But, `await` seems like a very bad name creating Lwt monads.

node/flows.re
callistonianembrace: The typechecker has difficult infering types here

node/flows.re
callistonianembrace: You're right. Fixing this in the next commit.

node/flows.re
EduardoRFS: that's not the used type, that's a polyvar you're not using the type

node/flows.re
EduardoRFS: Oh that's a neat case, possible improvements, you can reduce it to `~persist: (_, ~file: _) => _` to indicate the source of the problem or just revert the call site below from `persist(~file, param)` to `persist(param, ~file)`.

node/flows.re
callistonianembrace: Addressed this by using the type in the annotation:     `Lwt.t(result(unit, endpoint_errors('a))) =>  ...`

See https://github.com/marigold-dev/sidechain/pull/185/files#diff-43dae84d72a098cf13a754e41ae4fc754d43ee859e8be911f36ae612b542aaa5R425

node/flows.re
callistonianembrace: Addressed this by changing the order of params in the call site
This was interesting indeed.

node/flows.re
```diff
@@ -14,6 +14,7 @@ type ignore = [
   | `Not_current_block_producer
   | `Pending_blocks
 ];
+type endpoint_errors('a) = [> | `Failed_to_verify_payload] as 'a;
```
EduardoRFS, line 17: This type is not needed or being used on something meaningful, please remove it

node/flows.re
```diff
@@ -412,3 +411,40 @@ let request_withdraw_proof = (state, ~hash) =>
...
+      update_state({
+        ...state,
+        trusted_validator_membership_change: new_validators,
+      });
+    let.await () =
+      persist(
+        new_validators |> Trusted_validators_membership_change.Set.elements,
+        ~file,
+      );
+    Result.ok() |> Lwt.return;
```
EduardoRFS, line 446:
```suggestion
    await(Ok());
```

node/flows.re
```diff
@@ -412,3 +411,40 @@ let request_withdraw_proof = (state, ~hash) =>
...
+  } else {
+    Result.error(`Failed_to_verify_payload) |> Lwt.return;
```
EduardoRFS, line 448:
```suggestion
    await(Error(`Failed_to_verify_payload));
```

EduardoRFS (changes requested):

node/flows.re
```diff
@@ -412,3 +411,40 @@ let request_withdraw_proof = (state, ~hash) =>
...
+    let _: State.t =
+      update_state({
+        ...state,
+        trusted_validator_membership_change: new_validators,
+      });
+    let.await () =
+      persist(
+        new_validators |> Trusted_validators_membership_change.Set.elements,
+        ~file,
+      );
```
EduardoRFS, line 445: this persist can happen fully async as it is a function that never fails, use `Lwt.async`, this ensures that the entire flow doesn't use lwt which removes the need for the `handle_request_lwt`.

node/flows.re
```diff
@@ -412,3 +411,40 @@ let request_withdraw_proof = (state, ~hash) =>
...
 let request_ticket_balance = (state, ~ticket, ~address) =>
   state.Node.protocol.ledger |> Ledger.balance(address, ticket);
+
+let trusted_validators_membership =
+    (~file, ~persist, state, update_state, request)
+    : Lwt.t(result(unit, endpoint_errors('a))) => {
+  open Networking.Trusted_validators_membership_change;
+  let {signature, payload: {address, action} as payload_} = request;
+  let payload_hash =
+    payload_ |> payload_to_yojson |> Yojson.Safe.to_string |> BLAKE2B.hash;
```
EduardoRFS, line 421:
```suggestion
  let {signature, payload: {address, action} as payload} = request;
  let payload_hash =
    payload |> payload_to_yojson |> Yojson.Safe.to_string |> BLAKE2B.hash;
```

EduardoRFS (changes requested):

node/flows.re
callistonianembrace: Okay. Removing it.

node/flows.re
callistonianembrace: True. Resolving this in the next commit

node/flows.re
```diff
@@ -412,3 +410,40 @@ let request_withdraw_proof = (state, ~hash) =>
...
 let request_ticket_balance = (state, ~ticket, ~address) =>
   state.Node.protocol.ledger |> Ledger.balance(address, ticket);
+
+let trusted_validators_membership =
+    (~file, ~persist, state, update_state, request) => {
+  open Networking.Trusted_validators_membership_change;
+  let {signature, payload: {address, action} as payload} = request;
+  let payload_hash =
+    payload |> payload_to_yojson |> Yojson.Safe.to_string |> BLAKE2B.hash;
+  if (payload_hash |> Signature.verify(~signature)) {
```
EduardoRFS, line 420: This if can be replaced by something similar to the one below, the main goal is to keep the guards and the flow well defined

```suggestion
  let.assert () = (
    `Failed_to_verify_payload,
    payload_hash |> Signature.verify(~signature)
  )
```

EduardoRFS (approved):
A small suggestion, I may fix it tomorrow before merging it.

callistonianembrace:
Incorporated the suggestion!

EduardoRFS (approved):
anime girl happy at clean PR

PR #187, opened 2021-09-07
CLI: Add/Remove trusted validators from a node's local state

callistonianembrace:
## Depends

- [x] #184 
- [x] #185 

## Problem

Missing CLI for a node operator to add or remove trusted validators from node's state

## Solution

This PR uses endpoint from #185 to provide a convenient CLI

### DIFF
--- a/bin/sidecli.re
+++ b/bin/sidecli.re
@@ -87,6 +87,7 @@ let address = {
     );
   Arg.(conv((parser, printer)));
 };
+
 let address_tezos_interop = {
   let parser = string =>
     string
@@ -201,13 +202,13 @@ let create_transaction =
   Lwt.return(`Ok());
 };
 
-let create_transaction = {
-  let folder_node = {
-    let docv = "folder_node";
-    let doc = "The folder where the node lives.";
-    Arg.(required & pos(0, some(string), None) & info([], ~doc, ~docv));
-  };
+let folder_node = {
+  let docv = "folder_node";
+  let doc = "The folder where the node lives.";
+  Arg.(required & pos(0, some(string), None) & info([], ~doc, ~docv));
+};
 
+let create_transaction = {
   let address_from = {
     let doc = "The sending address, or a path to a wallet. If a bare sending address is provided, the corresponding wallet is assumed to be in the working directory.";
     let env = Arg.env_var("SENDER", ~doc);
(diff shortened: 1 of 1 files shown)


anchpop (approved):
Does this change make it so anyone who can connect to the node's endpoint can add/remove trusted validators? If so, is that intentional?

Also, I thought validators were going to be fetched from the consensus contract, so why is this necessary?

callistonianembrace:
> Also, I thought validators were going to be fetched from the consensus contract, so why is this necessary?
Adding/removing validators can be governed on the sidechain itself, and @EduardoRFS prefers to do them in sidechain whenever possible. Sidechain first, mainchain only if absolutely necessary

> Does this change make it so anyone who can connect to the node's endpoint can add/remove trusted validators? If so, is that intentional?

Please see https://github.com/marigold-dev/sidechain/pull/184#discussion_r704672814 Happy to clarify.

bin/sidecli.re
```diff
@@ -806,6 +807,104 @@ let self = {
...
+  let file = folder ++ "/identity.json";
+  let.await identity = Files.Identity.read(~file);
+  let payload = Trusted_validators_membership_change.{address, action: Add};
+  let payload_json_str =
+    payload
+    |> Trusted_validators_membership_change.to_yojson
+    |> Yojson.Safe.to_string;
+  switch (payload_json_str |> BLAKE2B.of_string) {
+  | None =>
+    await(`Error((false, "Internal Error: Failed to compute BLAKE2B")))
```
EduardoRFS, line 821: This doesn't make sense you're not hashing it.

bin/sidecli.re
```diff
@@ -806,6 +807,104 @@ let self = {
...
+  let docv = "validator_address";
+  let doc = "The validator address to be added/removed as trusted";
+  Arg.(required & pos(2, some(address_t), None) & info([], ~docv, ~doc));
+};
+
+let add_trusted_validator = {
+  Term.(
+    lwt_ret(const(add_trusted_validator) $ folder_node $ validator_address)
+  );
+};
```
EduardoRFS, line 898: Put them alongside the info declaration and also the code declaration, commands in this file have it's parts clustered.

EduardoRFS (changes requested):

bin/sidecli.re
callistonianembrace: Making it "Failed to obtain Blake2b from string". Feel free to offer alternatives otherwise.

bin/sidecli.re
callistonianembrace: Addressed

bin/sidecli.re
EduardoRFS: You're not computing `BLAKE2B`, I believe you're looking for the `BLAKE2B.hash` function.

bin/sidecli.re
```diff
@@ -806,6 +807,107 @@ let self = {
...
+let add_trusted_validator = (folder, address) => {
+  open Networking;
+  let file = folder ++ "/identity.json";
+  let.await identity = Files.Identity.read(~file);
+  let payload = Trusted_validators_membership_change.{address, action: Add};
+  let payload_json_str =
+    payload
+    |> Trusted_validators_membership_change.payload_to_yojson
+    |> Yojson.Safe.to_string;
+  switch (payload_json_str |> BLAKE2B.of_string) {
```
EduardoRFS, line 830: Look at `BLAKE2B.hash`

bin/sidecli.re
```diff
@@ -806,6 +807,107 @@ let self = {
...
+  open Networking;
+  let file = folder ++ "/identity.json";
+  let.await identity = Files.Identity.read(~file);
+  let payload =
+    Trusted_validators_membership_change.{address, action: Remove};
+  let payload_json_str =
+    payload
+    |> Trusted_validators_membership_change.payload_to_yojson
+    |> Yojson.Safe.to_string;
+  switch (payload_json_str |> BLAKE2B.of_string) {
```
EduardoRFS, line 889: Look at `BLAKE2B.hash`

EduardoRFS (changes requested):

bin/sidecli.re
callistonianembrace: Indeed. Fixed in the next commit

bin/sidecli.re
callistonianembrace: Indeed. Fixed in the next commit

bin/sidecli.re
callistonianembrace: Fixed in the next commit

EduardoRFS (approved):
Also looks good to me, just needs to be rebased, will do that tomorrow.

callistonianembrace:
Rebased over PR 185 to make any future rebasing easier :+1:

PR #192, opened 2021-09-09
Rename http_server to deku-node

callistonianembrace:
## Problem

The node binary is named http_server which isn't descriptive enough, arguably even wrong.

## Solution

This PR renames the main validator node to `deku-node`, after the proposed name Deku for the sidechain.


anchpop:
> The node binary is named http_server which isn't descriptive enough, arguably even wrong.

so true, thank you for fixing this

anchpop (approved):

EduardoRFS (approved):
deku vibing hard on the PR being merged

PR #194, opened 2021-09-22
Rename Conflicting State Modules

d4hines:
## Problem
We currently have two modules called `State` - one for the protocol and one for the node. Tshis has two disadvantages:
- It's confusing which one you're referring to in the code unless you rename the modules manually (especially when opening both modules, which we do)
- It's slightly harder to navigate to the right file - unique names make it easier in the IDE.

## Solution
This PR renames the modules and adds a guideline not to have conflicting names in HACKING.md


anchpop (approved):
haven't tested it locally, but it looks good to me and is a very good change, thank you!

node/building_blocks.re
```diff
@@ -1,9 +1,7 @@
 open Helpers;
 open Protocol;
 
-module Node = State;
-
-let is_valid_block = (state, block) => {
+let is_valid_block = (state: Node_state.t, block) => {
```
EduardoRFS, line 4: Never add unneeded type annotation

node/building_blocks.re
```diff
@@ -26,41 +24,40 @@ let is_valid_block = (state, block) => {
   Ok();
 };
 
-let is_next = (state, block) => Protocol.is_next(state.Node.protocol, block);
+let is_next = (state: Node_state.t, block) =>
```
EduardoRFS, line 27: Don't add the type annotation

node/building_blocks.re
```diff
@@ -26,41 +24,40 @@ let is_valid_block = (state, block) => {
...
+  Protocol.is_next(state.protocol, block);
 
-let has_next_block_to_apply = (state, ~hash) =>
-  Block_pool.find_next_block_to_apply(~hash, state.Node.block_pool)
+let has_next_block_to_apply = (state: Node_state.t, ~hash) =>
```
EduardoRFS, line 30: Don't add the type annotation

node/building_blocks.re
```diff
@@ -26,41 +24,40 @@ let is_valid_block = (state, block) => {
...
+  Block_pool.find_next_block_to_apply(~hash, state.block_pool)
   |> Option.is_some;
 
-let is_known_block = (state, ~hash) =>
-  Option.is_some(Block_pool.find_block(state.Node.block_pool, ~hash));
-let is_known_signature = (state, ~hash, ~signature) => {
+let is_known_block = (state: Node_state.t, ~hash) =>
```
EduardoRFS, line 34: Don't add the type annotation

node/building_blocks.re
```diff
@@ -26,41 +24,40 @@ let is_valid_block = (state, block) => {
...
+  Option.is_some(Block_pool.find_block(state.block_pool, ~hash));
+let is_known_signature = (state: Node_state.t, ~hash, ~signature) => {
```
EduardoRFS, line 36: Don't add the type annotation

node/building_blocks.re
```diff
@@ -26,41 +24,40 @@ let is_valid_block = (state, block) => {
...
   let.default () = false;
-  let.some signatures =
-    Block_pool.find_signatures(~hash, state.Node.block_pool);
+  let.some signatures = Block_pool.find_signatures(~hash, state.block_pool);
   Some(Signatures.mem(signature, signatures));
 };
 
-let is_signed_by_self = (state, ~hash) => {
+let is_signed_by_self = (state: Node_state.t, ~hash) => {
```
EduardoRFS, line 42: Don't add the type annotation

node/building_blocks.re
```diff
@@ -26,41 +24,40 @@ let is_valid_block = (state, block) => {
...
   // TODO: for the name of this function being correct we shuold check it recursively
   let.default () = false;
-  let.some signatures =
-    Block_pool.find_signatures(~hash, state.Node.block_pool);
+  let.some signatures = Block_pool.find_signatures(~hash, state.block_pool);
   Some(Signatures.is_self_signed(signatures));
 };
 
-let is_current_producer = (state, ~key) => {
+let is_current_producer = (state: Node_state.t, ~key) => {
```
EduardoRFS, line 49: Don't add the type annotation

node/building_blocks.re
```diff
@@ -26,41 +24,40 @@ let is_valid_block = (state, block) => {
...
   let.default () = false;
-  let.some current_producer = get_current_block_producer(state.Node.protocol);
+  let.some current_producer = get_current_block_producer(state.protocol);
   Some(current_producer.address == key);
 };
 
 // TODO: bad naming
 // TODO: check if block must have published a new snapshot
-let is_signable = (state, block) => {
+let is_signable = (state: Node_state.t, block) => {
```
EduardoRFS, line 57: Don't add the type annotation

node/building_blocks.re
```diff
@@ -72,55 +69,54 @@ let is_signable = (state, block) => {
 
 let sign = (~key, block) => Block.sign(~key, block);
 
-let produce_block = state =>
+let produce_block = (state: Node_state.t) =>
```
EduardoRFS, line 72: Don't add the type annotation

node/building_blocks.re
```diff
@@ -72,55 +69,54 @@ let is_signable = (state, block) => {
...
-    ~state=state.Node.protocol,
+    ~state=state.protocol,
     ~author=state.identity.t,
     ~main_chain_ops=state.pending_main_ops,
     ~side_chain_ops=state.pending_side_ops,
   );
 
-let is_valid_block_height = (state, block_height) =>
-  block_height >= 1L && block_height <= state.Node.protocol.block_height;
+let is_valid_block_height = (state: Node_state.t, block_height) =>
```
EduardoRFS, line 80: Don't add the type annotation

node/building_blocks.re
```diff
@@ -72,55 +69,54 @@ let is_signable = (state, block) => {
...
-    Validators.length(state.Node.protocol.validators);
+let signatures_required = (state: Node_state.t) => {
+  let number_of_validators = Validators.length(state.protocol.validators);
   // TODO: properly filter and check signatures
   Float.(to_int(ceil(of_int(number_of_validators) *. (2.0 /. 3.0))));
 };
 
 // mutations
-let append_signature = (state, update_state, ~hash, ~signature) => {
+let append_signature = (state: Node_state.t, update_state, ~hash, ~signature) => {
```
EduardoRFS, line 90: Don't add the type annotation

node/flows.re
```diff
@@ -2,8 +2,6 @@ open Helpers;
 open Protocol;
 open Building_blocks;
 
-module Node = State;
```
EduardoRFS, line 5: This is not an improvement, when we do `Node.something` I don't mean `the node state do something`, the way to read it is completely different, that's why there is this module alias here.

The only reason why the file `state.re` is not called `node.re` is because of how dune libraries naming works.

EduardoRFS (changes requested):
First, follow the guideline for a PR description, look at #182, second the changes degrade considerably the readability of the code, removing the ambiguity is not worth if you need to duplicate everything.

This PR is well defined, it talks about the problem of having two module files with the same name, which is problematic for navigation, and I agree with this, but the PR is also adding a bunch of unneeded type annotations which are completely unrelated to the topic. And it also expand the code in a lot of place and remove some `open` which is again completely unrelated to the topic.

## Action points

- Define the scope of the PR
- Write a proper PR description
- Define what is the proposed change
- Make a PR doing strictly the proposed change

node/building_blocks.re
d4hines: I did not intend to add unneeded type annotations, only those that were needed. I could easily have made a mistake in this respect. 

In general, do you prefer: `x.ModuleA.propA` over `let x : ModuleA = ... in x.propA`?

node/flows.re
d4hines: In Tezos there are lots of module aliasing, and I find it obscures the code considerably (other Tezos developers have said this too). Therefore I believe there should not be any module aliases that are not necessary. The rule should be:
- Try to avoid conflicting names
- Only if you can't should you do a module alias.

Issue #147, opened 2021-08-12
parallel state hashing


EduardoRFS:

d4hines:
See design in https://github.com/marigold-dev/deku/pull/218/files

PR #195, opened 2021-09-22
Proto: Add Async State Hashing

d4hines:
## Problem

This PR addresses async state hashing initially described in #147 and more fully specified and designed in #218.

To reduce the scope, this PR adds async hashing to the block producer only.

## Solution

See block producer section of #218 for details on the whole design. We'll address a truncated version of it here.

Currently the node stores its state, which includes the current state root hash and a timestamp of the last update to that hash. Additionally, nodes assert that a bloc's state root hash is valid before applying it, which means either:
  - the block's state root hash is equal to the current state root hash
  - or else it is equal to the next state root hash (generated synchronously on the spot). 

We'll extend this state with a field `next_state_root_hash : BLAKE2B.t`. This will store the result of asynchronously hashing a state.

The design is as follows:
1. After a node applies a block, it checks to see if the block's state root hash is equal to the current state root hash. If it is, it begins hashing it's new state.
2. When the hashing job is done, `next_state_root_hash` is set to the result.
3. When producing a block, use the `last_state_root_hash_update` field to determine if sufficient time has passed to include a new state root hash. If it has, use `next_state_root_hash` as the state root of the block; otherwise use the `current_state_root_hash`.
4. When applying a block, we change the meaning of valid to mean one of the following:
  - 1. The block's state root hash is equal to the current state root hash (unchanged)
  - 2. or else it is equal to the next state root hash (which was generated previously). Additionally,
    synchronously generate the next state root hash and update `next_state_root_hash`.
    
 You can see that the last case is defined inductively. We therefore need to define a base case: initially, the `current_state_root_hash` and `next_state_root_hash` are set to be pre-determined (distinct) hashes. 
 
 So when the genesis block is applied, we enter rule 4.2, setting the `current_state_root_hash` to the value of `next_state_root_hash`, and generating a new `next_state_root_hash`.

### DIFF
--- a/node/flows.re
+++ b/node/flows.re
@@ -36,6 +36,8 @@ let received_block':
         | `Invalid_state_root_hash
         | `Not_current_block_producer
         | `Pending_blocks
+        | `State_root_hash_update_is_early
+        | `State_root_hash_update_is_late
       ],
     ),
   ) =
@@ -51,6 +53,8 @@ let block_added_to_the_pool':
         | `Invalid_block_when_applying
         | `Invalid_state_root_hash
         | `Not_current_block_producer
+        | `State_root_hash_update_is_early
+        | `State_root_hash_update_is_late
       ],
     ),
   ) =
@@ -183,6 +187,12 @@ let rec try_to_apply_block = (state, update_state, block) => {
     `Block_not_signed_enough_to_apply,
     Block_pool.is_signed(~hash=block.Block.hash, state.Node.block_pool),
   );
+  let.assert () = (
+    `Invalid_state_root_hash,
+    state.protocol.state_root_hash == block.state_root_hash
+    || state.next_state_root_hash == block.state_root_hash,
+  );
+
   let.ok state = apply_block(state, update_state, block);
   reset_timeout^();
   let state = clean(state, update_state, block);
--- a/node/flows.re
+++ b/node/flows.re
@@ -158,6 +158,52 @@ let request_previous_blocks = (state, block) =>
     request_protocol_snapshot();
   };
 
+/** Starts asynchronously hashing a new state.
+    FIXME: fix this comment
+
+    Side-effectfully updates the current [next_state_root_hash] and
+    [last_state_root_update] when done.
+    TODO: Note that it is possible for multliple executions of this function
+    to be running concurrently, resulting in nondeterministic update
+    of [next_state_root_hash]. This needs to replaced by more sophisticated
+    mechanism when we implement async state hashing for validators per #147.
+
+    Why this exists:
+    Each block is associated with a particular state root hash.
+    The state root starts with genesis, and then updates periodically.
+    This means you don't have to download the entire history of the chain
+    to derive the state corresponding to a given block - only the state
+    corresponding to its state root hash is required.
+
+    Because hashing a state is an expensive operation, we do it in parallel
+    on a separate thread. This allows the block producer to continue
+    producing blocks even while the state root is being updated.
+    We spawn these jobs on an interval determined by
+    [should_start_hashing_new_state]).
+ */
+let hash_new_state_root = (state, update_state) => {
+  let current_epoch = state.Node.current_epoch;
+  Lwt.async(() => {
+    // Lwt_domain.detach causes the function to run in a separate thread.
+    // When the promise resolves, the rest of the function is run in main thread.
+    let.await (hash, _) =
+      Lwt_domain.detach((s: Node.t) => Protocol.hash(s.protocol), state);
+    // The state passed to this function may be stale so we get the current state.
+    let current_state = get_state^();
+    let _ =
+      update_state({
+        ...current_state,
+        finished_state_root_hashes:
+          State.Int_map.add(
+            current_epoch,
+            hash,
+            current_state.finished_state_root_hashes,
+          ),
+      });
+    Lwt.return();
+  });
+};
+
 let try_to_produce_block = (state, update_state) => {
   let.assert () = (
     `Not_current_block_producer,
@@ -193,7 +239,38 @@ let rec try_to_apply_block = (state, update_state, block) => {
     || state.next_state_root_hash == block.state_root_hash,
   );
 
+  let current_state_root_hash = state.Node.protocol.state_root_hash;
   let.ok state = apply_block(state, update_state, block);
+
+ // If the state root hash has changed by application of the block,
+ // then, by definition, the state root epoch has changed. If
+ // that's the case we increment the current_epoch and start
+ // hashing a new state
+  let new_epoch =
+    state.Node.protocol.state_root_hash != current_state_root_hash;
+
+  let state =
+    new_epoch
+      ? {
+        ...state,
+        current_epoch: state.current_epoch + 1,
+        finished_state_root_hashes:
+          // We remove the hash of the previous epoch to prevent
+          // the map from growing indefinitely. Additionally, this
+          // guarantees the state root epoch can only move forward.
+          // FIXME:
+          // State.Int_map.remove(
+          //   state.current_epoch,
+          //   state.finished_state_root_hashes,
+          // ),
+          state.finished_state_root_hashes
+      }
+      : state;
+
+  if (new_epoch) {
+    hash_new_state_root(state, update_state);
+  };
+
   reset_timeout^();
   let state = clean(state, update_state, block);
   switch (
--- a/protocol/block.rei
+++ b/protocol/block.rei
@@ -22,6 +22,7 @@ let genesis: t;
 let produce:
   (
     ~state: State.t,
+    ~next_state_root_hash: option(BLAKE2B.t),
     ~author: Address.t,
     ~main_chain_ops: list(Main_chain.t),
     ~side_chain_ops: list(Side_chain.t)
--- a/protocol/block.rei
+++ b/protocol/block.rei
@@ -19,10 +19,16 @@ type t =
 let sign: (~key: Address.key, t) => Signature.t;
 let verify: (~signature: Signature.t, t) => bool;
 let genesis: t;
+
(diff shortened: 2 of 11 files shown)


EduardoRFS:
As discussed in private, this PR should be separated from #194 and as mentioned here on the [guidelines](https://github.com/marigold-dev/deku#guidelines)

> Refer merged PRs to get an idea of what is considered a good PR description.

My tip, look on #182

protocol/block.rei
```diff
@@ -21,6 +21,7 @@ let verify: (~signature: Signature.t, t) => bool;
 let genesis: t;
 let produce:
   (
+    ~state_root_hash: BLAKE2B.t,
```
EduardoRFS, line 24: This signature is not right, the function already receives the state, it should be able access the state root hash available.

EduardoRFS: And especially this allows to produce a block with the wrong state_root_hash

node/node_state.rei
```diff
@@ -25,6 +25,31 @@ type t = {
...
+
+    At the start of each new finality period, we start hashing the current
+    state, requiring every validator to finish hashing that same state before
+    the start of the next finality period. At the end of the finality
+    period, the state hash finished in that period becomes the hash used
+    in the next finality period. Because a block always points to the state
+    hash of the previous period, nodes does not need to wait for the expensive
+    operation of hashing to complete before they can apply a block, and can thus
+    hash the state asynchronously. */
+  state_root_hash_list: list((BLAKE2B.t, BLAKE2B.t, float)),
```
EduardoRFS, line 52: Make this a record, the comment is a nice to have, but it would be a lot more clear if it was on the type definition of the what a `state_root_hash_list` means here.

Also I believe this can be it's own module as it models the domain of the system.

EduardoRFS: Also I don't understand why this contains the list of validators hash, what does it achieve?

node/flows.re
```diff
@@ -152,14 +152,94 @@ let request_previous_blocks = (state: Node_state.t, block) =>
...
+  // emits blocks with the most recent state root hash he has. Ostensibly
+  // it should not take longer than state root epoch to hash a state root.
+  // If this hold, the finality period is equal to 1 state root epoch.
+  // If the block producer hashes slower than this rate, then the finality
+  // period increases. If other validators hash slower than the finality period
+  // and slower than the block producer, then they will begin to fall behind,
+  // eventually getting out of sync if they
+  // fall far enough behind.
+  // TODO: validate that this all true.
+  let (state_root_hash, _, _) = List.hd(state.state_root_hash_list);
```
EduardoRFS, line 241: Don't use `List.hd`, if you know that it is an invariant(which I don't think it's the case), move it to a module and encapsulate it on the type system.

node/flows.re
```diff
@@ -152,14 +152,94 @@ let request_previous_blocks = (state: Node_state.t, block) =>
...
+        }
+      );
+    ();
+  } else {
+    ();
+  };
+};
+
+let try_to_produce_block =
+    (state: Node_state.t, update_state: Node_state.t => Node_state.t) => {
```
EduardoRFS, line 221: Don't add unneeded types.

node/flows.re
```diff
@@ -152,14 +152,94 @@ let request_previous_blocks = (state: Node_state.t, block) =>
...
+            Validators.hash(s.protocol.validators),
+          ),
+        state,
+      )
+      // When the promise resolves, the following callback is run in main thread.
+      >|= (
+        (((state_root_hash, _), validator_hash)) => {
+          // The state passed to this function may be stale so we get the current state.
+          let current_state = get_state^();
+          ignore @@
```
EduardoRFS, line 203: Don't use ignore, use `let _: something`, also don't use `@@`

node/flows.re
```diff
@@ -152,14 +152,94 @@ let request_previous_blocks = (state: Node_state.t, block) =>
...
+      Lwt_domain.detach(
+        (s: Node_state.t) =>
+          (
+            Protocol.hash(s.protocol),
+            Validators.hash(s.protocol.validators),
+          ),
+        state,
+      )
+      // When the promise resolves, the following callback is run in main thread.
+      >|= (
```
EduardoRFS, line 199: Don't use infix operators, refer to existing code on how to do it.

node/flows.re
```diff
@@ -152,14 +152,94 @@ let request_previous_blocks = (state: Node_state.t, block) =>
...
+      ~update_state: Node_state.t => Node_state.t,
+    ) => {
+  let (_, _, last_state_root_update) = List.hd(state.state_root_hash_list);
+  // If it's a new epoch then start hashing the current state
+  if (is_new_state_root_epoch(last_state_root_update, current_time)) {
+    open Lwt.Infix;
+    // Infix operators used because `let.async` transforms whole function
+    // into Lwt monad. Instead, we use infix and ignore the result because
+    // we only care about the side effects, not the result.
+    let _ =
```
EduardoRFS, line 188: never do `let _` on a promise

node/flows.re
```diff
@@ -152,14 +152,94 @@ let request_previous_blocks = (state: Node_state.t, block) =>
...
+let try_hash_new_state =
+    (
+      ~state: Node_state.t,
+      ~current_time,
+      ~update_state: Node_state.t => Node_state.t,
+    ) => {
+  let (_, _, last_state_root_update) = List.hd(state.state_root_hash_list);
+  // If it's a new epoch then start hashing the current state
+  if (is_new_state_root_epoch(last_state_root_update, current_time)) {
+    open Lwt.Infix;
```
EduardoRFS, line 184: Never open `Module.Infix`

EduardoRFS (changes requested):

node/flows.re
d4hines: Addressed here: https://github.com/marigold-dev/deku/pull/195/commits/2ebfd66460933c061c84bdb66f6dce484547a1ff

node/flows.re
d4hines: Addressed here: https://github.com/marigold-dev/deku/pull/195/commits/2ebfd66460933c061c84bdb66f6dce484547a1ff

node/flows.re
d4hines: Addressed here: https://github.com/marigold-dev/deku/pull/195/commits/2ebfd66460933c061c84bdb66f6dce484547a1ff

node/flows.re
d4hines: Addressed here: https://github.com/marigold-dev/deku/pull/195/commits/2ebfd66460933c061c84bdb66f6dce484547a1ff

protocol/block.rei
d4hines: Fixed here: https://github.com/marigold-dev/deku/pull/195/commits/08de202b8d9b9ea950c8cd2a29ef3a108bcbc4e5

node/node_state.rei
d4hines: I didn't understand an offline comment you made about validator hashes. Fixed here: https://github.com/marigold-dev/deku/pull/195/commits/08de202b8d9b9ea950c8cd2a29ef3a108bcbc4e5

node/flows.re
d4hines: Fixed here: https://github.com/marigold-dev/deku/pull/195/commits/08de202b8d9b9ea950c8cd2a29ef3a108bcbc4e5

node/flows.re
d4hines: fixed here: https://github.com/marigold-dev/deku/pull/195/commits/0d29ac726aff6248b33c9b7440f58eb2e470367d

d4hines:
@EduardoRFS, I've updated the PR to just focus on block producers - much smaller scope. Will follow up with validators.

EduardoRFS:
### What if the chain disconnects?

There is an important discussion to have, should we accept a state hash which is not what we are expecting? Let's say we're expecting the state to be hashed at height 10, but it was hashed at height 12, that is definitely not signable, but is that okay to apply?

Ideally we wouldn't not even apply as that seems like malformed state to me. But then we need to store the previous state so that we can verify it if the entire chain get's down and we need to start it up again.

### Comments on your design

> done : Set.t(BLAKE2B.t)

I feel like there is ways that this set will contain hashes that will never leave here, let's say deku hashed block height 20 and then skipped to block 1000 because it was disconnected temporarily. Not big problem but it's worth to think.

> If yes, he can sign the block.

Not necessarily, an additional condition which is really important here, is that there is a time limit to update the state root hash, so at some point it stops being signable, otherwise the block producer may never update the state root hash undermining the security of the chain.

> If it is a new state root epoch (see current implementation for logic), start hashing the current state
(which implicitly adds the resultant hash to Hashing).

This is wrong, there is no way to ensure that validators are gonna agree here and it may lead to some validators thinking that the hashing epoch is different leading to a stale chain.

As discussed, a state hash epoch starts when the state root hash changes and as that is part of the block we know when it changed last time.

> In this scheme, we know the validator is in sync whenever pending is empty.

I fail to see what implies this, there is no guarantee that we've been following the right chain since the beginning, and being in sync for Deku means exactly that.

d4hines:
> ...But then we need to store the previous state so that we can verify it if the entire chain get's down and we need to start it up again.

I'm not following this point. Can you explain?

EduardoRFS:
What I meant is the scenario where the chain is stale, which state should be hashed?

> we need to ensure the chain doesn't stall irrecoverably if we get one block out of sync with state hashing

The chain doesn't stale just because your node got out of sync, the will be in an invalid state but the remaining of the chain should keep running.

> In any case, I think this could be addressed in a subsequent PR 

Nope, this is a fundamental problem, otherwise if we merge this PR as it is, a stale chain will never be able to start running again.

also the other points needs to be addressed, as they're even worse.

d4hines:
Closing in favor of #234

PR #196, opened 2021-09-22
Make some dev dependencies more specific

anchpop:
## Problem

Some of the dev dependencies weren't specific enough. I'd make a change to `package.json`, the dependency solve would change, the lockfiles would update, then the build would break. Then I'd revert the change to `package.json` but because the lockfiles had already updated and they specified versions of the dependencies that didn't work, the build would still be broken. 

## Solution

Make the problematic dependency constraints more specific so the lockfile stops updating to invalid states


callistonianembrace (approved):
This is an improvement, but lockfile generation in esy isn't incremental. So if you add a new dependency, the solution will be recalculated despite this. With this PR, the solution within the bounds, so its definitely an improvement. But it still wont guarantee successful builds because of transitive dependencies - we'd have to pin all of them.

PS: `@esy-ocam/rtop` is resolved to an actual github url. So the version is ignored. This is fine and can be revisited later.

Approving because of the improvement is reasonable.

EduardoRFS:
Also reasonable to me, for some reason utop is always decreasing so this leads to problems.

callistonianembrace:
This PR needs a rebase, @anchpop.

PR #197, opened 2021-09-22
Add ocamlformat-rpc as devDependency

d4hines:
I got a warning like the following in VS Code when opening the project:

> OCamlformat_rpc is missing, displayed types might not be properly formatted.

I don't know if this is bad, but @zett98 suggested I install `ocamlformat-rpc`. I did this and the error went away. 

If this is bad/unnecessary, feel free to close - I didn't research the error thoroughly.


anchpop (approved):
Never had this issue before but I don't see the harm

callistonianembrace (commented):
I haven't run into this issue myself, but, I see no reason to block this PR either. Any chance @d4hines , would you be able to explain why the VSCode issue disappears? Just to make sure co-relation isn't causation.

Zett98:
I think it's related to this: https://github.com/ocaml/ocaml-lsp/releases/tag/1.8.0

callistonianembrace:
I took a cursory look at https://github.com/ocaml/ocaml-lsp/pull/386. IIUC, `ocamlformat-rpc` seems optional and shouldn't be a cause of concern. It might cause degraded performance - having to fork/exec ocamlfmt everytime. 

If you're wondering why we need ocamlformat when our syntax is in Reason - ocamlformat also format other files like `dune` etc.

It's good to have this PR merged in, IMO.

EduardoRFS (approved):
anime girl giving a star to Daniel for this PR

PR #200, opened 2021-09-24
Add an OCaml Style Guide

d4hines:
## Problem

We currently don't have a place to aggregate style requirements. We need one so as to not slow down PR's with things that could be easily addressed ahead of time. 

## Solution

Adds a section to the HACKING.md document.

Note I don't know what the preference is for handling promises that invoked purely for side effects, so we need to address the `FIXME:` before merging.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -17,7 +17,7 @@ An identifier named t on a module will always have type t, so under `Validators.
 ## OCaml Style Guide
 
 ### OCaml Basics
-- Never use the `ignore` function. Instead prefer `let _ = ...`
+- Never use the `ignore` function. Instead prefer `let (_: typ) = ...`
 - Never use the `@@` operator.
 - Never add extraneous type annotations.
 - Never use infix operators to chain monads together. Instead use `let` syntax.
--- a/HACKING.md
+++ b/HACKING.md
@@ -20,7 +20,8 @@ An identifier named t on a module will always have type t, so under `Validators.
 - Never use the `ignore` function. Instead prefer `let _ = ...`
 - Never use the `@@` operator.
 - Never add extraneous type annotations.
+- Never use infix operators to chain monads together. Instead use `let` syntax.
 
 ### LWT
-- Never use `Lwt`'s infix operators. Instead use the `let` syntax.
-- Never ignore the result of an `Lwt` promise. Instead (FIXME: what to do instead?)
+- Never ignore the result of an `Lwt` promise. Instead, if a promise is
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -13,3 +13,14 @@ Sometimes you will see it being used even for single cases `v |> f`, this is mor
...
+
+## OCaml Style Guide
+
+### OCaml Basics
+- Never use the `ignore` function. Instead prefer `let _ = ...`
+- Never use the `@@` operator.
+- Never add extraneous type annotations.
+
+### LWT
+- Never use `Lwt`'s infix operators. Instead use the `let` syntax.
```
EduardoRFS, line 25: more general this is not only for Lwt, this is for any monadic infix operator

HACKING.md
```diff
@@ -13,3 +13,14 @@ Sometimes you will see it being used even for single cases `v |> f`, this is mor
...
+- Never ignore the result of an `Lwt` promise. Instead (FIXME: what to do instead?)
```
EduardoRFS, line 26: handle it properly, if it's a side operation it can be done through Lwt.async or just return an Lwt.t

EduardoRFS (changes requested):

HACKING.md
d4hines: Addressed in https://github.com/marigold-dev/deku/pull/200/commits/d0b514aa96a740d257d3ba58ecbd4f9bd2e4eaaa

HACKING.md
d4hines: Addressed in https://github.com/marigold-dev/deku/pull/200/commits/d0b514aa96a740d257d3ba58ecbd4f9bd2e4eaaa

HACKING.md
```diff
@@ -13,3 +13,15 @@ Sometimes you will see it being used even for single cases `v |> f`, this is mor
 A type named t means the main type of a module so, `Validators.t` can be read as `validators` and `Wallet.t` as `wallet`.
 
 An identifier named t on a module will always have type t, so under `Validators.re` `let t = empty` is the same as `let validators = empty`
+
+## OCaml Style Guide
+
+### OCaml Basics
+- Never use the `ignore` function. Instead prefer `let _ = ...`
```
EduardoRFS, line 20:
```suggestion
- Never use the `ignore` function. Instead prefer `let (_: typ) = ...`
```

EduardoRFS (approved):
anime girl cleaning the room

PR #205, opened 2021-09-30
proto: move amount and ticket to operation kind

callistonianembrace:
## Problem

Every operation is currently associated with an amount and a ticket. This is no longer true after #182 - some sidechain operations, `Add_validator`, `Remove_validator` etc don't need these properties.

## Solution

This PR move these fields to `Withdraw` and `Transaction` operations only where they are needed

### DIFF
--- a/protocol/operation.re
+++ b/protocol/operation.re
@@ -54,8 +54,16 @@ module Side_chain = {
   // TODO: I don't like this structure model
   [@deriving yojson]
   type kind =
-    | Transaction({destination: Wallet.t})
-    | Withdraw({owner: Tezos_interop.Address.t})
+    | Transaction({
+        destination: Wallet.t,
+        amount: Amount.t,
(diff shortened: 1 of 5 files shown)


anchpop (approved):

protocol/operation.re
```diff
@@ -66,21 +74,16 @@ module Side_chain = {
     nonce: int32,
     block_height: int64,
     source: Wallet.t,
```
EduardoRFS, line 76: Question, does source makes sense for add and remove validators? When will the block producer add an Add and Remove validator operation?

callistonianembrace: I think it's useful for bookkeeping. It's good to know who proposed that a certain validator be added or removed

EduardoRFS: Agreed

EduardoRFS (approved):
anime girl cooking

PR #206, opened 2021-09-30
cli: adds commands to propose adding/removing validators  

callistonianembrace:
## Depends

- [x] #205
- [x] #228 
- [x] #325 
- [ ] #333 (for network participation of the new validator)
- [x] #334  

## Problem

We don't have a way to issue protocol operations that add or remove validators from the network

## Solution

This PR adds two CLI commands to allow validators to propose adding and removing validators from the network

### DIFF
--- a/bin/sidecli.re
+++ b/bin/sidecli.re
@@ -732,7 +732,7 @@ let address_t = {
 let validator_address = {
   let docv = "validator_address";
   let doc = "The validator address to be added/removed as trusted";
-  Arg.(required & pos(2, some(address_t), None) & info([], ~docv, ~doc));
+  Arg.(required & pos(1, some(address_t), None) & info([], ~docv, ~doc));
 };
 
 let add_trusted_validator = {
--- a/bin/sidecli.re
+++ b/bin/sidecli.re
@@ -88,7 +88,7 @@ let uri = {
   Arg.(conv((parser, printer)));
 };
 
-let address = {
+let wallet_address = {
   let parser = string =>
     Key_hash.of_string(string)
     |> Option.map(Address.of_key_hash)
@@ -111,6 +111,17 @@ let address_tezos_interop = {
     Format.fprintf(fmt, "%s", Tezos.Address.to_string(address));
   Arg.(conv((parser, printer)));
 };
+
+let address = {
+  let parser = string =>
+    string
+    |> Address.of_string
+    |> Option.to_result(~none=`Msg("Expected a public_key"));
+  let printer = (fmt, address) =>
+    Format.fprintf(fmt, "%s", Address.to_string(address));
+  Arg.(conv((parser, printer)));
+};
+
 let amount = {
   let parser = string => {
     let.ok int =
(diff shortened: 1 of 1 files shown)


callistonianembrace:
There are some TODOs to be addressed. Assigning this back to me.

anchpop (approved):

EduardoRFS (changes requested):

bin/sidecli.re
```diff
@@ -894,6 +907,152 @@ let remove_trusted_validator = {
...
+      ~secret=wallet.priv_key,
+      ~nonce=0l,
+      ~block_height=block_level,
+      ~source=wallet.address,
+      ~kind=Add_validator({address: validator_address}),
+    );
+
+  let.await () =
+    Networking.request_operation_gossip(
+      Networking.Operation_gossip.{operation: membership_change},
```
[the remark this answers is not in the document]
callistonianembrace: Is there any http communication where the payload doesn't have to be signed by the owner of the node?

node/networking.re
```diff
@@ -151,7 +151,10 @@ module Register_uri = {
 
 module Operation_gossip = {
   [@deriving yojson]
-  type request = {operation: Operation.Side_chain.t};
+  type request = {
+    signature: option(Signature.t),
```
EduardoRFS, line 155: This is not needed, every side chain operation already carries a signature.

node/flows.re
```diff
@@ -294,6 +294,32 @@ let received_signature = (state, update_state, ~hash, ~signature) => {
...
+            |> Yojson.Safe.to_string
+            |> BLAKE2B.hash;
+          if (payload_hash |> Signature.verify(~signature)) {
+            Ok();
+          } else {
+            Error(`Failed_to_verify_payload);
+          };
+        | _ => Ok()
+        }
+      );
```
EduardoRFS, line 322: This change should definitely be in another PR.

EduardoRFS (changes requested):

rosalogia (approved):

node/flows.re
callistonianembrace: Opened https://github.com/marigold-dev/deku/pull/228 to address this

callistonianembrace:
@d4hines On further testing, it looks like the sender wallet argument isn't really needed. Wallet argument is only useful for the  `~sender` argument of `create_operation` - it should be derived from the identity of the issuing node itself (which happens to be the block producer)

bin/sidecli.re
```diff
@@ -749,8 +760,123 @@ let remove_trusted_validator = {
...
+    lwt_ret(
+      const(propose_validator_removal)
+      $ folder_node
+      $ validator_address,
+    )
+  );
+};
+
+// Run the CLI
+Mirage_crypto_rng_unix.initialize();
```
EduardoRFS, line 879: This is not needed anymore

EduardoRFS (approved):
@callistonianembrace can you test the entire pipeline before we merge this? Ask @Zett98 and @d4hines if you need help

anime girl happy with feature being merged

bin/sidecli.re
callistonianembrace: RNG isn't being properly initialised. I discovered this while testing.

EduardoRFS: Yes, fixed that on main

d4hines:
This PR will be superseded b #450

EduardoRFS:
Closing the PR solves nothing

PR #208, opened 2021-09-30
proto: represents validators with key_hashes rather than keys

rosalogia:
## Depends

- [x] #277 
- [x] #283 

## Problem

Validators on Deku are currently represented as public keys rather than as hashes of their public keys. As @d4hines noted in the comments, it's true that validators must inevitably reveal their public keys to sign blocks. However, according to @EduardoRFS, it's generally desirable among users for their public keys to be obscured wherever possible even if at some point they are unavoidably revealed.

## Solution

This PR
- changes any occurrences of `Wallet.t` (represented by public keys) that can be replaced by `Address.t` (represented by key_hashes) to avoid revealing public keys unnecessarily if/when possible
- changes the consensus contract to
  - represent validators as LIGO `key_hash`es instead of `key`s 
  - add an additional check that validator keys in a `root_hash_action` correspond to validator `key_hash`es in storage
  - account for the case where some keys are not present by representing validator keys as `option key list`
 
It's worth noting that a certain change made in the consensus contract swaps the position of two match cases due to a bug in the LIGO compiler that stopped the contract from compiling (the compiler would mismatch types and get confused); swapping the position of the match cases was proposed as a temporary work around.


rosalogia:
Update; `state.re` is now creating an `Address.t` from the `Address.key` field for the purpose of passing public keys to those two functions above; this seems to work, and there are no more type errors reported by the compiler. However, one unit test, [Consensus > Hash Validators](https://github.com/marigold-dev/deku/blob/main/tests/test_tezos_interop.re#L404) is failing. This is somewhat to be expected, the test body was updated to be consistent with the new representation but the assertion was not. Can someone confirm for me that the "received" value here aligns with what we now expect the test to result in?

image

EduardoRFS:
@rosalogia you need to recalculate that, my tip, look on how the hashing works on the consensus.mligo, this PR will require you to fix that also.

You can play with ligo at > https://ligolang.org/

Also if you want please enhance the test description.

rosalogia:
Looking at this bit of [`consensus.mligo`](https://github.com/marigold-dev/deku/blob/main/tezos_interop/consensus.mligo#L62), it seems like we just hash the packed list of public keys right now. I wonder what makes most sense for us if we stop representing validators with public keys and represent them with public key hashes instead:

1. Packing a list of key hashes and hashing that
2. Simply concatenating the list of key_hashes after stringifying them
3. Something else?

Currently in [`tezos_interop.re`](https://github.com/marigold-dev/deku/blob/17c2e2121883575ea04af42bdac7327a77247b34/tezos_interop/tezos_interop.re#L669) we're doing option 1. It's not really obvious from the contract (since it needs to be changed too) which way of doing things is "correct", nor can I tell if this is something we have a choice over, so I'm deferring to you guys to determine which way of hashing validators (once they're represented by key_hashes) makes the most sense.

I'd rather not hastily change the expected value of the test, lest I render the test totally useless in preventing issues going forward :smile: thanks for your time and help in advance!

EduardoRFS:
> Packing a list of key hashes and hashing that

This is actually the only feasible options as michelson doesn't expose ways to operate on key_hash.

.gitignore
```diff
@@ -1,4 +1,6 @@
 _esy
+_build
```
callistonianembrace, line 2: Can you help me understand how this folder was created? Were you using `esy dune ...` for something?

rosalogia: Yep! I've been looking for type errors with `esy dune build --watch @check @runtest`

callistonianembrace: I think you'll need this in the `devDependencies`

```json
    "esy-fswatch": "ulrikstrid/esy-fswatch:esy.json#7328e82b62c1691cd13623fffde0b66d0a0f9498",
```

EduardoRFS: It is safe, `esy b` is the build environment for production, same as used when releasing a package, `esy` is the build environment + current env, which is used for development. 

Also that doesn't create the `_build` probably it was something else.

callistonianembrace: `esy b` is meant for development too. command environment is not reproducible and running Dune in it is not recommended. `DUNE_BUILD_DIR` isn't set in cmd environment and explains the _build folder

callistonianembrace: > esy sh -c 'echo $DUNE_BUILD_DIR'
It's empty. Checkout `esy command-env` too

I don't understand why you think development environment should be any less reproducible then in the release mode.

EduardoRFS: > It's empty

It's not, if it is it's a bug, commad_env is a strict superset of the build environment.

DUNE_BUILD_DIR showing up on both places

> I don't understand why you think development environment should be any less reproducible then in the release mode.

Because development is hacking land and being reproducible is not desirable at all when you want to try things. Again

anchpop:
@rosalogia is this ready for review? I remember you mentioned you had some questions about ligo - feel free to ask them here or on slack and I'll do my best to answer

rosalogia:
Rebase is finished :smile: all yours, @EduardoRFS. Hopefully I didn't get confused and mess anything up, but if so, point it out and I'll fix it quickly.

tests/test_tezos_interop.re
```diff
@@ -530,15 +530,17 @@ describe("consensus", ({test, _}) => {
...
         key_exn("edpktq5HiqUkHTyoBQETvzbyaiwtKQkaBEPkwgZyfMqhRajRuLpWR7"),
         key_exn("edpkuNpThN8QeagEdvjN3o5R7PSic85cwiXHa61vNpRAE65FNV5mJH"),
-      ]);
+      ]
+      |> List.map(Key_hash.of_key)
+      |> hash_validators;
     let hash = BLAKE2B.to_string(hash);
     expect.string(hash).toEqual(
-      "546d2bb2375cc919efc81a103a7ad3bd1227546b320f275e357bd9a5d5eef946",
+      "6d6ecacbc858e3a89d87f0d9bd76b0c11b07aa95191129104395d17c6c96d36b",
```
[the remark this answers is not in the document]
EduardoRFS: I tested this against Ligo, makes sense

bin/deku_node.re
```diff
@@ -208,16 +208,17 @@ let node = (folder, trusted_validator_membership_change_file) => {
       current_validators
       |> List.mapi((i, validator) => {
            (
-             validator,
+             Address.of_wallet(validator),
```
EduardoRFS: That's wrong, you're fetching the validators from the storage, it will be an address already

EduardoRFS: Fixed because this PR is fixes some problems

bin/sidecli.re
```diff
@@ -532,8 +532,11 @@ let setup_identity = (node_folder, uri) => {
   let.await () = ensure_folder(node_folder);
 
   let identity = {
-    let (key, t) = Crypto.Ed25519.generate();
-    {uri, t: Ed25519(t), key: Ed25519(key)};
+    let (key, t) =
+      Crypto.Ed25519.generate()
+      |> (((k, t)) => (Secret.Ed25519(k), Key.Ed25519(t)));
```
EduardoRFS: This code could be way simpler.

bin/sidecli.re
```diff
@@ -692,16 +691,16 @@ let add_trusted_validator = (node_folder, address) => {
...
-    Format.fprintf(fmt, "%s", Protocol.Wallet.to_string(address));
+    Format.fprintf(fmt, "%s", Address.to_string(address));
   Arg.(conv((parser, printer)));
 };
 
 let validator_address = {
-  let docv = "validator_address";
-  let doc = "The validator address to be added/removed as trusted";
+  let docv = "validator_wallet";
+  let doc = "The validator wallet to be added/removed as trusted";
```
EduardoRFS: This change is wrong

EduardoRFS (approved):
Merging this because as it is important and I fixed a lot of nits. 

anime girl not that happy

d4hines (approved):
I've re-reviewed.
LGTM

d4hines:
Actually, in my testing, the chain isn't working - it won't get past block height 1 :(

d4hines:
I retested systematically with Ambika and we could not reproduce the error i was seeing. It was probably a mistake on my end. This PR lgtm!

PR #209, opened 2021-10-01
fix ticket comparison

EduardoRFS:
## Problem

There is a subtle bug on the `compare` function for tickets, which lead to it always being `== 0` even if the ticketer was different. Making so that one could get one unit of a ticket, make another ticket with the same data and that would fuse with the first ticket, allowing essentially to multiply it's own ticket amount and then essentially withdraw someone else ticket from the consensus contract.

## Solution

Fix the comparison and implement some tests to avoid any regression, after implementing I noticed that a test was actually broken but it was not triggering due to the bug above.

### DIFF
--- a/protocol/ticket.re
+++ b/protocol/ticket.re
@@ -9,7 +9,16 @@ type t =
 
 // TODO: should we expose to_yojson here?
 let compare = (a, b) => {
-  let to_string = a =>
-    Bytes.to_string(a.data) ++ Tezos_interop.Address.to_string(b.ticketer);
-  String.compare(to_string(a), to_string(b));
+  module T = {
+    [@deriving ord]
+    type t = {
+      ticketer: string,
+      data: bytes,
+    };
+    let of_t = (Ticket.{ticketer, data}) => {
+      ticketer: Address.to_string(ticketer),
+      data,
+    };
+  };
+  T.(compare(of_t(a), of_t(b)));
 };
--- a/tests/test_ledger.re
+++ b/tests/test_ledger.re
@@ -5,22 +5,26 @@ open Ledger;
 
 describe("ledger", ({test, _}) => {
   // TODO: maybe have a "total amount" function to ensure invariant?
-  let make_ticket = (~data=?, ()) => {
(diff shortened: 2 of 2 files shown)


protocol/ticket.re
```diff
@@ -9,7 +9,16 @@ type t =
 
 // TODO: should we expose to_yojson here?
 let compare = (a, b) => {
-  let to_string = a =>
-    Bytes.to_string(a.data) ++ Tezos_interop.Address.to_string(b.ticketer);
```
EduardoRFS, line 13: This is exactly why I hate functions with two idents with same type.

tests/test_ledger.re
```diff
@@ -108,14 +112,7 @@ describe("ledger", ({test, _}) => {
...
 
-    let t =
-      transfer(
-        ~source=a,
-        ~destination=c,
-        Amount.of_int(99),
-        make_ticket(~data=t1.data, ()),
-        t,
-      );
+    let t = transfer(~source=a, ~destination=c, Amount.of_int(99), t1, t);
```
anchpop, line 115: approving this ofc bc it's important but can you remind me why `99` is chosen here?

anchpop (approved):

tests/test_ledger.re
EduardoRFS: it's just all the balance available of ticket t1

protocol/ticket.re
callistonianembrace: I think this is a variable naming issue. Naming is always tricky and our language thankfully provides ways to avoid naming altogether. Also, we could avoid declaring pure functions in local scope.

callistonianembrace (approved):

protocol/ticket.re
EduardoRFS: Yes and avoiding naming makes it really hard to reason, point free style is mostly a regression from readability. 

> Also, we could avoid declaring pure functions in local scope.

Not so sure I agree here, this function is an implementation detail, it should not pollute the scope for everyone else.

PR #210, opened 2021-10-04
Add explicit equal fn to Tezos_interop

Zett98:
## Problem

We use polymorphic equality operators on tezos interop now

## Solution

This PR adds specialized equal fn to modules in Tezos_interop.

### DIFF
--- a/tests/test_tezos_interop.re
+++ b/tests/test_tezos_interop.re
@@ -23,8 +23,7 @@ describe("key", ({test, _}) => {
       of_string("edpkvDqjL7aXdsXSiK5ChCMAfqaqmCFWCv7DaT3dK1egJt136WBiT6"),
     ).
       toBe(
-      // TODO: proper equals
-      ~equals=(==),
+      ~equals=Key.equal,
       Some(edpk),
     )
   });
@@ -65,8 +64,7 @@ describe("key_hash", ({test, _}) => {
   });
   test("of_string", ({expect, _}) => {
     expect.option(of_string("tz1LzCSmZHG3jDvqxA8SG8WqbrJ9wz5eUCLC")).toBe(
-      // TODO: proper equals
-      ~equals=(==),
+      ~equals=Key_hash.equal,
       Some(tz1),
     )
   });
@@ -94,8 +92,7 @@ describe("secret", ({test, _}) => {
       of_string("edsk4bfbFdb4s2BdkW3ipfB23i9u82fgji6KT3oj2SCWTeHUthbSVd"),
     ).
       toBe(
-      // TODO: proper equals
-      ~equals=(==),
+      ~equals=Secret.equal,
       Some(edsk),
     )
   });
@@ -155,8 +152,7 @@ describe("signature", ({test, _}) => {
       ),
     ).
       toBe(
-      // TODO: proper equals
-      ~equals=(==),
+      ~equals=Signature.equal,
       Some(edsig),
     )
   });
@@ -195,8 +191,7 @@ describe("contract_hash", ({test, _}) => {
   });
   test("of_string", ({expect, _}) => {
     expect.option(of_string("KT1Dbav7SYrJFpd3bT7sVFDS9MPp4F5gABTc")).toBe(
-      // TODO: proper equals
-      ~equals=(==),
+      ~equals=Contract_hash.equal,
       Some(kt1),
     )
   });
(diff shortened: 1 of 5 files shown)


tests/test_tezos_interop.re
```diff
@@ -24,7 +24,7 @@ describe("key", ({test, _}) => {
     ).
       toBe(
       // TODO: proper equals
-      ~equals=(==),
```
callistonianembrace, line 27: Perhaps after the review is complete and the PR is getting merged, the TODO comment can be removed

tezos_interop/tezos_interop.re
```diff
@@ -131,6 +131,15 @@ module Key = {
     // TODO: move this to a functor
     obj1(req(name, raw_encoding));
   };
+  let to_cstruct =
+    fun
+    | Ed25519(pub_) => Mirage_crypto_ec.Ed25519.pub_to_cstruct(pub_);
+  let equal = (a, b) => {
+    switch (a, b) {
+    | (Ed25519(_) as a, Ed25519(_) as b) =>
```
callistonianembrace, line 139: I think the keys have to be unwrapped from the respective Tags/Variants and then converted to cstruct right away. Instead of passing down the wrapped key to `to_cstruct`. This will make it easier in future when we support more curves

tezos_interop/tezos_interop.re
```diff
@@ -172,7 +181,15 @@ module Key_hash = {
     switch (t) {
     | Key.Ed25519(pub_) => Ed25519(Ed25519.Hash.hash_key(pub_))
     };
-
+  let get_hash =
+    fun
+    | Ed25519(h) => h;
```
callistonianembrace, line 186: Same as my previous feedback

> I think the keys have to be unwrapped from the respective Tags/Variants and then converted to cstruct right away. Instead of passing down the wrapped key to to_cstruct. This will make it easier in future when we support more curves

You'll see how `get_hash` becomes redundant then. It'll simply be,

```reasonml
  let equal = (a, b) => {
    switch (a, b) {
    | (Ed25519(a), Ed25519(b)) =>
      Int.equal(BLAKE2B_20.compare(a, b), 0)
    };
  };
```

Zett98: updated

anchpop (approved):

tezos_interop/tezos_interop.re
EduardoRFS: I also prefer the simple one, I know it leads to slightly more duplicated code, but I also wonder, why this is not generated by `[@deriving eq]`?

tezos_interop/dune
```diff
@@ -2,7 +2,7 @@
  (name tezos_interop)
  (libraries helpers lwt.unix mirage-crypto-ec tezos-micheline)
  (preprocess
-  (pps ppx_deriving_yojson ppx_blob))
+  (pps ppx_deriving_yojson ppx_deriving.eq ppx_blob))
```
EduardoRFS, line 5: eq was added but not used, why is that the case?

EduardoRFS (changes requested):

tests/test_tezos_interop.re
Zett98: removed

tezos_interop/tezos_interop.re
Zett98: added explicit equal fn to Ed25519 module for public keys and secret keys and used ```@deriving eq``` here

tezos_interop/dune
Zett98: it was used in Address and Ticket.

EduardoRFS:
Made some changes, and it looks nice to me, @Zett98 can we merge it?

https://github.com/marigold-dev/deku/compare/af4f5ad4670228bb37973249ec34e8016812f44f..638df6c99e1f72f43300cb2cf0f93705b2fe16ed

EduardoRFS (approved):
anime girl excited for PR cleaning code

Zett98:
> Made some changes, and it looks nice to me, @Zett98 can we merge it?
> 
> https://github.com/marigold-dev/deku/compare/af4f5ad4670228bb37973249ec34e8016812f44f..638df6c99e1f72f43300cb2cf0f93705b2fe16ed

Merge it

PR #215, opened 2021-10-04
node: ensure add/remove validator signature belongs to node operator

callistonianembrace:
**Problem**
Membership change list operations are local to node and must be signed only by the node operator. This verification is mising.

**Solution**
This PR ensures that validator membership changes are signed only the expected node operator and no one else.


EduardoRFS:
Made some changes

callistonianembrace:
Thank you!

node/flows.re
```diff
@@ -419,6 +419,11 @@ let trusted_validators_membership =
   let {signature, payload: {address, action} as payload} = request;
   let payload_hash =
     payload |> payload_to_yojson |> Yojson.Safe.to_string |> BLAKE2B.hash;
+  let.assert () = (
+    `Invalid_signature_author,
+    Address.compare(state.Node.identity.t, Signature.public_key(signature))
+    == 0,
```
Zett98, line 425: Why not Int.equal? Polymorphic comparison is quite bad.(https://blog.janestreet.com/the-perils-of-polymorphic-compare)

Zett98 (changes requested):

anchpop (approved):

d4hines:
This PR looks good to me. 

However, I had a bit of tricky time determining the context. `trusted_validators_membership` is poorly named. Every other function in the repo is named like `[verb]_[direct object of verb]`, but` trusted_validators_membership` doesn't match this format. Reading the code in context I figured out that it handles adding and removing validators. Maybe we could name it something like `change_trusted_validator_membership`?

d4hines (approved):

node/flows.re
EduardoRFS: Polymorphic comparison is mostly a problem for structured data structures, but I also agree that `==` should be Int.equal like on Base. #217

EduardoRFS (approved):

EduardoRFS:
anime girl scared of security problems like this one

Zett98 (approved):

Issue #199, opened 2021-09-23
support tz2 on tezos_interop


EduardoRFS:
Subst of #22, tz2 uses secp256k1 which I believe are not directly supported by mirage-crypto-ec.

- [x] Ask the authors of mirage-crypto-ec about it

EduardoRFS:
Indeed secp256k1 is not supported by mirage-crypto-ec, maybe we should switch to hacl-star.

Zett98:
> Indeed secp256k1 is not supported by mirage-crypto-ec, maybe we should switch to hacl-star.

Hacl-star doesn't support it either, but there is a library from nomadic labs

PR #219, opened 2021-10-05
add a dependency on secp256k1 library for tz2 interop

Zett98:
## Problem

We need to add a support for interop with tz2, but our current library mirage-crypto doesn't support it.

## Solution

Add a library for secp256k1 by nomadic labs as a dependency for our project

## Related
- #199


callistonianembrace (approved):

EduardoRFS (approved):
anime girl surprised on how small is secp256k1 library

PR #221, opened 2021-10-07
move ed25519 to separate crypto module

pkhry:
## Problem

We will be adding more crypto modules to Tezos_interop, so keeping them in the same file as the interop file is bad
## Solution

Move ed25519 module to separate crypto module
## Related
 - #22


callistonianembrace (approved):
LGTM. Looks like pure refactoring.

PR #223, opened 2021-10-08
remove dead command setup-node

EduardoRFS:
## Problem

The command `setup-node` was replaced a month ago by `setup-identity`, `setup-tezos` and the `./setup.sh` and it is now not used.

## Solution

As this was never actually used much, I propose that we just remove this command.


d4hines (approved):
LGTM

PR #224, opened 2021-10-08
remove duplicated file names on sidecli

EduardoRFS:
## Depends

- [x] #223 

## Problem

The sidecli code is duplicating the JSON file names in many places, this is problematic as there is no source of truth for it.

## Solution

This is not a proper solution but just an improvement, by making so that those file names appear only once and centralized on the sidecli.re file.


d4hines (approved):
LGTM

PR #218, opened 2021-10-05
Design doc for async state hashing

d4hines:
## Problem 
Design doc for async state hashing requested in  #147.

## Solution
This PR adds the design doc for this feature. This PR will not be merged, but only serves as a place to collect thoughts and feedback on the design.

See PR's #195 for implementation.

PR #227, opened 2021-10-11
Delayed State Hashing

d4hines:
## Depends

- [x] #308
- [x] #309
- [x] #398 

## Problem

Per #147, we want async state hashing. We have a design in #218; however, the problem remains a complex one, and we need to break apart the design in to incremental steps. This PR avoids asynchronicity, and instead introduces the smallest step possible in the right direction: the delayed state hashing.

## Solution

Recall that #218 introduces the new terminology of "state root epoch" to refer to the interval (in blocks) between state root hash updates. This PR delays the state root hash sent by one epoch, such that the next state root hash N is calculated during epoch N - 1 (refer to #218 for diagrams illustrating this).

To achieve this, we extend the protocol state with a fields `current_epoch` and `next_epoch`. `current_epoch` is merely an integer identifier for the epoch (that will be used in subsequent PR's). `next_epoch` is record defined to be the state root hash of the next epoch along with the data that produced that hash. The field acts like an expectation: when applying a block with a different state root hash than the current one (i.e, when the epoch changes), it _must_ match `next_epoch.state_root_hash`, else the block will be rejected. 

Updating this field is simple (at least when done synchronously): when applying a block with a new state root hash, update `next_epoc` to be the hash the current state (along with its data as a JSON string). This achieves our inductive definition: at epoch N - 1, we generate the hash of epoch N, and so on. Additionally, increment `current_epoch`. 

Of course, we need a base case. Let epoch of the empty chain (i.e, before appyling the genesis block) be `-1`. Let the state root hash associated with epoch -1 be the hash of some arbitrary data (not reachable as a protocol state). At epoch -1, we know that the next block will be the genesis block, and so we can predetermine the hash of epoch 0. Thus, at the initial state, we can set `next_epoch.state_root_hash` to the state root hash of the genesis block.

**Note to reviewers** - please make sure I didn't break snapshots etc. by adding the two fields `current_epoch` and `next_epoch`.

## Related

- #147
- #218
- #195 (this PR is a better first PR than #195 - I'll change 195 to depend on this one).

### DIFF
--- a/crypto/incremental_patricia.re
+++ b/crypto/incremental_patricia.re
@@ -33,6 +33,7 @@ module Make =
 
   let is_set = (bit, number) => 1 lsl bit land number != 0;
 
+  // Hash: 0e5751c026e543b2e8ab2eb06099daa1d1e5df47778f7787faab45cdf12fe3a8
   let empty_hash = BLAKE2B.hash("");
   let hash_of_t =
     fun
--- a/node/building_blocks.re
+++ b/node/building_blocks.re
@@ -56,6 +56,29 @@ let is_current_producer = (state, ~key_hash) => {
   Some(current_producer.address == key_hash);
 };
 
+// TODO: verify reasonable times for this with benchmarking.
+// currently set to [state_root_min_timeout] as a reasonable
+// default.
+let minimum_signable_time_between_epochs = 10.0;
+
+// TODO: verify reasonable times for this with benchmarking.
+// Currently set to [minimum_signable_time_between_epochs] + 10
+// as a reasonable default.
+let maximum_signable_time_between_epochs = 20.0;
+
+let block_has_signable_state_root_hash = (~current_time, state, block) => {
+  let protocol = state.Node.protocol;
+  let time_since_last_epoch = current_time -. protocol.last_state_root_update;
+  Crypto.(
+    if (BLAKE2B.equal(block.Block.state_root_hash, protocol.state_root_hash)) {
+      time_since_last_epoch <= maximum_signable_time_between_epochs;
+    } else {
+      BLAKE2B.equal(block.state_root_hash, state.next_state_root |> fst)
+      && time_since_last_epoch >= minimum_signable_time_between_epochs;
+    }
+  );
+};
+
 // TODO: bad naming
 // TODO: check if block must have published a new snapshot
 let is_signable = (state, block) => {
@@ -103,17 +126,50 @@ let is_signable = (state, block) => {
   && !is_signed_by_self(state, ~hash=block.hash)
   && is_current_producer(state, ~key_hash=block.author)
   && !has_next_block_to_apply(state, ~hash=block.hash)
-  && all_operations_are_trusted;
+  && all_operations_are_trusted
+  && block_has_signable_state_root_hash(~current_time, state, block);
 };
 
 let sign = (~key, block) => Block.sign(~key, block);
 
-let produce_block = state =>
+/** Calculates whether to start sending a new state root hash.
+
+    The state root epoch is the interval (in blocks) between state root
+    hash updates. Thus, a new epoch is triggered by applying a block with
+    a new state root hash. The block producer decides when to send
+    blocks with new state root hashes. To enforce that he does so on time,
+    validators reject blocks with updates that occur to soon or too late
+    (see [Protocol.apply]).
+
+    The block producer uses this function to determine when to send a
+    block with an updated state root hash.
+*/
+let should_start_new_epoch = (last_state_root_update, current_time) => {
+  let avoid_jitter = 1.0;
+  current_time
+  -. last_state_root_update
+  -. avoid_jitter >= minimum_signable_time_between_epochs;
+};
+
+let produce_block = state => {
+  let start_new_epoch =
+    should_start_new_epoch(
+      state.Node.protocol.last_state_root_update,
+      Unix.time(),
+    );
+  let next_state_root_hash =
+    if (start_new_epoch) {
+      Some(state.Node.next_state_root |> fst);
+    } else {
+      None;
+    };
   Block.produce(
     ~state=state.Node.protocol,
     ~author=state.identity.t,
+    ~next_state_root_hash,
     ~operations=state.pending_operations,
   );
+};
 
 let is_valid_block_height = (state, block_height) =>
   block_height >= 1L && block_height <= state.Node.protocol.block_height;
--- a/node/flows.re
+++ b/node/flows.re
@@ -187,6 +187,15 @@ let rec try_to_apply_block = (state, update_state, block) => {
     `Block_not_signed_enough_to_apply,
     Block_pool.is_signed(~hash=block.Block.hash, state.Node.block_pool),
   );
+
+  let (next_state_root_hash, _) = state.next_state_root;
+  // TODO: in the future, we should stop the chain if this assert fails
+  let.assert () = (
+    `Invalid_state_root_hash,
+    BLAKE2B.equal(state.protocol.state_root_hash, block.state_root_hash)
+    || BLAKE2B.equal(next_state_root_hash, block.state_root_hash),
+  );
+
   let.ok state = apply_block(state, update_state, block);
   reset_timeout^();
   let state = clean(state, update_state, block);
--- a/node/state.re
+++ b/node/state.re
@@ -23,6 +23,7 @@ type t = {
   block_pool: Block_pool.t,
   protocol: Protocol.t,
   snapshots: Snapshots.t,
+  next_state_root: (BLAKE2B.t, string),
   // networking
   // TODO: move this to somewhere else but the string means the nonce needed
   // TODO: someone right now can spam the network to prevent uri changes
@@ -54,10 +55,10 @@ let make =
   let initial_block_pool =
     Block_pool.make(~self_key=identity.key)
     |> Block_pool.append_block(initial_block);
-  let initial_snapshots = {
-    let initial_snapshot = Protocol.hash(initial_protocol);
+  let initial_snapshot = Protocol.hash(initial_protocol);
+  let initial_snapshots =
     Snapshots.make(~initial_snapshot, ~initial_block, ~initial_signatures);
-  };
+
   {
     identity,
     trusted_validator_membership_change,
@@ -67,6 +68,7 @@ let make =
     block_pool: initial_block_pool,
     protocol: initial_protocol,
     snapshots: initial_snapshots,
+    next_state_root: initial_snapshot,
     // networking
     uri_state: Uri_map.empty,
     validators_uri: initial_validators_uri,
@@ -120,6 +122,26 @@ let try_to_commit_state_hash = (~old_state, state, block, signatures) => {
     );
   });
 };
+
+let write_data_to_file = (path, protocol) => {
+  let protocol_bin = Marshal.to_string(protocol, []);
+  Lwt.async(() =>
+    Lwt_io.with_file(
+      ~mode=Output,
+      path,
+      oc => {
+        let.await () = Lwt_io.write(oc, protocol_bin);
+        Lwt_io.flush(oc);
+      },
+    )
+  );
+};
+
+let write_state_to_file = (~data_folder, protocol) =>
+  write_data_to_file(data_folder ++ "/state.bin", protocol);
+let write_prev_epoch_state_to_file = (~data_folder, protocol) =>
+  write_data_to_file(data_folder ++ "/prev_epoch_state.bin", protocol);
+
 let apply_block = (state, block) => {
   let old_state = state;
   let.ok (protocol, new_snapshot, receipts) =
@@ -130,28 +152,29 @@ let apply_block = (state, block) => {
       state.recent_operation_receipts,
       receipts,
     );
-  let state = {...state, protocol, recent_operation_receipts};
-  Lwt.async(() =>
-    Lwt_io.with_file(
-      ~mode=Output,
-      state.data_folder ++ "/state.bin",
-      oc => {
-        let protocol_bin = Marshal.to_string(state.protocol, []);
-        let.await () = Lwt_io.write(oc, protocol_bin);
-        Lwt_io.flush(oc);
-      },
-    )
-  );
+  let next_state_root =
+    new_snapshot |> Option.value(~default=state.next_state_root);
+  let state = {
+    ...state,
+    protocol,
+    next_state_root,
+    recent_operation_receipts,
+  };
+  write_state_to_file(~data_folder=state.data_folder, state.protocol);
   switch (new_snapshot) {
-  | Some(new_snapshot) =>
+  | Some(_) =>
     switch (Block_pool.find_signatures(~hash=block.hash, state.block_pool)) {
     | Some(signatures) when Signatures.is_self_signed(signatures) =>
       try_to_commit_state_hash(~old_state, state, block, signatures)
     | _ => ()
     };
+    write_prev_epoch_state_to_file(
+      ~data_folder=state.data_folder,
+      old_state.protocol,
+    );
     let snapshots =
       Snapshots.update(
-        ~new_snapshot,
+        ~new_snapshot=old_state.next_state_root,
         ~applied_block_height=state.protocol.block_height,
         state.snapshots,
       );
@@ -210,14 +233,18 @@ let load_snapshot =
   );
   let.assert () = (
     `State_root_not_the_expected,
-    // TODO: this List.hd will not fail, but it makes me anxious
-    state_root_hash == List.hd(all_blocks).state_root_hash,
+    // TODO: It may not hold in the future that the last block's
+    // state root hash is equal to this state root's hash. E.g.,
+    // we may send blocks from the requested epoch and future ones.
+    // In the future, the equivalent check should be something like this:
+    // let hd_srh == List.hd(all_blocks).state_root_hash;
+    // state_root_hash == List.find(b => b.srh != hd_srh)
+    state_root_hash == last_block.state_root_hash,
   );
   let.assert () = (
     `Snapshots_with_invalid_hash,
     BLAKE2B.verify(~hash=state_root_hash, state_root),
   );
-
   let of_yojson = [%of_yojson:
     (
       Core.State.t,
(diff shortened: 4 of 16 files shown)


protocol/block.re
```diff
@@ -156,30 +156,14 @@ let genesis =
...
+  let state_root_hash =
+    Option.value(~default=state.State.state_root_hash, next_state_root_hash);
   make(
     ~previous_hash=state.State.last_block_hash,
-    ~state_root_hash=
-      update_state_hashes ? fst(State.hash(state)) : state.state_root_hash,
+    ~state_root_hash,
     ~handles_hash=Ledger.handles_root_hash(state.ledger),
-    ~validators_hash=
-      update_state_hashes
```
d4hines, line 181: This is missing from my implementation, which breaks current behavior - we can't merge till this is fixed.

As the simplest fix we could hash the validators whenever we hash the state to maintain the current behavior. In the future we might want to do something more sophisticated.

d4hines: Captured in issue https://github.com/marigold-dev/deku/issues/230

d4hines: Addressed in https://github.com/marigold-dev/deku/pull/227/commits/e7c67e0fbc87c8580d7388a258da5b0bc5f55222

callistonianembrace (commented):
@d4hines Could you please elaborate the bug that commit https://github.com/marigold-dev/deku/pull/227/commits/e7c67e0fbc87c8580d7388a258da5b0bc5f55222 addresses in the commit message? But the looks of it, I think it takes into account the hash of the validators too. Good change. Thanks for catching it.

node/building_blocks.re
```diff
@@ -55,6 +55,47 @@ let is_current_producer = (state, ~key) => {
   Some(current_producer.address == key);
 };
 
+/** Blocks that change the state root epoch before
```
anchpop, line 58: Newb question, what is the state root epoch?

d4hines:
@EduardoRFS , rebased.

protocol/protocol.re
```diff
@@ -169,7 +169,12 @@ let make = (~initial_block) => {
        are in the right place, otherwise invariants
        can be broken */
     last_block_hash: initial_block.Block.previous_hash,
-    state_root_hash: initial_block.Block.state_root_hash,
+    // The initial protocol state includes an unreachable
+    // hash as the first state root hash. When the genesis
+    // block is applied, the the state root hash of the block
+    // will be different, triggering the creation of the next hash.
+    // See [FIXME: add ref].
+    state_root_hash: BLAKE2B.hash("Tetsutetsu Tetsutetsu"),
```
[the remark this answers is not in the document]
d4hines: @anchpop, correct - `make` is for a new protocol.

In general we want to constrain our types to match reality/the semantics as much as possible (as long as it doesn't become too cumbersome to work with). `Option(BLAKE2B.t)` is too broad - it's not valid for any subsequent state to have the value `None`. Better would be a GADT, where functions that operate on the state root hash specify a non-empty state root hash. 

But while a GADT would be the most precises type, in our case I think it's overkill. We can avoid the mental overhead of a GADT simply by keeping the type `BLAKE2B.t` and choosing an "unreachable" hash - i.e one that could never be reached through a normal state transition.

Does that make any more sense?

protocol/block.rei
```diff
@@ -19,9 +19,16 @@ type t =
 let sign: (~key: Address.key, t) => Signature.t;
 let verify: (~signature: Signature.t, t) => bool;
 let genesis: t;
+
+type next_hashes = {
+  state_root: BLAKE2B.t,
+  validators: BLAKE2B.t,
+};
```
EduardoRFS, line 26: I believe next_hashes is not needed anymore after #246 and it can be replaced by next_state_root_hash parameter on produce

protocol/protocol.re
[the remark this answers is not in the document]
EduardoRFS: yup that's the entire reasoning about having values which are not reachable as default, it drastically simplifies the thought process for understanding this simple operations.

EduardoRFS (changes requested):
Small rebase needed but this PR looks great.

node/building_blocks.re
```diff
@@ -55,6 +55,47 @@ let is_current_producer = (state, ~key) => {
...
+      state root hash and the maximum_signable_time_between_epochs
+      has not elapsed.
+   - Or the block has the same state root hash as [next_state_root_hash]
+     and the minimum_signable_time_between_epochs as elapsed.
+   */
+let block_has_signable_state_root_hash = (~current_time, state, block) => {
+  let protocol = state.Node.protocol;
+  let time_since_last_epoch = protocol.last_state_root_update -. current_time;
+
+  let same_epoch = protocol.state_root_hash == block.Block.state_root_hash;
```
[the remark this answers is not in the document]
d4hines: Fixed

d4hines:
With help from @EduardoRFS I've debugged this PR. It's now tested and working.

There are a few fixes we had to do that are not directly related to this change. They are in separate commits, and I will turn them into separate PR's tomorrow

d4hines:
I haven't gotten to @Zett98 's comments about simplifying the logic in building_blocks.re, but will do that tomorrow too.

node/building_blocks.re
```diff
@@ -96,18 +137,69 @@ let is_signable = (state, block) => {
...
+  -. avoid_jitter >= state_root_min_timeout;
+};
+
+let produce_block = state => {
+  let start_new_epoch =
+    should_start_new_epoch(
+      state.Node.protocol.last_state_root_update,
+      Unix.time(),
+    );
+  let next_hashes =
```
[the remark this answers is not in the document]
d4hines: fixed in latest commit

node/building_blocks.re
```diff
@@ -55,6 +55,47 @@ let is_current_producer = (state, ~key) => {
...
+  let protocol = state.Node.protocol;
+  let time_since_last_epoch = protocol.last_state_root_update -. current_time;
+
+  let same_epoch = protocol.state_root_hash == block.Block.state_root_hash;
+  let next_epoch = state.next_state_root_hash == block.state_root_hash;
+  let min_epoch_timeout =
+    time_since_last_epoch >= minimum_signable_time_between_epochs;
+  let max_epoch_timeout =
+    time_since_last_epoch <= maximum_signable_time_between_epochs;
+  switch (same_epoch, next_epoch, min_epoch_timeout, max_epoch_timeout) {
```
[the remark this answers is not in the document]
d4hines: I'll address this tomorrow

protocol/block.rei
d4hines: fixed in latest commit

protocol/block.re
d4hines: This is handled by #246

EduardoRFS:
@d4hines CI is not green

d4hines:
> @d4hines CI is not green

Fixed! Also cleaned up the history.

node/building_blocks.re
```diff
@@ -55,6 +55,47 @@ let is_current_producer = (state, ~key) => {
...
+      has not elapsed.
+   - Or the block has the same state root hash as [next_state_root_hash]
+     and the minimum_signable_time_between_epochs as elapsed.
+   */
+let block_has_signable_state_root_hash = (~current_time, state, block) => {
+  let protocol = state.Node.protocol;
+  let time_since_last_epoch = protocol.last_state_root_update -. current_time;
+
+  let same_epoch = protocol.state_root_hash == block.Block.state_root_hash;
+  let next_epoch = state.next_state_root_hash == block.state_root_hash;
```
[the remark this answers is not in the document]
d4hines: I've made this much more clear in the latest diff. Check it out and see if that's better.

node/building_blocks.re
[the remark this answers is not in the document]
d4hines: I've made this much more clear in the latest diff. Check it out and see if that's better. @Zett98

anchpop (approved):

EduardoRFS (changes requested):
As mentioned at slack, couple comments are wrong and I would ask you to remove them.

node/state.re
```diff
@@ -76,6 +77,7 @@ let make =
     block_pool: initial_block_pool,
     protocol: initial_protocol,
     snapshots: initial_snapshots,
+    next_state_root_hash: BLAKE2B.hash("Tetsutetsu Tetsutetsu"),
```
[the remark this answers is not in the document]
d4hines: Yes, definitely. Fixed in latest diff.

node/building_blocks.re
```diff
@@ -55,6 +55,47 @@ let is_current_producer = (state, ~key_hash) => {
...
+let minimum_signable_time_between_epochs = 60.0;
+/** Blocks that do not change the state root epoch after
+    this interval has passed are not signable.
+
+    TODO: verify reasonable times for this with benchmarking.
+    Currently set to [minimum_signable_time_between_epochs] + 10
+    as a reasonable default. */
+let maximum_signable_time_between_epochs = 70.0;
+
+/* Returns true if the block's state root hash and has not
```
Zett98, line 73: > if the block's state root hash and has not

is "and" not needed here? I'm confused a bit

node/state.re
```diff
@@ -145,7 +148,17 @@ let apply_block = (state, block) => {
       state.recent_operation_results,
       results,
     );
-  let state = {...state, protocol, recent_operation_results};
+  let next_state_root_hash =
```
Zett98, line 151:
```reasonml
new_snapshot
|> Option.map(fst)
|> Option.value(~default=state.next_state_root_hash)
````

File: protocol/protocol.re [hunk 12]
```diff
@@ -168,24 +168,33 @@ let make = (~initial_block) => {
        are in the right place, otherwise invariants
        can be broken */
     last_block_hash: initial_block.Block.previous_hash,
-    state_root_hash: initial_block.Block.state_root_hash,
-    last_state_root_update: 0.0,
+    // The initial protocol state includes an unreachable
+    // hash as the first state root hash. When the genesis
+    // block is applied, the the state root hash of the block
```
Zett98, line 173: then the?

protocol/protocol.re
```diff
@@ -168,24 +168,33 @@ let make = (~initial_block) => {
...
     last_applied_block_timestamp: 0.0,
     last_seen_membership_change_timestamp: 0.0,
   };
   apply_block(empty, initial_block) |> fst;
 };
 let apply_block = (state, block) => {
   let.assert () = (`Invalid_block_when_applying, is_next(state, block));
-  let (valid_hash, hash) =
+  let hash =
     if (block.state_root_hash == state.state_root_hash) {
```
Zett98, line 188: please remove polymorphic equality

node/state.re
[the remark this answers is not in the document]
d4hines: Fixed in the diff I'm about to push.

node/state.re
```diff
@@ -287,5 +302,10 @@ let load_snapshot =
       all_blocks,
     );
   //TODO: snapshots?
+  // It's ok for the [next_state_root_hash] to be incorrect
+  // here, because by definition when we load a snapshot we're
+  // out of sync and can't sign blocks anyway. On the next state
+  // root epoch, we'll update the [next_state_root_hash] to the
+  // correct value, at which point we'll also be in sync.
```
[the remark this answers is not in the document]
d4hines: Not sure. We may need to remove the check in `try_to_apply_block`.

node/building_blocks.re
d4hines: You are correct. I've fixed the comment.

protocol/protocol.re
d4hines: Fixed in latest diff.

protocol/protocol.re
d4hines: Fixed in latest diff.

node/state.re
d4hines: Fixed in latest diff.

node/state.re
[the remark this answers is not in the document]
d4hines: Ok fixed now.

node/state.re
```diff
@@ -270,7 +283,9 @@ let load_snapshot =
       block_height,
       last_block_hash,
       state_root_hash,
-      last_state_root_update: 0.0,
+      // The node uses this time to check that state root hash updates
+      // are within an acceptable time frame, and thus needs to be current.
```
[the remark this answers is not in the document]
d4hines: I've updated this comment to be more clear.

Zett98 (approved):

node/building_blocks.re
```diff
@@ -96,18 +119,62 @@ let is_signable = (state, block) => {
...
-  && contains_only_trusted_add_validator_op(block.side_chain_ops);
+  && contains_only_trusted_add_validator_op(block.side_chain_ops)
+  && block_has_signable_state_root_hash(~current_time, state, block);
 };
 
 let sign = (~key, block) => Block.sign(~key, block);
 
-let produce_block = state =>
+/** The state_root_min_timeout determines how often the block producer
+    starts hashing a new state root. It is closely related to the
```
EduardoRFS, line 129: This comment is kind of weird, on the async version it doesn't determines, but here it does, but it does determines when it hashes, not when it starts hashing. Slightly different.

node/building_blocks.re
```diff
@@ -55,6 +55,29 @@ let is_current_producer = (state, ~key_hash) => {
...
+
+// TODO: verify reasonable times for this with benchmarking.
+// Currently set to [minimum_signable_time_between_epochs] + 10
+// as a reasonable default.
+let maximum_signable_time_between_epochs = 70.0;
+
+let block_has_signable_state_root_hash = (~current_time, state, block) => {
+  let protocol = state.Node.protocol;
+  let time_since_last_epoch = current_time -. protocol.last_state_root_update;
+  Crypto.(
```
d4hines, line 71: Let's change this to `open Crypto;`

node/building_blocks.re
```diff
@@ -96,18 +119,62 @@ let is_signable = (state, block) => {
...
 
-let produce_block = state =>
+/** The state_root_min_timeout determines how often the block producer
+    starts hashing a new state root. It is closely related to the
+    state root epoch (see notes in [should_start_new_epoch]]).
+
+    The faster hashing is, the shorter this time should be. Faster updates
+    to the state root hash means shorter times to join the network, since a
+    node only needs to download and apply the blocks from the previous state root
+    update to the current block.
```
EduardoRFS, line 135: "since a node needs to wait a state root hash change on Tezos before being considered in sync"

node/building_blocks.re
```diff
@@ -96,18 +119,62 @@ let is_signable = (state, block) => {
...
+let state_root_min_timeout = 60.0;
+
+/** Calculates whether to start sending a new state root hash.
+
+    The state root epoch is the interval (in blocks) between state root
+    hash updates. Thus, a new epoch is triggered by applying a block with
+    a new state root hash. The block producer decides when to send
+    blocks with new state root hashes. To enforce that he does so on time,
+    validators reject blocks with updates that occur to soon or too late
+    (see [Protocol.apply]).
```
EduardoRFS, line 146: The validator still applies the block if needed, he just doesn't sign it.

Probably not this function, I believe it will be `Flows.is_signable`.

node/state.re
```diff
@@ -270,7 +282,10 @@ let load_snapshot =
       block_height,
       last_block_hash,
       state_root_hash,
-      last_state_root_update: 0.0,
+      // The node will only sign blocks that have an
+      // acceptable state root hash relative to the last_state_root_update.
+      // see [building_blocks.re].
+      last_state_root_update: Unix.time(),
```
EduardoRFS, line 288: This should not be Unix.time, as otherwise the node may sign a block that contains an old state_root_hash even if it's already too late.

node/state.re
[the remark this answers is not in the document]
EduardoRFS: That is a no-go, the check is definitely needed.

node/flows.re
```diff
@@ -187,6 +187,12 @@ let rec try_to_apply_block = (state, update_state, block) => {
     `Block_not_signed_enough_to_apply,
     Block_pool.is_signed(~hash=block.Block.hash, state.Node.block_pool),
   );
+  let.assert () = (
```
EduardoRFS, line 190:
```suggestion
  // TODO: in the future if this is triggered we should stop the node
  let.assert () = (
```

protocol/protocol.re
```diff
@@ -168,26 +167,31 @@ let make = (~initial_block) => {
...
        can be broken */
     last_block_hash: initial_block.Block.previous_hash,
-    state_root_hash: initial_block.Block.state_root_hash,
-    last_state_root_update: 0.0,
+    // The initial protocol state includes an unreachable
+    // hash as the first state root hash. When the genesis
+    // block is applied, then the state root hash of the block
+    // will be different, triggering the creation of the next hash.
+    // Hash: 4ba060ef0b65c3690407175d7bd0e09975f9d8f5907fb9e99de27ea4de804652
+    state_root_hash: Crypto.BLAKE2B.hash("Tetsutetsu Tetsutetsu"),
```
EduardoRFS, line 175: add `open Crypto` on top of the file instead.

node/state.re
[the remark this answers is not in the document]
d4hines: We can calculate the next state root hash by loading the received snapshot and hashing the resultant state.

protocol/protocol.re
```diff
@@ -168,26 +167,31 @@ let make = (~initial_block) => {
...
+    // The node only signs blocks with state root hashes that are neither too
+    // early nor too late relative to [last_state_root_update].
+    last_state_root_update: Unix.time(),
```
EduardoRFS, line 178: This should definitely be 0.0

EduardoRFS (changes requested):

crypto/incremental_patricia.re
```diff
@@ -33,6 +33,7 @@ module Make =
 
   let is_set = (bit, number) => 1 lsl bit land number != 0;
 
+  // Hash: 0e5751c026e543b2e8ab2eb06099daa1d1e5df47778f7787faab45cdf12fe3a8
```
rosalogia, line 36: What's the purpose of this?

node/building_blocks.re
```diff
@@ -55,6 +55,29 @@ let is_current_producer = (state, ~key_hash) => {
...
+// Currently set to [minimum_signable_time_between_epochs] + 10
+// as a reasonable default.
+let maximum_signable_time_between_epochs = 70.0;
+
+let block_has_signable_state_root_hash = (~current_time, state, block) => {
+  let protocol = state.Node.protocol;
+  let time_since_last_epoch = current_time -. protocol.last_state_root_update;
+  Crypto.(
+    if (BLAKE2B.equal(block.Block.state_root_hash, protocol.state_root_hash)) {
+      time_since_last_epoch <= maximum_signable_time_between_epochs;
```
rosalogia, line 73: What exactly are these two conditions and how/why do they guarantee a signable state root hash? As far as I can tell, looks like here we're saying:

If a block's state root hash is the same as the state root hash published by Tezos (? assuming that's what `protocol.state_root_hash` is), we have a signable state root hash if the time since the last epoch is less than or equal to the maximum.

Otherwise, that is, if the block's state root hash and the published state root hash differ, we have a signable state root hash if the block's state root hash is equal to the next state root hash held by the state, and we've exceeded the time since the last epoch.

This seems to reduce to "If we're consistent with the published SRH and aren't late to the next epoch, we may sign the SRH. Otherwise, if we're consistent with the next SRH and have exceeded the time allotted to the last epoch, we may sign the SRH.

Is this correct? Is my use of language or my comprehension faulty?

crypto/incremental_patricia.re
d4hines: It's to aid debugging. It's useful to know the hash of our hard-coded states so you can recognize them if you print them while debugging.

d4hines:
I've finished testing this PR. It now depends on #398 . 

For testing, I set the minimum state root epoch time to 10s, and modified start.sh to sleep a random number of seconds between 10 and 20 before signing the initial block to simulate arbitrary delays.

I then let the chain produce a few state root hash updates and checked on better-call.dev to see that they were being posted. I then shut down the chain, waited a few minutes, started it up agin, and checked better-call.dev again.

node/building_blocks.re
d4hines: What I'm trying to say is:

You may sign the block iff
- Either, the SRH of the block matches the current SRH you have AND the epoch change is not late (i.e, the time between now and the last state root update is less than the `maximum_signable_time_between_epochs`)
- Or, the SRH matches the `next_state_root_hash` you calculated previously, and the epoch change is not too early (i.e the time between now and the last state root update is greater than the `minimum_signable_time_between_blocks`)

Does that make sense?

rosalogia: Totally, thanks

EduardoRFS (approved):
anime girl supper happy with this PR being merged

PR #232, opened 2021-10-13
Lift state hashing to the node

d4hines:
## Depends

- [x] #227

## Problem

Per #147, we're trying to achieve async state hashing. But this is a complex problem (see #218). 

Before we make hashing async, we first need to lift the process of hashing out of the protocol and into the node. 

Remember, `./protocol` does not actually do consensus - rather, the `Protocol` library contains just the core state information  that is guaranteed to be agreed upon by every honest in-sync node. Therefore, any consensus checking should be done in `./node` - any checking done in `./protocol` is only as a redundancy/failsafe.

## Solution

This change should not modify the behavior in any way from #227 (reviewers, help me check this!). All it does is move the hashing process to `Node.State.apply_block`. In the next  PR, we'll introduce (partial) asynchronicity. 

## Related

- #147

### DIFF
--- a/node/state.re
+++ b/node/state.re
@@ -77,51 +77,6 @@ let make =
   };
 };
 
-let commit_state_hash = state =>
-  Tezos_interop.Consensus.commit_state_hash(~context=state.interop_context);
-let try_to_commit_state_hash = (~old_state, state, block, signatures) => {
-  let signatures_map =
-    signatures
-    |> Signatures.to_list
-    |> List.map(signature => {
-         let address = Signature.address(signature);
-         let key = Signature.public_key(signature);
-         let signature = Signature.signature(signature);
-         (address, (key, signature));
-       })
-    |> List.to_seq
-    |> Address_map.of_seq;
-
-  let validators =
-    state.protocol.validators
-    |> Validators.to_list
-    |> List.map(validator =>
-         Address.to_key_hash(validator.Validators.address)
-       );
-  let signatures =
-    old_state.protocol.validators
-    |> Validators.to_list
-    |> List.map(validator => validator.Validators.address)
-    |> List.map(address => Address_map.find_opt(address, signatures_map));
-
-  Lwt.async(() => {
-    /* TODO: solve this magic number
-       the goal here is to prevent a bunch of nodes concurrently trying
-       to update the state root hash */
-    let.await () =
-      state.identity.t == block.Block.author
-        ? Lwt.return_unit : Lwt_unix.sleep(120.0);
-    commit_state_hash(
-      state,
-      ~block_height=block.block_height,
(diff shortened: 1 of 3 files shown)


anchpop (commented):
I don't fully understand the problem but this change makes senes to me. we need to make a flowchart and keep it up to date lol

anchpop (approved):

node/state.re
```diff
@@ -160,8 +150,10 @@ let apply_block = (state, block) => {
       },
     )
   );
-  switch (new_snapshot) {
-  | Some(new_snapshot) =>
+  if (block.state_root_hash == old_state.protocol.state_root_hash) {
+    Ok(state);
+  } else {
+    let new_snapshot = Protocol.hash(state.protocol);
```
EduardoRFS, line 156: This may be problematic, before we hashed the state before the block application and now we're hashing it after, I'm not sure on the implications of this tho.

node/state.re
```diff
@@ -129,26 +129,16 @@ let try_to_commit_state_hash = (~old_state, state, block, signatures) => {
 
 let apply_block = (state, block) => {
   let old_state = state;
-  let.ok (protocol, new_snapshot, results) =
-    apply_block(state.protocol, block);
+  let.ok (protocol, results) = apply_block(state.protocol, block);
```
EduardoRFS, line 132: We should probably check here to see if block.state_root_hash matches the exped block.state_root_hash

EduardoRFS (changes requested):

node/state.re
d4hines: We already introduced a check in #227 for this: https://github.com/marigold-dev/deku/pull/227/files#diff-43dae84d72a098cf13a754e41ae4fc754d43ee859e8be911f36ae612b542aaa5R194-R199

Or do you mean something else?

d4hines:
I've rebased this. I've also tested it and it seems to work (produces blocks, just like #227).

d4hines:
I've rebased this on top of the bug fix in #227

d4hines:
This PR has been rebased and tested.

node/state.re
```diff
@@ -237,31 +242,6 @@ let load_snapshot =
...
-          Option.value(~default=prev_state_root, next_state_root),
-        ));
-      },
-      (protocol, next_state_root),
-      all_blocks,
-    );
-  //TODO: snapshots?
-  Ok({...t, next_state_root, block_pool, protocol});
+  let t = {...t, next_state_root, protocol, block_pool};
+  List.fold_left_ok(apply_block, t, all_blocks);
```
EduardoRFS, line 248: very good move

EduardoRFS (approved):
all might saying good job

PR #233, opened 2021-10-13
move Tezos_interop crypto code to crypto library

EduardoRFS:
## Problem

Currently we have many pieces of the codebase doing cryptographic definitions, simple things like defining what is a Key, Key_hash, Secret and Signature are duplicated.

## Solution

Here I define a library that will be used in the future for abstracting anything `crypto` in the project in the future.


d4hines (approved):
LGTM

PR #234, opened 2021-10-13
Optimistic async state hashing

d4hines:
## Depends

- [ ] #232

## Problem

We want async state hashing per #147. 

This review is highly contingent on #232 - please review that one first. 

## Solution

#232 lifts state hashing from inside `Protocol` to inside the node. 

This PR is an incremental step towards full async hashing. In the best case, hashing is done fully asynchronously; however, if the hash is not finished, the node hashes the state synchronously on demand. 

To achieve this, the PR builds on #232 in two steps (1 commit per step):

1. First, it drops the notion of `next_state_root_hash`. Instead, we extend the node state with two fields:
  - `current_epoch` - an integer identifier for the current state root hash epoch.
   - `finished_hashes` - a map from epochs to the corresponding hash.
   - In the "setter" function, we build in a filter that drops any hashes older than the current epoch. This prevents the map from growing forever and also prevents the epoch from moving backwards.
  - This commit should not change the behavior in any way from #232 - reviewers, please help me check this!
2.  Next, we add an additional way to add hashes to the map, namely via async state hashing. We also slightly modify the `apply_block` function to be optimistic: if the map already contains the hash (because of async state hashing), then we use that value - otherwise,  we generate the hash synchronously. This allows us to have aysnc state hashing while never getting out of sync because of slow hashing. The cost of course is that some hashes may be done twice. 

In the next PR, we'll add more sophistication on the case where the hash is not already finished, adding a notion of in-sync and out-of-sync. 

## Related

- #147

### DIFF
--- a/node/state.re
+++ b/node/state.re
@@ -24,6 +24,7 @@ type t = {
   block_pool: Block_pool.t,
   protocol: Protocol.t,
   snapshots: Snapshots.t,
+  next_state_root_hash: BLAKE2B.t,
   // networking
   // TODO: move this to somewhere else but the string means the nonce needed
   // TODO: someone right now can spam the network to prevent uri changes
@@ -77,6 +78,7 @@ let make =
     block_pool: initial_block_pool,
     protocol: initial_protocol,
     snapshots: initial_snapshots,
+    next_state_root_hash: initial_block.state_root_hash,
     // networking
     uri_state: Uri_map.empty,
     validators_uri: initial_validators_uri,
@@ -151,6 +153,7 @@ let try_to_commit_state_hash = (~old_state, state, block, signatures) => {
     );
   });
 };
+
 let apply_block = (state, block) => {
   let old_state = state;
   let.ok (protocol, new_snapshot, results) =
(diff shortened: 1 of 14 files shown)


node/state.re
```diff
@@ -42,6 +44,27 @@ type t = {
...
+  finished_hashes: Int_map.add(epoch, (hash, data), state.finished_hashes),
+};
+
+let get_next_hash = state =>
+  Int_map.find_opt(state.current_epoch + 1, state.finished_hashes);
+
+/** Increments the [current_epoch]. Additionally,
+    removes any [finished_hashes] from previous epochs.
+    This ensures the state root hash only moves forward in time. */
+let increment_epoch = state => {
```
EduardoRFS, line 58: Better name would be `start_new_epoch`

EduardoRFS (changes requested):

d4hines:
This had a bug similar to the one I addressed in #227 but I've rebased and fixed.

d4hines:
This PR has been rebased and tested.

d4hines:
Some further testing leads me to believe that although this branch works, it's not working as intended. I think it's hitting the fallback case and synchronously hashing every time.

d4hines:
Superseded by #412

PR #239, opened 2021-10-14
Use cmdliner for node

### LINKED ISSUE #2: Use cmdliner for node http_server and genesis_helper

anchpop:
## Problem

The node does not use cmdliner. That means if you just run the node and pass no arguments you get a pretty unhelpful error message. 

```
$ esy x deku-node
Fatal error: exception (Invalid_argument "index out of bounds")
```

## Solution

Use cmdliner to process the command line arguments. Now the error is much more explanatory:

```
$ esy x deku-node
deku-node: required argument folder_node is missing
Usage: deku-node [OPTION]... folder_node
Try `deku-node --help' for more information.
```

I'm not sure that `folder_node` is the right name (I just copied it from sidecli), feel free to change it.

Also this helps with the logging PR. Cmdliner comes with built-in combinators for handling logging (e.g. changing the verbosity, enabling or disabling the colors, etc.) which we use in sidecli currently but can't use in deku-node without this PR or one like it. 

## Disclaimer

I feel like it probably works, but I don't know how to actually run the node so I'm not actually able to test it properly. I have a meeting tomorrow with daniel where I'll figure that and other things out.

## Related

Closes #2

### DIFF
--- a/bin/deku_node.re
+++ b/bin/deku_node.re
@@ -2,6 +2,7 @@
 // Can a core send a message and the other receive it in the past?
 // TODO: should start signing before being in sync?
 
+open Cmdliner;
 open Opium;
 open Helpers;
 open Protocol;
@@ -187,15 +188,11 @@ let handle_ticket_balance =
     },
   );
 
-let folder = Sys.argv[1];
-let trusted_validator_membership_change_file =
-  folder ++ "/trusted-validator-membership-change.json";
-
-let node = {
+let node = (folder, trusted_validator_membership_change_file) => {
   let.await identity = Files.Identity.read(~file=folder ++ "/identity.json");
   let.await trusted_validator_membership_change_list =
     Files.Trusted_validators_membership_change.read(
-      ~file=folder ++ "/trusted-validator-membership-change.json",
+      ~file=folder ++ trusted_validator_membership_change_file,
     );
   let trusted_validator_membership_change =
     Trusted_validators_membership_change.Set.of_list(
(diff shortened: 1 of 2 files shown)


bin/deku_node.re
```diff
@@ -285,28 +279,54 @@ let node = {
...
-     )
-  |> App.start
-  |> Lwt_main.run;
+  Term.(const(node) $ folder_node);
+};
 
-let (forever, _) = Lwt.wait();
-let () = Lwt_main.run(forever);
+let () = {
+  Random.self_init();
```
d4hines, line 329: This needs to be removed if #245 is merged first.

bin/deku_node.re
```diff
@@ -285,28 +279,54 @@ let node = {
...
+    |> handle_trusted_validators_membership(
+         ~persist=Files.Trusted_validators_membership_change.write,
+         ~file=trusted_validator_membership_change,
+       )
+    |> App.start
+    |> Lwt_main.run;
+
+  let (forever, _) = Lwt.wait();
+  let () = Lwt_main.run(forever);
+  ();
```
d4hines, line 316: Maybe just:
```suggestion
Lwt_main.run(forever)
```
?

bin/deku_node.re
```diff
@@ -285,28 +279,54 @@ let node = {
...
+};
+let node = {
+  let folder_node = {
+    let docv = "folder_node";
+    let doc = "The folder where the node lives.";
```
d4hines, line 321: We can do better than this. Maybe something like: 
```suggestion
    let doc = "Path to the folder containing node configuration data.";
```

d4hines (changes requested):
Looks good! Only a few minor changes suggested.

bin/deku_node.re
EduardoRFS: please rebase on #245

anchpop: @EduardoRFS rebased

EduardoRFS (approved):
Finally closing #2
small anime girl running to a better future

PR #241, opened 2021-10-14
PoC: Fix nodejs unsafe Buffer() warning

callistonianembrace:
## Problem

We keep running into `Buffer() is deprecated due to security and usability issues...` when running Taquito scripts

## Solution

This issue stems from `xhr2-cookies` that `@taquito/rpc` uses - they use `new Buffer()` which is deprecated due to a security issue.

This is PoC PR that demo's a fix: by vendoring xhr2-cookies, using `Buffer.from()` in the affected places and using esy's resolutions feature to use the vendored source, we get past this issue.

Ideal fix would be a properly audit xhr2-cookies, apply security recommendations (Buffer.from etc) and publish it under our namespace until upstream authors do something about it. Also, good issue to report at taquito


EduardoRFS:
@callistonianembrace can you report at taquito and look how much work it would be to use another library instead of xhr2-cookies? I don't like the idea of vendoring a JS library on our side.

PR #243, opened 2021-10-14
Async State Hashing

d4hines:
## Depends

- [ ] #234

## Problem

We want full async state hashing per #147.  This PR implements this.

Recall that per #218, this has the following requirements (reposted for convenience):

1. If an honest node hashes a particular state, every other node will either
  eventually hash that same state or has already hashed that state.
2. The block producer is forced to produce state root hash updates on time.
  - 2a. Validators reject blocks with state root hash changes that occur too
    early.
  - 2b. Validators reject blocks with a state root hash that is too old.
3. The state root hash can only advance forward - that is, when a validator sees
  a block with state root hash h, and then a subsequent block with state root
  hash h', the validator rejects any subsequent blocks with state root hash h.
4. Honest validators that are in sync and have sufficient hashing power always
  have the next state root hash before it is needed (i.e, hashing is no longer a
  bottleneck).
5. Honest validators that are behind in hashing mark themselves as out of sync
  and continue to follow but don't sign blocks.
6. Honest validators can get back in sync when finished hashing (assuming they
  were in sync before they fell behind).

## Solution

This PR removes the synchronous state hashing in `./node/state.re`, and instead relies fully on the async hashing
introduced in #234 .

This covers requirements 1-5. Requirement 6 is an optimization that we will address in the next PR. Here's the argument for 1-5:

1. Hashing now occurs in a single place: [after applying a block and when a new epoch is detected](https://github.com/marigold-dev/deku/pull/243/files#diff-43dae84d72a098cf13a754e41ae4fc754d43ee859e8be911f36ae612b542aaa5R240-R242). All honest nodes will eventually call `try_to_apply_block` on the same inputs, and thus will eventually try to hash the same states.
2. In #227 we introduced the check [block_has_signable_state_root_hash](https://github.com/marigold-dev/deku/pull/243/files#diff-56e44657a8ef45dd1d224a6694e7bc62162c155e1ad1d9f4bb7c72974738fcedR81-R102) (I've linked to the most recent version because it's been slightly refactored since then). Validators only sign blocks if the hash is the expected one and neither too early too late according to this function.
3. In #234 we introduced the notion of `next_state_root_block_height` - a field on the state that identifies the block height of the state that will be the next state root hash. Every new epoch, we update this value with the [start_new_epoch function](https://github.com/marigold-dev/deku/pull/243/files#diff-51206670a0c86886493d1334cd2ce87b212a91a2b872443ff420fe72a1373a1dR55-R67). Additionally, we introduced `finished_hashes` - a block heights to their corresponding state root hashes. Whenever a new epoch is started, any hashes corresponding to older epochs are deleted from this map. If a dishonest block producer tried to send a block with an old state root hash, validators would not be able to find it in `finished_hashes`, thus satisfying our property that the state root epoch cannot move backwards.
4.  In #227, we shifted the state root hash to be the one from the previous epoch. Starting in #234 and finishing in this PR, the generation of that hash is fully async. So long as hashing finishes before the next epoch, block producers will send new state root hashes on time, and validators will be able to sign them. Conversely, if they fail to produce them on time, the block producer will continue sending the current state root hash (these blocks may be rejected if the `max_signable_time_between_epochs` interval passes). If validators fail to finish the hash in time, they will not be able to sign the block. In both cases, they must wait until the next epoch and try again to finish hashing it in time (this violates requirement 6, which we'll discuss further below).
5. As mentioned, the new `[block_has_signable_state_root_hash]` added to the `is_signable` ensures validators won't sign blocks unless they can verify it has the correct hash (implying they finished that hash already). However, it is not required to know the state root hash to apply a block - you can see [we've removed the check for a valid hash from `Protocol.apply`](https://github.com/marigold-dev/deku/pull/243/files#diff-c4ae22b7bdcedfacfae88c0f46d64b678eeec3b92b72875d4810402744807affL181-L189). Thus, nodes will continue following the chain by applying blocks even if they can't sign them.

Regarding requirement 6, this is an important optimization to allow nodes to quickly get back in sync if they fall slightly behind, without waiting for a whole epoch to pass. The idea is to keep track of which blocks could have been signed but were missing a hash, and whenever a hash finishes, check to see if any have become signable. 

If 1-5 are correct, the chain should work without this optimization. We'll address 6 in the next PR. 

Reviewers, please check my work on points 1-4! This design is slightly different than what's outlined in #218. Also, the correctness of this PR depends on the correctness of its dependencies, so we should iron those out first.

## Related

- #147

### DIFF
--- a/node/flows.re
+++ b/node/flows.re
@@ -184,7 +184,7 @@ let try_to_sign_block = (state, update_state, block) =>
   };
 
 /** Starts asynchronously hashing a new state. Side-effectfully updates
-    the map of epoches to hashes when done.
+    the state with the results.
 
     Each block is associated with a particular state root hash.
     The state root starts with genesis, and then updates periodically.
@@ -195,21 +195,29 @@ let try_to_sign_block = (state, update_state, block) =>
     on a separate thread. This allows the block producer to continue
     producing blocks even while the state root is being updated.
  */
-let hash_new_state_root = (state, update_state) => {
+let hash_new_state_root = (old_state, update_state) => {
   Lwt.async(() => {
     // Lwt_domain.detach causes the function to run in a separate thread.
     // When the promise resolves, the rest of the function is run in main thread.
-    let.await next_state_root =
-      Lwt_domain.detach((s: Node.t) => Protocol.hash(s.protocol), state);
-    let _ =
-      update_state(
-        // The state may be stale so we must retrieve the current state again.
-        State.add_finished_hash(
-          state.protocol.block_height,
-          next_state_root,
-          get_state^(),
-        ),
+    let.await new_snapshot =
+      Lwt_domain.detach(
+        (s: Node.t) => Protocol.hash(s.protocol),
+        old_state,
       );
+    let new_state =
+      State.add_finished_hash(
+        old_state.protocol.block_height,
+        new_snapshot,
+        // The state may be stale so we must retrieve the current state again.
+        get_state^(),
+      );
+    let snapshots =
+      Snapshots.update(
+        ~new_snapshot,
+        ~applied_block_height=old_state.protocol.block_height,
+        new_state.snapshots,
+      );
+    let _ = update_state({...new_state, snapshots});
     Lwt.return();
   });
 };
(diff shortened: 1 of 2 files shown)


d4hines:
Superseded by the line of PR's starting with #412

PR #245, opened 2021-10-14
initialize rng inside of crypto

EduardoRFS:
## Problem

mirage-crypto-rng which is used by mirage-crypto-ec needs to be initialized with a random generator before any other crypto call happens, this leads to a problem where we initialize the RNG in many places.

## Solution

Because the crypto module needs to be initialized before any crypto function can be used it creates a nice spot to do RNG initialization, and this is what I do on this PR.

As a bonus I remove all the external mentions to mirage-crypto.

### DIFF
--- a/bin/deku_node.re
+++ b/bin/deku_node.re
@@ -2,9 +2,6 @@
 // Can a core send a message and the other receive it in the past?
 // TODO: should start signing before being in sync?
 
-Random.self_init();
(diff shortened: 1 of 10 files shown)


EduardoRFS:
@callistonianembrace hmm there is any case where the RNG shouldn't be initialized? We need the RNG to even load the application.

Issue #198, opened 2021-09-23
support tz3 on tezos_interop


EduardoRFS:
Subset of #22, tz3 uses P256, which should be currently supported by mirage-crypto.

PR #247, opened 2021-10-15
add a dependency to hacl-star

pkhry:
## Problem

We need to add support for tz3 and p256 module form mirage crypto doesn't support compression of public_keys, so we need hacl-star
## Solution

add a dependency on hacl-star, also it would require installation of libffi6 to build the sidechain
## Related
- #198


package.json
```diff
@@ -70,4 +72,4 @@
     "@opam/opium": "github:EduardoRFS/opium:opium.opam#9fdbfbd0eedf238e1be0663ab95977be27d9a6e1",
     "@opam/rock": "github:EduardoRFS/opium:rock.opam#9fdbfbd0eedf238e1be0663ab95977be27d9a6e1"
   }
-}
+}
```
d4hines, line 75: Can you fix this missing new line?
```suggestion
}

```

d4hines (approved):
LGTM

EduardoRFS:
As discussed in private, now that the code for P256 compression was written why not upstream it to mirage-crypto-ec?

PR #250, opened 2021-10-15
Add mechanism to sign blocks with missed state root hashes

d4hines:
## Depends

- [ ] #243

## Problem

In #243, we achieved fully async hashing; however, as a consequence, if a node didn't finish hashing a state before receiving a block with that state root epoch, he was unable to sign the block at all. 

Per requirement 6 in #218, we want a way for nodes to "recover" from late state root hashing without having to wait a full epoch. This PR provides that mechanism.

## Solution

Extend the node state with a map `blocks_with_unknown_hash` from state root hashes to blocks. Any time a block
would be signable except that the state root hash isn't known, add it to this map. Whenever a state root hash finishes,
check this map for blocks under that hash, and try signing them.

### Problems with this solution
- I believe that when a hash is finished and we sign missed blocks, we should try applying those blocks if they have sufficient signatures, but I don't think my implementation does this.
- ~I realized as I'm writing this PR description that the map only grows - we need a way of filtering the map of old blocks, otherwise it would be easy to DOS a node by sending lots of blocks with unknown hashes.~ Fixed in latest version: we filter out any blocks from the map that have a block height lower than the current block height https://github.com/marigold-dev/deku/pull/250/files#diff-43dae84d72a098cf13a754e41ae4fc754d43ee859e8be911f36ae612b542aaa5R247-R263

## Related

- #147

### DIFF
--- a/node/building_blocks.re
+++ b/node/building_blocks.re
@@ -100,6 +100,17 @@ let block_has_signable_state_root_hash = (~current_time, state, block) => {
   };
 };
 
+/** The possibile return values of [is_signable]. */
+type signability =
+/** The block would be signable, except the state root hash is unknown.
+    (perhaps we'll finish it soon). */
+| Signable_with_state_root_hash(BLAKE2B.t)
+/** The block is signable as is. */
+| Signable
+/** The block is not signable for some reason other than a missing state
+    root hash. */
+| Not_signable
+
 // TODO: bad naming
 // TODO: check if block must have published a new snapshot
 let is_signable = (state, block) => {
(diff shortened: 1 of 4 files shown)


node/building_blocks.re
```diff
@@ -55,6 +55,62 @@ let is_current_producer = (state, ~key) => {
...
+    switch (next_state_root) {
+    | Some((next_state_root_hash, _)) =>
+      block.state_root_hash == next_state_root_hash
+      && time_since_last_epoch >= minimum_signable_time_between_epochs
+    | None => false
+    };
+  };
+};
+
+/** The possibile return values of [is_signable]. */
```
callistonianembrace, line 103:
```suggestion
/** The possible return values of [is_signable]. */
```
typo

EduardoRFS:
@d4hines can we close this one?

PR #251, opened 2021-10-15
use crypto directly inside deku

Zett98:
## Problem

we use Tezos_interop address types for plain ed25519 types.
Edit: 
Address.t and wallet.t are represented as ed25519 internally, but use tezos_interop module, which only had ed25519 previously, now that we will have support for tz2 and tz3 it will introduce redundant pattern matching into those modules and modules that call functions defined there.
## Solution

use ed25519 directly
## Related
- #22

### DIFF
--- a/bin/deku_node.re
+++ b/bin/deku_node.re
@@ -223,7 +223,7 @@ let node = {
       |> List.mapi((i, validator) => {
            (
              switch (validator) {
-             | Tezos_interop.Key.Ed25519(k) => k
+             | k => k
              },
              Printf.sprintf("http://localhost:444%d", i) |> Uri.of_string,
            )
--- a/bin/sidecli.re
+++ b/bin/sidecli.re
@@ -88,10 +88,9 @@ let uri = {
 };
 
 let address = {
-  open Tezos_interop.Key_hash;
   let parser = string =>
-    Tezos_interop.Key_hash.of_string(string)
-    |> Option.map((Ed25519(hash)) => Wallet.address_of_blake(hash))
+    Crypto.Ed25519.Key_hash.of_string(string)
+    |> Option.map(hash => Wallet.address_of_blake(hash))
     |> Option.to_result(~none=`Msg("Expected a wallet address."));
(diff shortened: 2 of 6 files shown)


Zett98:
> This seems like a regression to me, why would we use Ed25519 directly? Please write a better description for the problem.

Updated

bin/sidecli.re
```diff
@@ -88,18 +88,17 @@ let uri = {
 };
 
 let address = {
-  open Tezos_interop.Key_hash;
   let parser = string =>
-    Tezos_interop.Key_hash.of_string(string)
-    |> Option.map((Ed25519(hash)) => Wallet.address_of_blake(hash))
+    Crypto.Ed25519.Key_hash.of_string(string)
+    |> Option.map(hash => Wallet.address_of_blake(hash))
```
EduardoRFS, line 93: This is still a regression, address is not supposed to accept only ed25519. It was like that because ed25519 was the only supported platform.

bin/deku_node.re
```diff
@@ -223,7 +223,7 @@ let node = {
       |> List.mapi((i, validator) => {
            (
              switch (validator) {
-             | Tezos_interop.Key.Ed25519(k) => k
+             | k => k
              },
```
callistonianembrace, line 227: This switch statement seems like a noop. I guess the tuple could simply be `(validator, sprintf("...", i))`

protocol/address.re
```diff
@@ -7,8 +7,8 @@ let key_to_yojson = key =>
   `String(Tezos_interop.Secret.to_string(Ed25519(key)));
 let key_of_yojson = json => {
   let.ok string = [%of_yojson: string](json);
-  let.ok Ed25519(key) =
-    Tezos_interop.Secret.of_string(string)
+  let.ok key =
+    Ed25519.Secret.of_string(string)
```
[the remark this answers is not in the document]
EduardoRFS: Overall I don't see the point of this PR it does a change that will need to be undone in the future.

PR #254, opened 2021-10-19
another small refactor on crypto library

EduardoRFS:
## Problem

This is a collection of 4 small problems.

1. Crypto.S is exposed but not used
1. Crypto_intf is a module but only contains types
1. Ed25519 doesn't contain an interface file
1. crypto.rei is actually defined on Crypto_intf.Intf

## Solution

1. stops exposing Crypto.S, it can be exposed again in the future if it becomes needed
1. Make Crypto_intf a module without implementation, useful for build speed
1. adds a simple interface for Ed25519, also useful for build speed
1. move Crypto_intf.Intf to crypto.rei

### DIFF
--- a/crypto/crypto.re
+++ b/crypto/crypto.re
@@ -1,6 +1,6 @@
 Random.self_init();
 Mirage_crypto_rng_unix.initialize();
 
-module Base58 = Base58;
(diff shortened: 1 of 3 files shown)


callistonianembrace (approved):
Reviewed out of curiosity. LGTM

d4hines (approved):
LGTM

PR #267, opened 2021-10-20
move shared tezos code to separate library

Zett98:
## Problem

Adding tz2 and tz3 support to tezos_interop creates a need to explicitly handle all the cases that propagate to protocol code. 
## Solution

Move shared Tezos code to a separate folder to use it inside interop and protocol. It will make the addition of tz2 and tz3 to both the interop and protocol way more straightforward.
## Related

- #22 
- #94


EduardoRFS:
@Zett98 can you rebase or close this?

PR #268, opened 2021-10-20
Depend on tezos inside protocol

Zett98:
## Depends

- [ ] #267

## Problem

We currently use Ed25519 addresses everywhere, but if we intend to support tz2 and tz3 we should move to using Tezos addresses instead.
## Solution

Use Tezos addresses inside protocol, this will streamline the addition of tz2 and tz3 addresses in the future. 
## Related

- #22
- #94


bin/deku_node.re
```diff
@@ -219,9 +219,7 @@ let node = {
       current_validators
       |> List.mapi((i, validator) => {
            (
-             switch (validator) {
```
callistonianembrace, line 222: Nit: this change, and another one similar to this later on, looks unrelated to the PR's objectives.

rosalogia (approved):

EduardoRFS:
@Zett98 I believe this PR is outdated right?

PR #272, opened 2021-10-21
Automated Setup Scripts

d4hines:
## Depends
- [ ] #280 

## Problem

Running a Deku cluster is difficult. There are many steps:
1. Run a Tezos node
2. Issue a new wallet from https://faucet.tzalpha.net
3. Activate that wallet
4. Make contract on Ligo website
5. Configure nodes
6. Start tmux script

Running a Tezos node itself has several substeps addressed in https://hackmd.io/Er_EPZIrQSardvo4bpuYJg?view

I'd like to automate this entire process so that a Deku cluster can be started with a single command, and so that we can test this command regularly in CI.

## Solution

@fraidev has done good work to solve 1 in the above HackMD - the steps he outlines there can be automated by as script, which I've done in `./node.sh`. 

To solve 2, I've created https://github.com/marigold-dev/faucet-bot, which will give tez to anyone that asks. We can use this to set up wallets in an automated fashion, which removes the need for 3.

For 4, Ligo has a Docker image you can use for portable binaries. Once we can automate getting a wallet with funds, we can easily deploy a contract automatically as well.

~5 is already partially automated by `setup.sh`. But @Zett98 showed me some minor fixes that you currently have to do manually. These could easily be automated as well.~ I've fully automated 5 now.

In my testing, this script sort of works with `./start.sh`; however, I have to run the `./start.sh` twice, and if I stop the node, I have to do a full `./setup.sh clean` for things to work again (that means there's some bug somewhere, but I'm not sure where).

This script is a bit hacky, but I think it's a big improvement on current process.

### Future Directions

It might make sense to have a plugin architecture for the sidechain CLI, and maintain a Tezos plugin that can do all this in principled way. This was not easy to figure out. Such a plugin might depend directly on Tezos, greatly simplifying this whole process. Or maybe we punt and make the plugin depend on docker (or Nix).


setup.sh
```diff
@@ -1,31 +1,93 @@
 #! /bin/sh
```
anchpop, line 1: I suggested this in the past but for reasons I don't remember it was never changed. This allows me to run the script without directly mentioning `bash` (it uses features not available in standard `sh`)

```suggestion
#! /bin/bash
```

anchpop:
When running the script, I get 

```
=========== Adding funds to myWallet tz1RBxwDcANvbDWsUW6beo7Jtq6AFhZCyKQG ===========
curl: (28) Failed to connect to faucet-bot.marigold.dev port 3000: Connection timed out
```

setup.sh
```diff
@@ -49,51 +111,130 @@ cat <<EOF
...
-  printf "  }"
-done
+
+  message "Starting Node "
+  $node_manager stop
+  $node_manager node start --rpc-port 8732 --allow-all-rpc :8732
+
+  message "Waiting for Bootstrap "
+  sleep 5
+  $node_manager client bootstrapped >/dev/null
```
anchpop, line 170: I wonder if people may find it useful to see the output so they can see the node is bootstrapping and not just hanging?
```suggestion
  $node_manager client bootstrapped
```

anchpop:
Aside from `faucet-bot.marigold.dev` not working this script is perfect. Thanks so much!

d4hines:
Closing in favor of Flextesa-based solution

PR #280, opened 2021-10-29
Lower required confirmations for Tezos transactions

d4hines:
## Depends

NA

## Problem

Currently the sidecli configures a Deku node to require 10 confirmations to every Tezos transaction it posts. This takes a very long time, especially on granadanet. The result is that the chain appears to just halt when running `./start.sh`.

## Solution

I propose lowering it to 1. Even at 2 the chain is very slow, appearing broken. In future PR's we  can accept a CLI option or environment variable during setup, but I believe in these early stages the default should lean towards the fastest possible setup for devs and early testers.

## Related

NA

### DIFF
--- a/bin/sidecli.re
+++ b/bin/sidecli.re
@@ -570,7 +570,7 @@ let setup_tezos = (node_folder, rpc_node, secret, consensus_contract) => {
       rpc_node,
       secret,
       consensus_contract,
-      required_confirmations: 10,
+      required_confirmations: 1,
     };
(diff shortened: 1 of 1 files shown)


EduardoRFS:
That's not a sane default, yes it's okay for development, but you can change locally on your machine easily.

d4hines:
I can make a PR to make required confirmations configurable; however, we need to have instructions for setting up the node easily. I'll say something like:

> **In production you should set a higher level of required confirmations, else your application may be unstable**

Does that work?

PR #282, opened 2021-11-02
fix random generator error

Zett98:
## Problem

Running Crypto.Ed25519.generate() fails due to uninitialized random generator
## Solution

pass the generator instance explicitly


EduardoRFS:
As this only changes type, I will be merging by myself, added a description on the workaround so that we remember what it does in the future.

EduardoRFS (approved):
anime girl stressed with Edu merging a PR that he wrote himself

PR #283, opened 2021-11-03
fix wallet address mixup

EduardoRFS:
## Problem

For a long time, the Address and Wallet module in the protocol are actually the opposite, `Wallet.t` describes a `key_hash`  and `Address.t` a `key`, this is clearly wrong.

On #208 the bulk of the diff is related to naming, so to simplify that PR we need to solve this problem.

## Solution

Here I switch the modules by name and fix it on the entire codebase, also fix the naming in a couple places, not all of them were fixed because on #208 some if it will actually not change.


anchpop (approved):

protocol/validators.re
```diff
@@ -3,7 +3,7 @@ open Crypto;
 
 // TODO: we should avoid dead validators to avoid double timeout, A(dead) -> B(dead) -> C
 [@deriving (yojson, ord)]
-type validator = {address: Address.t};
+type validator = {address: Wallet.t};
```
rosalogia, line 6: I don't understand this change. Why call the field `address` if representing it by a Wallet? Seems like we now represent keys as wallets, and key_hashes as addresses, so isn't this inconsistent?

EduardoRFS: Because on your PR this will change, from `Wallet.t` to `Adderss.t` back, so the field was kept to avoid noise on the review.

As soon as you apply your patch over this, it will look as `{ address: Address.t }`

rosalogia: Awesome, thanks.

PR #284, opened 2021-11-05
fix tmuxinator script

### LINKED ISSUE #126: fix tmuxinator script
After `inject-genesis` and `make-credentials` are removed the `tmuxinator` script will be broken, and I don't know how it's supposed to work I'm opening this issue as a follow up.

@anchpop

anchpop:
## Problem

The tmuxinator script exists to make running testing node more convenient. Devs can run `tmuxinator` to spin up 3 nodes and inject a block. However, the script has been broken for a long time. Most importantly, the sidecli command `inject-genesis` was removed.

## Solution

Use `produce-block` rather than `inject-genesis`. Also, I added `rm $(find ./ -name state.bin)` to ensure the node always runs from a fresh state. 

## Related

Closes #126

### DIFF
--- a/.tmuxinator.yml
+++ b/.tmuxinator.yml
@@ -11,7 +11,7 @@ root: ./data
 
 # Project hooks
 # Runs on project start, always
-on_project_start: esy build
+on_project_start: esy && rm $(find ./ -name state.bin) # Make sure the project is built, and delete any `state.bin`s to ensure we're running from a fresh state
 # Run on project start, the first time
 # on_project_first_start: command
 # Run on project start, after the first time
@@ -22,7 +22,7 @@ on_project_start: esy build
 # on_project_stop: command
 
 # Runs in each window and pane before window/pane specific commands. Useful for setting up interpreter versions.
-pre_window: esy shell
+# pre_window: command
 
 # Pass command line options to tmux. Useful for specifying a different tmux.conf.
 # tmux_options: -f ~/.tmux.mac.conf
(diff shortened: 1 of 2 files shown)


.tmuxinator.yml
```diff
@@ -45,13 +45,11 @@ pre_window: esy shell
...
+      # Due to an esy bug, multiple `esy x`s can't run at the same time, so we use sleep to space them out
+      # Also, we need to add a delay to the first one to give the `on_project_start` command time to run
       panes:
-        - sleep 1 && $cur__target_dir/default/bin/deku_node.exe 0
-        - sleep 1 && $cur__target_dir/default/bin/deku_node.exe 1
-        - sleep 1 && $cur__target_dir/default/bin/deku_node.exe 2
-        - sleep 1 && $cur__target_dir/default/bin/deku_node.exe 3
-        - $cur__target_dir/default/bin/sidecli.exe make-credentials && sleep 3 && $cur__target_dir/default/bin/sidecli.exe inject-genesis
+        - sleep 1 && esy x deku-node 0
+        - sleep 3 && esy x deku-node 1
```
d4hines, line 52:
```suggestion
        - sleep 2 && esy x deku-node 1
```

d4hines: In my testing a 1 second delay between nodes was enough. @anchpop if you can confirm this, then please accept all the suggestions. Otherwise, reject all.

.tmuxinator.yml
```diff
@@ -45,13 +45,11 @@ pre_window: esy shell
...
+        - sleep 5 && esy x deku-node 2
```
d4hines, line 53:
```suggestion
        - sleep 3 && esy x deku-node 2
```

.tmuxinator.yml
```diff
@@ -45,13 +45,11 @@ pre_window: esy shell
...
+        - sleep 7 && esy x sidecli produce-block 0
```
d4hines, line 54:
```suggestion
        - sleep 4 && esy x sidecli produce-block 0
```

d4hines (approved):
I've tested this a lot in #288. Works :+1: 

Only change we might want to make is reducing the delay between nodes to 1 second instead of 2.

anchpop:
@d4hines I had issues with 1 second :( although actually maybe invoking the node via `$cur__target_dir/default/bin/deku_node.exe` wouldn't require the delay?

EduardoRFS (changes requested):
This script is not supposed to work and it will break with #227 look on start.sh

anchpop:
Closing this PR as I no longer feel like the tmux script is necessary or useful

PR #285, opened 2021-11-06
Add logging MVP

anchpop:
## Depends

#239

## Problem

We have no unified way of adding logs. Logs are often useful when debugging, but clutter up the console output otherwise, and it's annoying to have to remove and re-add them every time. 

## Solution

This uses the `Logs` library to add logging. The level of detail can be adjusted by passing `-v`, `-vv`, or `-vvv` to `esy x sidecli` or `esy x deku-node`. 

An upcoming PR will enhance logging to make the logs more informative. 

## Related

#113, #64

PS I was going to just reopen #113 but Github wouldn't let be because I force-pushed the branch while it was closed


oteku:
You should use standard levels instead of `v` accumulators
https://erratique.ch/software/logs/doc/Logs/index.html#levels

anchpop:
> https://erratique.ch/software/logs/doc/Logs/index.html#levels

So you would have us pass `-Warning` instead of `-vv`?

d4hines:
@anchpop we either have to rebase or close this PR.

sandbox.sh
```diff
@@ -180,7 +180,7 @@ start_deku_cluster() {
   sleep 1
 
   echo "Producing a block"
-  HASH=$(deku-cli produce-block "$data_directory/0" | awk '{ print $2 }')
+  HASH=$(deku-cli produce-block "$data_directory/0" | tail -n 1 | cut -d ' ' -f 2)
```
aguillon, line 183: I suggest to use `sed -n 's/block.hash: \([a-f0-9]*\)/\1/p')` instead, more explicit and less fragile IMO.

oteku:
Logging are important for observability of the application. Format must be **consistent** and **timestamped**.
Consider adding a `Log` submodule in `Helpers` module with this interface:
```
module type Logs = sig
  val info: string -> unit
  val debug: string -> unit
  val warn: string -> unit
  val error: string -> unit
end
```
that will use erratique Logs module with predefined formatters

src/node/flows.ml
```diff
@@ -41,7 +41,7 @@ let string_of_error = function
   | `Snapshots_with_invalid_hash -> "Snapshots_with_invalid_hash"
 
 let print_error err =
-  Format.eprintf "\027[31mError: %s\027[m\n%!" (string_of_error err)
+  Logs.err (fun m -> m "\027[31mError: %s\027[m\n" (string_of_error err))
```
aguillon, line 44: Here and for other calls of Logs functions: remove `\n` from the chain, as it is automatically inserted by the library.

aguillon (changes requested):
Thank you for your PR, sorry for not reviewing it earlier.

I have several problems with your proposition:
- as suggested by Thomas, it would be easier to centralize the functions and make style more uniform by defining helpers for the 5 logging levels. Here is an example: https://github.com/aantron/dream/blob/2386083170766255021171870b8568d1c8c933b5/src/server/log.ml#L348 (although you don't have to use a record for this). Formatting should be decided in these functions, not on the caller side.
- `App` level should probably be `Info`, as Deku may be considered a library for Parametric Deku and libraries should not use `App`.
- Right now your output gets formatted by Dream (not sure why, since I'm not seeing you call Dream.log!) and it probably shouldn't be. In particular, the `REQ NN` column is useless and I think we should not have it.
- ~~I think logging adds a small overhead~~, which you could mitigate by using functions from Logs_lwt inside of your helpers (together with `async`). Edit: overhead is a bad argument, but still. Edit edit : definitely a bad idea: https://github.com/marigold-dev/deku/pull/113#issuecomment-948567230

We should also log more things, but this can come as a separate PR/discussion.

ulrikstrid:
Related infra work: https://github.com/marigold-dev/K8S-manifests/pull/4

oteku:
> @oteku
> 
> > Logging are important for observability of the application. Format must be **consistent** and **timestamped**. Consider adding a `Log` submodule in `Helpers` module with this interface:
> > ```
> > module type Logs = sig
> >   val info: string -> unit
> >   val debug: string -> unit
> >   val warn: string -> unit
> >   val error: string -> unit
> > end
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > that will use erratique Logs module with predefined formatters
> 
> I added a module for this, because it's nice to have it in `Helpers` anyway, but I think timestamping might be better in the reporter

Yes that's the good way to do with Logs, may comment about timestamping was generic, not considering the lib used.

esy.json
```diff
@@ -108,7 +108,7 @@
       }
     },
     "@opam/piaf": {
-      "source": "anmonteiro/piaf:piaf.opam#81cf20519cdebeecc1e2cca6d45d5a23acf644f5",
+      "source": "anchpop/piaf:piaf.opam#770d5557bdea7f7e34bb8bea702e0b7adafbb9d4",
```
aguillon, line 111: Ok!
https://github.com/anchpop/piaf/commit/770d5557bdea7f7e34bb8bea702e0b7adafbb9d4

sandbox.sh
```diff
@@ -275,9 +275,9 @@ start_deku_cluster() {
   # See deku-cli produce-block --help
   echo "Producing a block"
   if [ "$mode" = "docker" ]; then
-    HASH=$(docker exec -t deku-node-0 /app/deku_cli.exe produce-block /app/data | awk '{ print $2 }' | tail -n1 | tr -d " \t\n\r")
+    HASH=$(docker exec -t deku-node- /app/deku_cli.exe produce-block /app/data 0 | sed -n 's/block.hash: \([a-f0-9]*\)/\1/p' | tail -n1 | tr -d " \t\n\r")
```
aguillon, line 278: `tail -n 1` should not be necessary anymore

sandbox.sh
```diff
@@ -275,9 +275,9 @@ start_deku_cluster() {
...
   else
-    HASH=$(deku-cli produce-block "$DATA_DIRECTORY/0" | awk '{ print $2 }')
+    HASH=$(deku-cli produce-block "$DATA_DIRECTORY/0" | sed -n 's/block.hash: \([a-f0-9]*\)/\1/p')
```
aguillon, line 280: No `tr` here?

src/helpers/cmdliner_ext.ml
```diff
@@ -0,0 +1,48 @@
...
+let setup_log =
+  let open Term in
+  let log_label =
+    let docv = "Log label" in
+    let doc = "This will be prepended to all logs." in
+    Arg.(value & opt (some string) None & info ["log-label"] ~docv ~doc) in
+  let json_logs =
+    let docv = "Json logs" in
+    let doc = "This determines whether logs will be printed in json format." in
+    Arg.(value & flag & info ~doc ~docv ["json-logs"]) in
```
aguillon, line 35: @ulrikstrid is this enough for JSON logging, or do you need an environment variable too?

src/bin/log/log.ml
```diff
@@ -0,0 +1,2 @@
+let src = Logs.Src.create "deku.cli"
+include (val Logs.src_log src : Logs.LOG)
```
aguillon, line 2: This is not what we want IMO. We want a few functions (that you have to define yourself) matching the various levels of logging, for `app` (should be used by deku_node and people using deku as a library only), `info` (should be used by us mostly, when we produce blocks or update the hash), `debug` and `error`. These functions should be used as `Logs.info "New protocol snapshot hash: %s" (BLAKE2B.to_string hash)` directly in Deku.

I'll let @oteku @d4hines decide if we want colorized output or not.

Also, move this to helpers please.

src/node/log/log.ml
```diff
@@ -0,0 +1,2 @@
+let src = Logs.Src.create "deku.node"
+include (val Logs.src_log src : Logs.LOG)
```
aguillon, line 2: See comment on /bin/log/log.ml.

tests/test_tezos_rpc.ml
```diff
@@ -30,7 +30,7 @@ let fetch_block_operations ~block_hash () =
   | Ok operations ->
     List.iter
       (fun operation ->
-        Format.printf "operation.hash: %s\n%!"
+        Format.printf "operation.hash: %s"
```
aguillon, line 33: Why not use logging everywhere in the tests actually?

src/helpers/cmdliner_ext.ml
ulrikstrid: We can pass flags when running in kubernetes so this is perfect :ok_hand:

esy.json
ulrikstrid: https://github.com/anmonteiro/piaf/pull/127/files

But we should just remove the `s` here instead and it will get merged

tests/e2e/asserter.ml
```diff
@@ -7,13 +7,15 @@ let assert_state_correct data_folder expected_state_root_hash
   print_endline @@ "Checking state at path: " ^ data_folder;
   let file = data_folder ^ "/state.bin" in
   let protocol_state = Lwt_main.run @@ Files.State_bin.read ~file in
-  Format.printf "Minimum expected block height: %Ld. Actual height: %Ld\n%!"
-    expected_block_height protocol_state.block_height;
+  Logs.app (fun e ->
```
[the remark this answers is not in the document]
aguillon: Never-never? Shouldn't deku parametric use app (but not us, I agree)?

src/bin/log/log.ml
aguillon: I should have been more clear about not using include. Also, I guess I was wrong about `app` being used in the deku_node module.

src/node/flows.ml
```diff
@@ -132,7 +131,7 @@ let rec request_protocol_snapshot tries =
 let () =
   Lwt.async_exception_hook :=
     fun exn ->
-      Printexc.to_string exn |> Format.eprintf "global_exception: %s\n%!";
+      Log.error (Printexc.to_string exn |> Printf.sprintf "global_exception: %s");
```
ulrikstrid, line 134: Another issue with this API is that we always assume strings, here we pay for conversion `exn -> string` and 2x string interpolation.
To make it slightly better we could do something like
```ocaml
Log.error (Printf.sprintf "global_exception: %a" Fmt.exn exn);
```
But it's still as expensive just slightly nicer, and we pay for most of it even if we didn't want to log errors for some reason.

tests/e2e/asserter.ml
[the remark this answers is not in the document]
ulrikstrid: `app` is treated differently than other logs, I think it's fine to use it in the CLI and here in tests. Other than that  even parametric deku should use it sparingly

src/helpers/log.ml
```diff
@@ -0,0 +1,5 @@
+let src = Logs.Src.create "deku"
+module Log = (val Logs.src_log src : Logs.LOG)
+let error s = Log.err (fun e -> e "%s" s)
```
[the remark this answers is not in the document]
aguillon: This is not a problem for now, we'll fix it later with either a ppx or by using `[@@inline]`.

aguillon (approved):
Looks good to me, thanks!

nix/overlay.nix
```diff
@@ -23,6 +23,16 @@ in {
         propagatedBuildInputs =
           prev.lib.lists.remove osuper.uuidm o.propagatedBuildInputs;
       });
+      piaf = osuper.piaf.overrideAttrs (o: {
+        src = prev.fetchFromGitHub {
+          owner = "anmonteiro";
```
ulrikstrid, line 28: If we raise this PR in the overlays we can merge it directly

PR #286, opened 2021-11-09
Fix trusted-validator-membership-changes.json path

d4hines:
## Problem

The `deku-node` binary is broken. It double-prepends the `--folder` path when looking for `trusted-validator-membership-changes.json` and then fails when it can't find it.

## Solution

Fix the path. I did this by moving the place where we specify the path next to all the other places so it's more obvious what the path should be, preventing this mistake from happening again in the future.

## Related
NA

### DIFF
--- a/bin/deku_node.re
+++ b/bin/deku_node.re
@@ -188,11 +188,15 @@ let handle_ticket_balance =
     },
   );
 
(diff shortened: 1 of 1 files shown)


rosalogia (approved):

EduardoRFS (approved):
anime girl happy with deku node fixed

Issue #213, opened 2021-10-04
./setup.sh produces invalid configuration


Zett98:
As in title.

EduardoRFS:
Which invalid configuration?

EduardoRFS:
@Zett98 was the problem addressed or it was just a mistake of you?

anchpop:
@Zett98 what should be the contents of `trusted-validator-membership-change.json`?

d4hines:
No longer relevant since this code is long gone.

PR #287, opened 2021-11-10
Fix setup script file creation

d4hines:
## Problem
The setup script is currently incomplete/broken.

First, per #213, it does not generate the tursted-membership-validator-membership-change.json, without which the chain will not start.

Second, when we merged #283, we changed the type of an address from a public key to a public key hash (see [this line in the diff](https://github.com/marigold-dev/deku/commit/83abc7b1a1f4d6bad8d3ccd6cc4819682a0be553#diff-7c1b39fcf3624f72144844bfa6111a41df1b4c94844ad0e70456e0bf36f39ca9R5)). However, we did not change the setup script accordingly, and it still generates a public key in validators.json.

## Solution

Generating the trusted-membership-validator-change.json is very simple - I just copied the code for validators.json and made a slight modification.

Since `sidecli self` returns both the public key and the key hash, we can simply modify the setup script to use the key hash in the right place.

## Related
- #213
- #283

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -16,8 +16,9 @@ for i in ${VALIDATORS[@]}; do
 
   sidecli setup-identity $FOLDER --uri "http://localhost:444$i"
   KEY=$(sidecli self $FOLDER | grep "key:" | awk '{ print $2 }')
+  ADDRESS=$(sidecli self $FOLDER | grep "key:" | awk '{ print $2 }')
   URI=$(sidecli self $FOLDER | grep "uri:" | awk '{ print $2 }')
-  VALIDATORS[$i]="$i;$KEY;$URI"
+  VALIDATORS[$i]="$i;$KEY;$URI;$ADDRESS"
 done
 
 # To register the validators, run consensus.mligo with the list of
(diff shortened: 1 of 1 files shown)


anchpop (approved):
I really don't understand this script well enough to review it properly (man I wish it was python or js), but I understand the change and it seems reasonable

setup.sh
```diff
@@ -74,11 +75,32 @@ done
   echo "]"
 }
 
+trusted_validator_membership_change_json() {
```
callistonianembrace, line 78: This is unnecessary, @d4hines Initially, the file needs to be an empty array json file.

callistonianembrace (commented):
setup.sh points to https://ide.ligolang.org/p/-x6CdYJ5tIEaVzD9lGYsaA in the instructions, but that link contains an outdated contract. I have created a new link here: https://ide.ligolang.org/p/-U2zMPhmFo3e7sZzluIEpA Feel free to use it.

EduardoRFS:
@d4hines I believe this can be closed if #288 is the one to be merged.

PR #288, opened 2021-11-10
Add a sandbox script

d4hines:
## Depends
- [x] #290 
  
## Problem

There are several problems with running on public infrastructure for the "inner loop" of local development:
- It's quite complicated to set up. You have to 
  - Run a Tezos node, which itself has many substeps, addressed in @fraidev 's [instructions](https://hackmd.io/Er_EPZIrQSardvo4bpuYJg?view)
  - Set up a new wallet
  - Get funds for that wallet using the faucet or our new faucet-bot
  - Configure validator identities
  - Compile the ligo contract, instantiate it with those identities, and deploy the contract
  - Only now can you start the cluster using `start.sh` or tmuxinator
- All of this is quite slow, since bootstrapping the node and deploying the contract take a long time on testnets.
- This workflow depends on public infrastructure being online (and of course requires an internet connection to work, etc.)

## Solution

Fortunately, the [Flextesa](https://tezos.gitlab.io/flextesa/) project provides an easy way to run your own local Tezos network in a sandbox, to which you can deploy the consensus contract.

I've created a script `sandbox.sh` that extends the original `setup.sh` to do the following:
- setup a flextesa sandbox network using Docker
- create the validator identities/config
- compile and deploy the consensus contract to the sandbox

## Testing
Make sure you have docker installed
```
./sandbox.sh setup
./start.sh
```

You can then tear down the sandbox with `./sandbox.sh prune`


d4hines:
Comment for reviewers: The history looks messy but it's not bad. First several commits are the dependencies. Then I tried to clean up the script in stages so the changes stay clear (first renaming, then fixing the formatting, then the change).

anchpop (approved):

Zett98 (approved):

rosalogia (approved):

sandbox.sh
```diff
@@ -0,0 +1,205 @@
...
+    "action": [ "Add" ],
+    "address": "$ADDRESS"
+EOF
+    printf "  }"
+  done
+  echo ""
+  echo "]"
+}
+
+create_new_validators() {
```
EduardoRFS, line 73: This is not only creating new validators, it setup's the entire deku environment

sandbox.sh
```diff
@@ -0,0 +1,205 @@
...
+  echo "  Does the following:"
+  echo "  - Starts a Tezos sandbox network with Flextesa."
+  echo "  - Generates new validator identities."
+  echo "  - Deploys a new contract to the Tezos sandbox configured to use these validators."
+  echo "stop"
+  echo "  Stops the Tezos node"
+}
+
+case "$1" in
+start)
```
EduardoRFS, line 191: I would say rename the commands to `setup` and `prune`. As they're not safe commands, while `start` and `stop` should never destroys data.

EduardoRFS (changes requested):
Just couple changes, in general I don't like this patch but the alternatives are worse.

EduardoRFS (approved):
dragon maid approving this PR

PR #290, opened 2021-11-10
Add command line option for setting the required confirmations on setup

d4hines:
## Problem

We need a way to lower the required Tezos confirmations for doing local development and testing. However, this way should be clearly labeled as dangerous and not production-worthy.

## Solution

Add a CLI option prefixed with `unsafe`.

## Related

This PR supersedes #280


EduardoRFS (approved):
anime girl concerned about this PR

PR #302, opened 2021-11-10
crypto: Signature.{sign,verify} accepts a hash

EduardoRFS:
## Problem

Currently `Signature.{sign,verify}` accept any message, but this is not correct, a Deku node should always sign only a hash(BLAKE2B), this also reflects the behavior at Tezos where [CHECK_SIGNATURE](https://gitlab.com/tezos/tezos/-/blob/cf95d1507a13efb3ff4fb247aabb44efc0082fa7/src/lib_crypto/ed25519.ml#L338) always hashes the message before.

## Solution

This PR guarantees that on the type level by making so that the functions accept a hash directly. Because of that this removes the double hashing for a lot of operations, but for communication with Tezos it is still needed as seen at [consensus.mligo](https://github.com/marigold-dev/deku/blob/4798b2e221280036a389233000a1d600cd530317/tezos_interop/consensus.mligo#L64) where we first hash it, then pass to `Crypto.check` which will hash it again.

### DIFF
--- a/protocol/block.re
+++ b/protocol/block.re
@@ -78,7 +78,10 @@ let (hash, verify) = {
         ~handles_hash,
         ~validators_hash,
       );
-    // double hash because tezos always uses blake2b on CHECK_SIGNATURE
+    /*
+       double hash because tezos always uses blake2b on CHECK_SIGNATURE
+       https://gitlab.com/tezos/tezos/-/blob/cf95d1507a13efb3ff4fb247aabb44efc0082fa7/src/lib_crypto/ed25519.ml#L338
+     */
     let hash = BLAKE2B.hash(BLAKE2B.to_raw_string(hash));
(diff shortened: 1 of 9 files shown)


protocol/block.re
```diff
@@ -78,6 +78,8 @@ let (hash, verify) = {
         ~handles_hash,
         ~validators_hash,
       );
+    // double hash because tezos always uses blake2b on CHECK_SIGNATURE
```
d4hines, line 81: This comment isn't super clear. It would be better to flesh it out and include the link you include in the PR desc.

d4hines (approved):
LGTM but it would be nice to have a better comment.

Zett98 (approved):

protocol/block.re
EduardoRFS: done boss

PR #308, opened 2021-11-12
Add genesis block to the block pool on init

d4hines:
## Problem

Currently we don't add genesis to the block pool. This leads to the node trying to fetch the genesis block over the network, which is triggered by #227.

## Solution

Add genesis to the block pool as part of `make`.

## Related

- #227


EduardoRFS (approved):
anime girl approves this bug fix

PR #309, opened 2021-11-12
Don't search for blocks with lower block height

d4hines:
## Problem

There is a problem with the current block-fetching algorithm that causes the node to search for blocks it has already applied.

## Solution

Add a check to prevent this.

Additionally, to make it easier to interpret the error messages, I alphabetized the error variants.


EduardoRFS (approved):
anime girl approving this PR

PR #312, opened 2021-11-15
refactor: Protocol_signatures should contain an address

EduardoRFS:
## Depends

- [x] #311 

## Problem

Right now we're hashing the public key quite often due to Protocol_signatures not containing an address.

## Solution

Just add an `address` to Protocol_signatures and separate the serialization from the internal representation. Also some dead functions were removed.

### DIFF
--- a/protocol/protocol_signature.re
+++ b/protocol/protocol_signature.re
@@ -1,18 +1,45 @@
+open Helpers;
 open Crypto;
 
-[@deriving (ord, yojson)]
+[@deriving ord]
 type t = {
   // TODO: what is the name of a signature?
   signature: Signature.t,
   public_key: Wallet.t,
+  address: Address.t,
 };
 let public_key = t => t.public_key;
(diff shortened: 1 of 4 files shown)


protocol/protocol_signature.re
```diff
@@ -1,22 +1,45 @@
...
+      signature: t.signature,
+      public_key: t.public_key,
+    });
+  let of_yojson = json => {
+    let.ok {signature, public_key} = Serialized_data.of_yojson(json);
+    let address = Address.of_wallet(public_key);
+    Ok({signature, public_key, address});
+  };
+  (to_yojson, of_yojson);
+};
```
EduardoRFS, line 34: This was mostly done to avoid sending redundant data over the network(address and key), but also to ensure that the public_key matches the address.

anchpop (approved):

Zett98 (approved):

PR #314, opened 2021-11-15
Ligo: fix esy test command

anchpop:
## Problem

`esy test` ran only the zinc tests, not all ligo tests

## Solution

Edit `esy test` to run all tests


Zett98 (approved):

EduardoRFS:
Why zinctest is not registered as `runtest`?

EduardoRFS:
Also this seems to be completely unrelated to ligo.

anchpop:
@EduardoRFS This is changing Ligo's package.json. Running all the tests via `runtest` takes a long time so I added `zinctest` to only run the zinc tests

rosalogia (approved):

PR #315, opened 2021-11-15
esy: export gmp and openssl for npm release

callistonianembrace:
## Problem

`esy release` doesn't work on macos without including the dylibs in the bin wrapper.

```
dyld: Library not loaded: _____________________________________________________________________________________/i/esy_openssl-93ba2454/lib/libssl.1.1.dylib
  Referenced from: /opt/homebrew/lib/node_modules/sidechain/3/i/sidechain-22dd4d5f/bin/sidecli
  Reason: image not found
```

## Solution

This PR includes packages necessary for exported NPM package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,12 @@
   "esy": {
     "build": "dune build -p #{self.name}",
     "release": {
+      "rewritePrefix": true,
+      "includePackages": [
+        "root",
+        "esy-gmp",
+        "esy-openssl"
+      ],
       "bin": [
         "sidecli",
         "deku-node"


d4hines (approved):
LGTM

EduardoRFS:
I don't want to use `esy release` the binaries should be enough to run anything.

callistonianembrace:
>  the binaries should be enough to run anything.

They're not. Unless they're statically linked. They fail on macos. When we package them for Docker, we'll need openssl and other shared libs otherwise. I dont understand the objection to using it from the sandbox with `esy release`.

EduardoRFS:
@callistonianembrace gmp and openssl can easily be statically linked even on macOS and Windows.

EduardoRFS:
Closing this, as, we don't have plans to use `esy release` in the short term.

PR #317, opened 2021-11-16
fix: use Crypto.Random instead of Stdlib.Random

EduardoRFS:
## Problem

Currently the `find_random_validator_uri` function is using `int32` directly from the `Stdlib.Random` instead of `Crypto.Random` this may be problematic as the RNG is initialized implicitly.

## Solution

This PR fixes that by adding `Random.int32` to `Crypto` and using it directly on `building_blocks.re`

### DIFF
--- a/crypto/random.re
+++ b/crypto/random.re
@@ -3,3 +3,4 @@ Mirage_crypto_rng_unix.initialize();
 
 open Mirage_crypto_rng;
 let generate = bits => generate(bits);
+let int32 = Stdlib.Random.int32;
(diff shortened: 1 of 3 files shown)


rosalogia:
If the problem with using `int32` directly from `Stdlib.Random` was that the RNG is initialized implicitly, does this PR initialize the RNG explicitly? If so, where? If not, then how does this PR solve the problem?

EduardoRFS:
@rosalogia `Crypto.Random` doesn't need to have the RNG initialized. As it doesn't support custom RNGs.

rosalogia:
> @rosalogia `Crypto.Random` doesn't need to have the RNG initialized. As it doesn't support custom RNGs.

But it looks like `Crypto.Random` module is just calling `Stdlib.Random.int32`; how is this different?

d4hines (approved):
LGTM

PR #318, opened 2021-11-16
pin mirage crypto to forked version

Zett98:
## Problem

We need p256 public key compression and decompression, but mirage-crypto PR that adds public key compression and decompression is still under review, but all the tests pass.

## Solution

Pin mirage-crypto to the branch that adds public_key compression and decompression until the mirage-crypto version with the public_key compression pr get's released on opam.

## Related

- #198
- #22


EduardoRFS (approved):
all might thumbs up this PR

PR #313, opened 2021-11-15
Smart contracts: add .vscode settings.jsons

anchpop:
## Problem

Custom configuration was required to get vscode IDE features working with ocaml platform

## Solution

Add that configuration to .vscode/settings.json folders.

PR #319, opened 2021-11-16
Ligo: Add support for compiling options to zinc

anchpop:
## Depends

- [x] #313

## Problem

There was no way to construct the option variants `None` or `Some` in usercode

## Solution

Add implementations for the necessary constants  - also had to implement unit literals.

Work remains to be done to add support to the interpreter


Zett98:
can you format the code in separate commit? it's hard to spot the changes as the whole file is changed

Zett98 (approved):

rosalogia (approved):

PR #322, opened 2021-11-17
dune: make dtop.exe private

callistonianembrace:
## Problem

Installing Deku need not install the custom repl meant for development. `esy release` is broken because it tries to build the REPL which contains only development dependencies

## Solution

This PR makes `dtop` private and local to project

## Related

#315


EduardoRFS (approved):
anime girl dancing with a simple PR

PR #325, opened 2021-11-18
sidecli: Fix validator_address arg position

callistonianembrace:
## Problem

`add-trusted-validator` and `remove-trusted-validator` wrongly mention `validator_address` as 3rd positional argument

## Solution

This PR fixes the issue by making them 2nd positional argument

## Related

- #206


EduardoRFS (approved):
anime girl saying yes sir for this PR

PR #328, opened 2021-11-18
Remove label names from Zinc for records and tuples

Zett98:
## Depends

- [x] #321

## Problem

Records are represented as Stack_item.t LMap.t, which is a Map underneath that uses labels (which are strings) as keys
## Solution

Turn labels used by Records into ints and Map to Array to allow simple o(1) access to the field value
- #163

Issue #163, opened 2021-08-17
running ligo on sidechain


EduardoRFS:
### Old

- [x] create reproducible ligo dev environment
- [x] compile ligo to zinc
- [x] write binary wrapper for ligo
- [x] #289
- [x] modify zinc interpreter to support named terms
- [x] Everything in zinc-types that depends on ligo should be removed
- [x] zinc-types (and probably the zinc interpreter actually but I think that can wait) needs to be moved out of ligo and into a separate package 
- [x] modify zinc interpreter to support exceptions
- [ ] modify zinc interpreter to include explicit "done" operations (not sure this is a good idea...)
- [x] add strings to zinc
- [ ] implement mathy functions on zinc (hash, euclidian division, bytes.pack, etc.)
- [x] make ligo build with esy

### What is a contract?

A contract is a piece of user defined code, coupled to a piece of storage that it controls, this code can be executed externally or internally and it is allowed to trigger on chain operations.

### Contracts on deku

- [x] #293
- [ ] #294
- [ ] #295
- [ ] #296

Related: #278

anchpop:
Good first projects

- [ ] Add an operation that originates a contract 
- [ ] Add lists to Zinc
- [ ] Remove label names from Zinc (replace with arrays & integers)

d4hines:
We're doing ligo through michelson now.

PR #330, opened 2021-11-18
add booleans and if then else to zinc

### LINKED ISSUE #344: Booleans in zinc
We need booleans in the zinc interpreter and compiler

### LINKED ISSUE #345: If/then/else in zinc
Once we have booleans (#344), we should add if/then/else support to zinc

Zett98:
## Depends

- [x] #321
- [x]  #328 

## Problem

We don't support booleans and ``if ..  then .. else`` in ligo -> zinc transition 
## Solution

add support for booleans and ``if .. then .. else``
this also fixes a bug where the compilation of booleans in pascaligo would return variants instead of booleans
## Related
- #163
 - Closes #344
 - Closes #345


rosalogia (approved):

Zett98:
Added boolean operators and rebased

anchpop:
@Zett98 Going to rewrite it a bit to make if lazy (important bc of the semantics of ligo), but merging now anyway, thanks for the help

PR #332, opened 2021-11-19
Ligo: Fix top level let dependency order

anchpop:
## Problem

Pointed out to me by @Zett98 - there was a bug in how we convert top-level-lets into single expressions. The idea is to convert this:

```reason
let a = 1;
let b = a;
let c = 1;
```

to this:

```reason
let a = 1;
let b = let a = 1 in a;
let c = let a = 1 in let b = a in 1;
```

That way each top-level let is self-contained. But the logic was wrong, and the conversion was actually doing this:

```reason
let a = 1;
let b = let a = 1 in a;
let c = let b = a in let a = 1 in 1; // Order got flipped
```

Which doesn't compile haha.

## Solution

Correct the order

### DIFF
--- a/ligolang/src/passes/13deku-zincing/compiler.ml
+++ b/ligolang/src/passes/13deku-zincing/compiler.ml
@@ -392,8 +392,8 @@ let compile_module :
             tail_compile ~raise empty_environment (let_wrapper expression)
           in
           ( (fun a ->
-              make_expression_with_dependencies [ (expr_var, expression) ]
-                (let_wrapper a)),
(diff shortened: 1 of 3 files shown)


anchpop:
@EduardoRFS It's needed to make the definition of `a` available when compiling `b`, and the definitions of `a` and `b` available when compiling `c`

EduardoRFS:
@anchpop aren't top level let's just a normal let?

Zett98 (approved):

rosalogia (approved):

PR #333, opened 2021-11-19
cli: adds register-uri command for validators to update uri

callistonianembrace:
## Problem

Validators dont have a way to update their URIs. This is also needed for deploying testnets over multiple IPs.

## Solution

This PR adds a CLI command `register-uri` that uses the network endpoints `request_nonce` and `register_uri`

## Related

- #206

### DIFF
--- a/bin/sidecli.re
+++ b/bin/sidecli.re
@@ -779,6 +779,48 @@ let remove_trusted_validator = {
   );
 };
 
+let info_register_uri = {
+  let doc = "Register a URI for validators against their keys. Useful to update/assign a URI (ip address / domain / port) to a validator's wallet/keys";
+  Term.info("register-uri", ~version="%‌%VERSION%%", ~doc, ~exits, ~man);
+};
+
+let register_uri = (node_folder, updatee_validator_secret, uri: Uri.t) => {
+  open Networking;
+  let.await identity = read_identity(~node_folder);
+  let.await Request_nonce.{nonce} =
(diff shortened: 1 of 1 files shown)


bin/sidecli.re
```diff
@@ -779,6 +779,48 @@ let remove_trusted_validator = {
...
+      uri,
+    };
+  let.await () = Networking.request_register_uri(payload, identity.uri);
+  await(`Ok());
+};
+
+let register_uri = {
+  let updatee_validator_secret = {
+    let docv = "updatee validator secret";
+    let doc = "Secret key of the validator whose URI is being registered";
```
AngryStitch, line 804: Please provide `exits` and `man`:
```
  let exits = Term.default_exits in
  let man =
    [`S Manpage.s_description;
     `P "Something";
 ]
```

Our future selves will be grateful

bin/sidecli.re
```diff
@@ -779,6 +779,48 @@ let remove_trusted_validator = {
...
+    Arg.(
+      required & pos(1, some(edsk_secret_key), None) & info([], ~doc, ~docv)
+    );
+  };
+  let validator_uri = {
+    let docv = "validator uri";
+    let doc = "New URI of the validator";
+    Arg.(required & pos(2, some(uri), None) & info([], ~doc, ~docv));
```
AngryStitch, line 812: Here too!

d4hines:
This PR will be superseded by #450

PR #334, opened 2021-11-19
fix: trusted membership flow must add not remove

callistonianembrace:
## Problem

`remove-trusted-validator` is supposed to help us add a removal membership to the trusted list. (in retrospect, the command name is a misnomer too)

## Solution

This PR corrects adds the mechanism to add validator removals to the trusted list


EduardoRFS (approved):
anime girl approve this finding

PR #337, opened 2021-11-22
deku docker image

callistonianembrace:
## Problem

We need Docker images for deployments.

## Solution

This PR adds a docker file (alpine) to create docker images.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -34,6 +34,8 @@ jobs:
           file: ./docker/alpine.Dockerfile
           context: .
           platforms: linux/amd64,linux/arm64
+          cache-from: type=gha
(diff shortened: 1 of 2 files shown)


callistonianembrace:
@EduardoRFS  If the Docker builds are slowing down the core team's process, we could dispatch a custom event and have the docker pipeline in the infra repo

https://github.community/t/triggering-by-other-repository/16163

This way the core team gets to run absolutely necessary checks to make sure they can proceed with further development, while the infrastructure team can deal with packaging and distribution.

cc @ulrikstrid @oteku

callistonianembrace:
Closing this as it was decided that docker pipeline must live in a separate repo. PR re-raised here: https://github.com/marigold-dev/docker/pull/1

PR #339, opened 2021-11-22
Made LIGO dialect a test parameter

rosalogia:
## Problem

Currently, Zinc tests take as a parameter whether or not the contract is written in Reason LIGO or not, which affects the file extension of the contract file that the test searches for. This is not as flexible as it should be, since test contracts may come in any dialect of LIGO.

## Solution

Create an enum for representing dialects in the test file, and make dialect a parameter of the `expect_simple_compile_to` function. Default to PascaLIGO and determine file extension based on dialect. Change all tests where `~reason` was true to explicitly set their dialect to `Reasonligo`.

### DIFF
--- a/ligolang/src/test/zinc_test.ml
+++ b/ligolang/src/test/zinc_test.ml
@@ -57,19 +57,18 @@ type test =
   unit ->
   unit
 
-type dialect = Reasonligo | Cameligo | Pascaligo | Jsligo
-
-let expect_simple_compile_to ?(dialect = Pascaligo) ?(index = 0)
+let expect_simple_compile_to ?(dialect = Self_ast_imperative.Syntax.PascaLIGO) ?(index = 0)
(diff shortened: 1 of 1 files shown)


anchpop (approved):
Thanks!

ligolang/src/test/zinc_test.ml
```diff
@@ -57,15 +57,21 @@ type test =
   unit ->
   unit
 
-let expect_simple_compile_to ?reason:(enabled = false) ?(index = 0)
+type dialect = Reasonligo | Cameligo | Pascaligo | Jsligo
```
anchpop, line 60: Would you be able to use `Ligo_compile.Helpers.v_syntax` instead of making a new enum?

rosalogia: Most likely :+1:

Zett98 (approved):

PR #340, opened 2021-11-23
Add GitHub Actions for Docker

ulrikstrid:
## Problem

We want the docker image to be built automatically

## Solution

Adds github actions and pushes to the github registry (for now, we can add Docker hub later)

## Related

- #337

### DIFF
--- a/esy.lock/index.json
+++ b/esy.lock/index.json
@@ -3,11 +3,9 @@
   "root": "sidechain@link-dev:./package.json",
   "node": {
     "yarn-pkg-config@github:esy-ocaml/yarn-pkg-config#db3a0b63883606dd57c54a7158d560d6cba8cd79@d41d8cd9": {
-      "id":
-        "yarn-pkg-config@github:esy-ocaml/yarn-pkg-config#db3a0b63883606dd57c54a7158d560d6cba8cd79@d41d8cd9",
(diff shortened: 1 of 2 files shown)


.github/workflows/docker.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Build and push
+        # Only push if on main branch
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          file: ./docker/alpine.Dockerfile
+          context: .
+          platforms: linux/amd64,linux/arm64
+          push: github.ref == 'refs/head/main' && github.event_name == 'push'
+          tags: marigold-dev/deku:latest
```
callistonianembrace, line 38: We need commit hash in the tags to trace things when deployments need debugging.

callistonianembrace: *too

callistonianembrace: Not saying `latest` images are not needed. Perhaps another PR could deal with such tags

callistonianembrace (approved):

.github/workflows/docker.yml
ulrikstrid: I need to fix the boolean expression above, will add a commit hash tag as well 👌

ulrikstrid: Will tag with `latest` and the git sha

PR #342, opened 2021-11-23
refactor: move consensus hashing to Tezos.Deku

EduardoRFS:
## Problem

The hash functions described at `Tezos_interop.Consensus` are crucial for the core state machine of Deku but they currently live inside of `Tezos_interop` which is a library designed to describe IO communication with Tezos.

## Solution

Move the hashing functions to `Tezos.Deku.Consensus` so that it can safely be used inside of the core.

### DIFF
--- a/protocol/block.re
+++ b/protocol/block.re
@@ -71,7 +71,7 @@ let (hash, verify) = {
     // TODO: maybe should also be previous?
 
     let hash =
-      Tezos_interop.Consensus.hash_block(
+      Tezos.Deku.Consensus.hash_block(
         ~block_height,
         ~block_payload_hash,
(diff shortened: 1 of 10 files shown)


oteku:
Why is `Deku` a submodule of `Tezos` ?
From a namespacing point of view `Deku.Consensus` make more sense.

EduardoRFS:
@oteku this is the consensus of Deku living inside of Tezos, as it is defined by `consensus.mligo` which is a Tezos component from Deku PoV.

AngryStitch (approved):
@oteku and I reviewed this PR, it's good for us! Thanks!

PR #343, opened 2021-11-23
unpin mirage crypto

Zett98:
## Problem

We pinned mirage-crypto to a out fork because it didnt support public key compression.

## Solution
Our pr is merged now, so we can just use newest mirage crypto version

## Related
- #318


EduardoRFS (approved):
all mighty happy with code merged

PR #349, opened 2021-11-23
fix sandbox and start script

EduardoRFS:
## Problem

The `./sandbox.sh tear-down` is not removing old data and the `./start.sh` is currently starting too fast.

## Solution

Fix the tear-down flow of `sandbox.sh`. Also increase the start script delay to one second, this is required due to the fact that starting the node is now slower as it needs to talk to a Tezos node.


d4hines (approved):
LGTM

rosalogia (approved):

rosalogia (approved):

PR #351, opened 2021-11-29
Add static builds and use in docker

ulrikstrid:
## Problem

See #337 for reasoning

## Solution

Instead of running `esy release` etc we build deku statically and use a leaner scratch container. We might want to use busybox or something with a shell available. 

## Related

- #337 - Alternative to


EduardoRFS:
The build time of this is better, but just to confirm, there is caching right?

ulrikstrid:
There *should* be caching, now that the Dockerfile is better structured we should have more opportunities for it as well. Could you mark it as non-essential for PR merges if it's too slow?

EduardoRFS (approved):
It's still too slow to put it as a requirement, but we can merge this.
anime girl saying yes sr to a docker PR

PR #353, opened 2021-12-02
esy: use coq overrides from esy-opam-overrides

callistonianembrace:
## Depends

- [x] #352 

## Problem

Older overrides wouldn't install coq artifacts correctly. 

## Solution

Fixed and upstreamed the overrides [here](https://github.com/esy-ocaml/esy-opam-override). This PR removes local overrides for coq so that the upstream repo is looked into for overrides.

### DIFF
--- a/ligolang/esy.lock/index.json
+++ b/ligolang/esy.lock/index.json
@@ -1,5 +1,5 @@
 {
-  "checksum": "18b1de9b2251f9f3aa14bd2aa35cd125",
+  "checksum": "bc2e0e9d550babb112d764092b09ef6b",
   "root": "ligo@link-dev:./package.json",
   "node": {
     "yarn-pkg-config@github:esy-ocaml/yarn-pkg-config#db3a0b63883606dd57c54a7158d560d6cba8cd79@d41d8cd9": {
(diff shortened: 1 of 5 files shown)


ligolang/src/test/zinc_test.ml
```diff
@@ -60,6 +60,7 @@ type test =
 let expect_simple_compile_to ?(dialect = Self_ast_imperative.Syntax.PascaLIGO) ?(index = 0)
     ?(initial_stack = []) ?expect_failure ?expected_output_env ?expected_output
     contract_file (expected_zinc : Zinc_types.program) : test =
+ let open Ligo_compile.Helpers in
```
ulrikstrid, line 63: Is this supposed to be included?

callistonianembrace: Thanks for catching it

callistonianembrace: It's coming from a commit https://github.com/marigold-dev/deku/pull/353/commits/a6d7e747970986e3d40406b6e586f41a50cc3f61 while rebasing. Removing it from this PR

PR #354, opened 2021-12-09
Functorizing Zinc

Zett98:
## Problem

currently we cant change AST repr of contract/hash/key/address depending on the user input and we need a way to provide functionality to work with those values to interpreter module
## Solution

Functorize Zinc_types and Zinc_interpreter by  executor module (dummy name for now).As of right now  with this pr we represent everything as strings during compilation and testing, but for example when we will be using it in Deku we would be able to use our internal representations of hash/contract/address/key and related functionality from Deku.

## Related
- #300


anchpop (approved):

rosalogia (approved):

rosalogia:
Hm, that should've been a PR :sweat_smile: my bad.

PR #363, opened 2021-12-15
networking: {user,consensus}-operation-gossip endpoints

EduardoRFS:
## Problem

Currently the `/operation-gossip` endpoint is used for both `consensus` operations(adding and removing validators) and also for `user` operations, like transactions and withdraws. This will be problematic in the future when we proper split the operations per group.

## Solution

This PR makes two new endpoints `/user-operation-gossip` and `/consensus-operation-gossip`, each responsible for it's own kind of operations.

### DIFF
--- a/node/flows.re
+++ b/node/flows.re
@@ -300,38 +300,55 @@ let received_signature = (state, update_state, ~hash, ~signature) => {
   };
 };
 
-let received_operation =
-    (state, update_state, request: Networking.Operation_gossip.request) =>
-  if (!List.mem(request.operation, state.Node.pending_side_ops)) {
-    let.ok () =
-      Operation.Side_chain.(
-        switch (request.operation.kind) {
-        | Add_validator(_)
-        | Remove_validator(_) =>
-          let.assert () = (
-            `Invalid_signature_author,
-            Address.compare(
-              state.Node.identity.t,
-              Signature.address(
-                request.operation.Operation.Side_chain.signature,
-              ),
-            )
-            == 0,
-          );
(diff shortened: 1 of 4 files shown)


node/flows.re
```diff
@@ -300,38 +300,55 @@ let received_signature = (state, update_state, ~hash, ~signature) => {
...
+          ...state,
+          pending_side_ops: [user_operation, ...state.Node.pending_side_ops],
+        },
       );
+    Ok();
+  } else {
+    Ok();
+  };
+};
+let received_consensus_operation = (state, update_state, consensus_operation) => {
```
Zett98, line 330: Maybe be just if { expr } Ok()

node/flows.re
```diff
@@ -300,38 +300,55 @@ let received_signature = (state, update_state, ~hash, ~signature) => {
...
-            == 0,
-          );
-          Ok();
-        | _ =>
-          Lwt.async(() => {
-            Networking.broadcast_operation_gossip(state, request)
-          });
-          Ok();
-        }
+let received_user_operation = (state, update_state, user_operation) => {
```
Zett98, line 303: Maybe just if { expr} Ok()

Zett98 (approved):
Otherwise lgtm

anchpop (approved):

node/flows.re
EduardoRFS: good catch

node/flows.re
EduardoRFS: very good catch

PR #364, opened 2021-12-16
add let operators to helpers

Zett98:
## Problem

We currently don't have let operators defined for ocaml
## Solution

define them


anchpop (approved):

oteku:
For code homogeneity, you don't want `let.ok` and `let*` for same operation `Option.bind`,  etc...
This should be take during ocaml rewriting regarding that `let.ok` is not possible and depending on coding style @EduardoRFS want to push

EduardoRFS:
Agreed with @oteku , so not merging this yet

EduardoRFS:
Closing this in favor of #419, if you want to reopen, feel free.

PR #365, opened 2021-12-16
Add json test

anchpop:
## Problem

We need to convert zinc programs to json, but this is untested

## Solution

Test it


anchpop:
somehow this is a duplicate

PR #375, opened 2021-12-20
Make zinc a deku subdir

anchpop:
## Problem

We're going to need to make zinc a deku dependency in a more convenient way

## Solution

Make a `vendor` folder in deku, put zinc in it, and mark it as `vendored_dirs`


rosalogia:
Awesome, looks good :+1: Glad we worked this out

rosalogia (approved):

PR #376, opened 2021-12-20
tezos_interop: parse tezos operation as batch

EduardoRFS:
## Problem

An operation on Tezos may actually trigger many Tezos internal operations, currently the script `listen_transactions.js` dispatch each one of those internal operations individually, which is problematic as it allows to apply only a portion of all Tezos internal operations in a single hash.

Also it doesn't model well how Tezos operations happens and the script is supposed to be a low level solution.

## Solution

Dispatch all Tezos operations in a single entry instead of multiple entries. Then internally this is dispatched to multiple protocol operations.

### DIFF
--- a/bin/deku_node.re
+++ b/bin/deku_node.re
@@ -280,16 +280,7 @@ let node = folder => {
     };
   Tezos_interop.Consensus.listen_operations(
     ~context=interop_context, ~on_operation=operation =>
-    switch (
-      Flows.received_main_operation(
-        Server.get_state(),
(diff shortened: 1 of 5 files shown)


anchpop (approved):
One operation can contain multiple transaction because if you trigger a smart contract which triggers another contract etc. those are all one operation, but multiple transactions?

EduardoRFS:
@anchpop they're multiple internal operations and a single external operations.

PR #377, opened 2021-12-21
Fix typo in docker workflow

ulrikstrid:
This change will make sure we push new versions when a new push to main happens

## Problem

We don't push deku images to the docker registry

## Solution

Fix the typo in the check so that it evaluates to true on the `main` branch


oteku:
lgtm

EduardoRFS:
@oteku please always approve on github.

EduardoRFS (approved):
dragon maid saying okay to this PR

PR #378, opened 2021-12-21
add Bytes.pack/unpack

Zett98:
## Problem

We need to add Bytes.{pack/unpack} to zinc

## Solution

add a yojson encoding for now, after this we could probably add internal specialized encoding and aligning with tezos for external interactions
## Related

- #300

### DIFF
--- a/zinc/lib/zinc_interpreter/Zinc_interpreter.ml
+++ b/zinc/lib/zinc_interpreter/Zinc_interpreter.ml
@@ -52,6 +52,14 @@ module Make (D : Domain_types) = struct
         in
         let open Zinc in
         match (code, env, stack) with
+        | (Operation Unpack :: c, e, (Stack_item.Z (Plain_old_data (Bytes s'))) :: s ) -> 
+          let unpacked =  Stack_item.of_yojson @@ Yojson.Safe.from_string @@  Bytes.to_string s' in
+          let continue = Result.fold  unpacked in
+          continue ~error:(fun _ -> failwith "failed to unpack value") ~ok:(fun unpacked -> Steps.Continue (c,e, unpacked :: s))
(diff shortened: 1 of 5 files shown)


zinc/lib/zinc_interpreter/Zinc_interpreter.ml
```diff
@@ -52,6 +52,14 @@ module Make (D : Domain_types) = struct
...
         let open Zinc in
         match (code, env, stack) with
+        | (Operation Unpack :: c, e, (Stack_item.Z (Plain_old_data (Bytes s'))) :: s ) -> 
+          let unpacked =  Stack_item.of_yojson @@ Yojson.Safe.from_string @@  Bytes.to_string s' in
+          let continue = Result.fold  unpacked in
+          continue ~error:(fun _ -> failwith "failed to unpack value") ~ok:(fun unpacked -> Steps.Continue (c,e, unpacked :: s))
+        | (Operation Pack :: c, e, s' :: s ) -> 
+          let packed = Bytes.of_string @@ Yojson.Safe.to_string  @@ Stack_item.to_yojson s' in
+          let packed = Stack_item.Z (Plain_old_data (Bytes packed)) in
+          Steps.Continue (c,e, packed :: s)
```
anchpop, line 62: This will work, but I think we should add a check that the thing being packed is actually packable according to the [michelson spec](https://tezos.gitlab.io/michelson-reference/). Also, I think we may want to add a check to prevent people from packing lambdas, IIRC @EduardoRFS mentioned we may not want to support it and I'm not confident it's behavior will make sense (it will probably be fine, I'm just not confident). That said, I'm fine with merging as-is and fixing it later if you want

anchpop (approved):

Zett98:
closing in favor of #407

PR #383, opened 2021-12-22
Contract address should be hash of the operation that originated it

anchpop:
## Problem

Contract addresses are currently the hash of the signature. This is an issue because with some signature schemes signatures can be duplicated. 

## Solution

Make contract addresses be the hash of the operation (technically the hash of the operation hash, bc it's faster)


Zett98 (approved):

anchpop:
@Zett98 this pr is unfinished, we need #384

d4hines:
Closing, as I believe this is stale. @anchpop correct me if I'm wrong.

PR #384, opened 2021-12-23
Split address into implicit and contract address

anchpop:
## Depends

- [x] #396 

## Problem

Currently, address is assumed to be an implicit address. But now that we're adding contracts, an 'address' could also be a contract address. But most places of the code that talk about addresses currently really mean to talk about implicit addresses. 

## Solution

Move `Protocol.Address` into `Protocol.Address.Implicit`. Then, almost everywhere that currently references `Protocol.Address` should now reference `Protocol.Address.Implicit`.

### DIFF
--- a/src/bin/files.re
+++ b/src/bin/files.re
@@ -1,7 +1,6 @@
 open Helpers;
 open Node;
 open State;
-open Core;
 
 exception Invalid_json(string);
 let read_json = (of_yojson, ~file) => {
@@ -26,7 +25,7 @@ module Identity = {
 module Wallet = {
   [@deriving yojson]
   type t = {
-    address: Address.t,
+    address: Crypto.Key_hash.t,
     priv_key: Crypto.Secret.t,
   };
   let read = read_json(of_yojson);
@@ -35,7 +34,7 @@ module Wallet = {
 module Validators = {
   [@deriving yojson]
   type t = {
-    address: Address.t,
+    address: Crypto.Key_hash.t,
     uri: Uri.t,
   };
   let read =
--- a/src/bin/files.rei
+++ b/src/bin/files.rei
@@ -1,4 +1,3 @@
-open Core;
 open Node;
 open State;
 
@@ -11,7 +10,7 @@ module Identity: {
 
 module Wallet: {
   type t = {
-    address: Address.t,
+    address: Crypto.Key_hash.t,
     priv_key: Crypto.Secret.t,
   };
   let read: (~file: string) => Lwt.t(t);
@@ -19,8 +18,9 @@ module Wallet: {
 };
 
 module Validators: {
-  let read: (~file: string) => Lwt.t(list((Address.t, Uri.t)));
-  let write: (list((Address.t, Uri.t)), ~file: string) => Lwt.t(unit);
+  let read: (~file: string) => Lwt.t(list((Crypto.Key_hash.t, Uri.t)));
+  let write:
+    (list((Crypto.Key_hash.t, Uri.t)), ~file: string) => Lwt.t(unit);
 };
 
 module Interop_context: {
--- a/src/bin/node_state.re
+++ b/src/bin/node_state.re
@@ -1,5 +1,4 @@
 open Helpers;
-open Core;
 open Protocol;
 open Node;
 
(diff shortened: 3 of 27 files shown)


deku/node/networking.re
```diff
@@ -178,7 +178,7 @@ module Withdraw_proof = {
 module Ticket_balance = {
   [@deriving yojson]
   type request = {
-    address: Address.t,
+    address: Address.Implicit.t,
     ticket: Ticket.t,
   };
   [@deriving yojson]
```
anchpop, line 184: This is using `Address.Implicit.t` because I don't know if/how it should work for contracts.

deku/crypto/crypto.rei
```diff
@@ -19,4 +19,5 @@ module BLAKE2B_20: (module type of BLAKE2B_20);
 module Secret: (module type of Secret);
 module Key: (module type of Key);
 module Key_hash: (module type of Key_hash);
+module Operation_hash: (module type of BLAKE2B_20);
```
Zett98, line 22: this shouldn't be just blake2B_20, tezos has `o` prefix, and for contract address it's not just blake2b_20, it's blake2B_20.hash of blake2b_20.t

deku/crypto/crypto.re
```diff
@@ -10,4 +10,5 @@ module Random = Random;
 module Secret = Secret;
 module Key = Key;
 module Key_hash = Key_hash;
+module Operation_hash = BLAKE2B_20;
```
Zett98, line 13: needs to have base58 encoding derived

deku/node/networking.re
Zett98: we probably should have balance on contracts to comply with tezos, as people can send tez directly to contract

deku/protocol/address.re
```diff
@@ -1,23 +1,55 @@
 open Crypto;
-open Key_hash;
 
-[@deriving ord]
-type t = Key_hash.t;
+module Implicit = {
```
Zett98, line 3: maybe Deku_address.t and Contract_address.t?

deku/protocol/address.re
```diff
@@ -1,23 +1,55 @@
...
+let of_implicit = implicit => Implicit(implicit);
+let to_implicit = t =>
+  switch (t) {
+  | Implicit(implicit) => Some(implicit)
+  | _ => None
+  };
+
+let to_string = t =>
+  switch (t) {
+  | Implicit(t) => "dz1" ++ Key_hash.to_string(t)
```
Zett98, line 33: Key_hash will have tz1 when stringified will look like `dz1tz1<rest>` is this intended? Should we really support tz2 and tz3 inside deku as internal addresses?

Zett98 (changes requested):
Correct me if i'm wrong but i think that we need to create separate key_hash for deku which is just ed25519 wrapped in abstract type with dz1 prefix(used for validators and implicit accounts). And contract_hash which is a hash of the operation and it's type should be abstract.

anchpop:
@Zett98 

> we probably should have balance on contracts to comply with tezos, as people can send tez directly to contract

probably, but it's outside the scope of this PR

> maybe Deku_address.t and Contract_address.t?

good point, I'm not sure the names I chose are good

EduardoRFS:
@anchpop in general this design is not great.

> Address.Implicit.t

This type is both conceptually and in implementation identical to Key_hash.t, which makes most of this change noise.

EduardoRFS:
Also Contract_hash is not a crypto thing, it should not be moved to Crypto.

EduardoRFS:
@anchpop why not make this patch to main?

anchpop:
> @anchpop why not make this patch to main?

Done

rosalogia (approved):

rosalogia (commented):
I agree that the design is imperfect; if we can reduce the aforementioned noise we should, but otherwise I'm okay with being explicit about contract addresses this way.

anchpop:
> Correct me if i'm wrong but i think that we need to create separate key_hash for deku which is just ed25519 wrapped in abstract type with dz1 prefix(used for validators and implicit accounts). And contract_hash which is a hash of the operation and it's type should be abstract.

I think I spoke about this with @EduardoRFS and he said to just use tz1/kt1 for the prefixes

EduardoRFS (commented):
I forgot to post those comments, so I will be addressing them myself.

anchpop:
@EduardoRFS LGTM, do the patches address the comments you left?

EduardoRFS (approved):
anime girl smiling at a happy ending for this PR

Zett98 (approved):

PR #387, opened 2022-01-07
Add nix flake support

d4hines:
## Problem

There's no way to build Deku in Nix/NixOS.

Additionally, it would be nice to have the advantages nix provides (reproducible, declarative setup, etc).

## Solution

This PR adds a nix flake for a completely reproducible nix setup.

It's not full nix support, since nix and esy are not compatible. Instead, we use nix
to build a chroot environment with esy installed (this is how Steam and VS Code work on NixOS) .
Users can just do `nix run` to get dropped into an environment with esy installed, and then
do esy development workflow as usual. Additionally, we can install any nix package we want
by adding to the `extraPackages` list - notice `ligo` is installed this way.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,19 @@
+{
+  description = "Deku development environment";
+  inputs.nixpkgs.url = "github:nixos/nixpkgs";
+  inputs.flake-utils.url = "github:numtide/flake-utils";
+  inputs.esy-fhs.url = "github:d4hines/esy-fhs";
+  inputs.esy-fhs.inputs.nixpkgs.follows = "nixpkgs";
+  outputs = { self, nixpkgs, flake-utils, esy-fhs }:
+    flake-utils.lib.eachDefaultSystem (system:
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -0,0 +1,18 @@
+{
+  description = "Deku development environment";
+  inputs.nixpkgs.url = "github:nixos/nixpkgs";
+  inputs.flake-utils.url = "github:numtide/flake-utils";
+  inputs.esy-fhs.url = "github:d4hines/esy-fhs";
```
ulrikstrid, line 5: To de-duplicate nixpkgs you can use `follows` on `nixpkgs`
```suggestion
  inputs.esy-fhs.url = "github:d4hines/esy-fhs";
  inputs.esy-fhs.inputs.nixpkgs.follows = "nixpkgs";
```

EduardoRFS:
@ulrikstrid is this good for you? If so I will merge.

PR #388, opened 2022-01-09
@anchpop/ticket table

anchpop:
## Problem

We're going to start moving tickets around as data. (e.g. putting them in contract storage, putting them in the parameter field, etc.) This means we should have some abstraction that helps ensure tickets are not duplicated. (For example, we need to prevent contracts from sending the same ticket twice.)

## Solution

Create a ticket_table type. This is a table that you can add a `(ticket_id * Amount.t)` to, and it returns to you a *handle*. You can join two tickets in the table by joining their handles, split a ticket to get two handles, etc. 

The ticket_table api is designed to make it difficult to accidentally duplicate a ticket (although it can be done).


anchpop:
@EduardoRFS rebased against your PRs

core/ledger.re
```diff
@@ -0,0 +1,176 @@
...
+    Address_and_ticket_map.find_opt(source, ticket, t.ledger)
+    |> Option.to_result(~none=`Not_enough_funds);
+
+  let.ok (Ticket_table.{split_at, remaining}, ticket_table) =
+    Ticket_table.split(source_handle, ~at=amount, t.ticket_table)
+    |> Result.map_error(
+         fun
+         | `Invalid_ticket => failwith("Should be impossible")
+         | `Not_enough_funds as e => e,
+       );
```
anchpop, line 147: this is all over this module unfortunately and I'm not sure how to get rid of it, is this the right thing to do?

anchpop:
@EduardoRFS rebased onto main

PR #389, opened 2022-01-11
tezos_interop: promote bundles

EduardoRFS:
## Problem

Currently the bundles are not promoted and because of that `ppx_blob` fails to load leading to some weird behavior on merlin, also prettier is not running because it's on fallback mode.

## Solution

Move the bundles to promote and remove fallback from prettier.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.bundle.js binary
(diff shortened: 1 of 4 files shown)


anchpop (approved):
Approving to avoid blocking, but I'm not sure what a bundle is or what it means for a bundle to be promoted.

Guesses: 
bundle = webpack output?
promoting = writing the bundle to a file when running `dune build`?

EduardoRFS:
Closing this PR as Taquito should go away.

PR #390, opened 2022-01-11
core: Amount

EduardoRFS:
## Problem

As discussed in the past we need a way to separate the core state machine from the protocol / consensus.

## Solution

This PR is the first step in this direction by introducing the core library and moving the Amount module to it. Also using Core.Amount module everywhere.

### DIFF
--- a/bin/sidecli.re
+++ b/bin/sidecli.re
@@ -4,6 +4,7 @@ open Node;
 open State;
 open Protocol;
 open Cmdliner;
+open Core;
 
 Printexc.record_backtrace(true);
 
--- a/core/amount.re
+++ b/core/amount.re
@@ -1,4 +1,5 @@
-[@deriving ord]
+[@deriving (eq, ord)]
+// TODO: this should definitely be using Z.t
 type t = int;
 
 let zero = 0;
--- a/core/amount.rei
+++ b/core/amount.rei
@@ -1,9 +1,9 @@
-[@deriving (ord, yojson)]
+[@deriving (eq, ord, yojson)]
 type t;
+
 let zero: t;
 let (+): (t, t) => t;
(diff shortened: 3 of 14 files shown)


anchpop (approved):

EduardoRFS:
@aguillon amount should be non zero as that logically is unsound, what would it mean to have -10 EUR? You can argue that it would mean that you're in debt, but that doesn't mean that you have negative amount, it's just that you're encoding debt on the negative side of the amount.

Essentially if amount can be negative then you can have infinity money.

> In core/amount.re: should we create an exception type to replace the asserts (L9 and 14)?

For sure, opened a PR for this, will be merging this one.

aguillon:
> @aguillon amount should be non zero as that logically is unsound, what would it mean to have -10 EUR? You can argue that it would mean that you're in debt, but that doesn't mean that you have negative amount, it's just that you're encoding debt on the negative side of the amount.

I get that it must be greater than 0 (≥0), however in the code you write
`  // TODO: test this, should amount be non-zero?
  assert(t >= 0);
`
so what I understand here is that you were wondering about having `>0` instead. This is what I'm asking: although I can imagine that an actual account should never go down to 0 (because then it would not be able to pay for gas fees and whatnot), I think this should be handled at a higher level. So ≥0 is right, right?

EduardoRFS:
@aguillon yes you're correct, your interpretation of the comment is correct, can you open a PR addressing #399 and also removing this comment?

EduardoRFS:
It's a mistake of mine at 

> @aguillon amount should be non zero as that logically is unsound,

The correct one is

> @aguillon amount should be bigger or equal to zero as negative amount is logically unsound

My bad on this

PR #392, opened 2022-01-11
core: Ticket_id

EduardoRFS:
## Depends

- [x] #391

## Problem

Same as #390, but also that the Ticket name is not correct, as it doesn't encapsulate a full Ticket.

## Solution

Core.Ticket_id is introduced here to replace Protocol.Ticket, Ticket_id is a better name as it's not actually a ticket itself, this also extends the API with `[@deriving eq]`.

## Related

- #390

### DIFF
--- a/core/dune
+++ b/core/dune
@@ -1,5 +1,5 @@
 (library
  (name core)
- (libraries crypto)
+ (libraries crypto tezos)
  (preprocess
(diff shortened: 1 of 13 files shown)


rosalogia (approved):

anchpop (approved):
`Ticket` -> `Ticket_id` is a same change I made in my ticket table branch haha, glad it's done here

PR #393, opened 2022-01-11
protocol: rename some modules

EduardoRFS:
## Depends

- [x] #392

## Problem

As we're adding modules to `Core`, some of those endup having the same name as the `Protocol` one's which is specially annoying inside of the protocol as it shadows the `Protocol` module itself.

## Solution

This PR prefix a couple modules with it's library name, so that it can easily be referenced internally, ideally this could be avoided but dune doesn't allow to self reference the library. In the future all modules should move to this name schema.


protocol/operation.rei
```diff
@@ -22,14 +22,6 @@ module Main_chain: {
...
   let make:
     (~tezos_hash: Tezos.Operation_hash.t, ~tezos_index: int, ~kind: kind) => t;
-  let verify:
-    (
-      ~hash: BLAKE2B.t,
-      ~tezos_hash: Tezos.Operation_hash.t,
-      ~tezos_index: int,
-      ~kind: kind
-    ) =>
-    result(t, string);
```
EduardoRFS, line 31: This function should not be used, and it's not, that's why I removed it from the interface.

protocol/operation.rei
```diff
@@ -69,15 +61,4 @@ module Side_chain: {
...
-  let verify:
-    (
-      ~hash: BLAKE2B.t,
-      ~signature: Protocol_signature.t,
-      ~nonce: int32,
-      ~block_height: int64,
-      ~source: Address.t,
-      ~kind: kind
-    ) =>
-    result(t, string);
```
EduardoRFS, line 81: This function should not be used, and it's not, that's why I removed it from the interface.

protocol/operation.rei
aguillon: Sorry, I'm new here: why shouldn't it be used?

anchpop: It's called in of_yojson, which is the only time you should have to verify

anchpop (approved):

protocol/operation.rei
EduardoRFS: @aguillon it's an internal function, the only way to create an operation is by making one or receiving one from the network, so verify should never be called.

This function is not unsound on itself, is just a bad API.

PR #394, opened 2022-01-11
core: Ledger

EduardoRFS:
## Depends

- [x] #393 

## Problem

Same as #390.

## Solution

As Ledger is an API that was reworked it's not that bad. So mostly move it to Core.

## Related

- #390

### DIFF
--- a/protocol/dune
+++ b/protocol/dune
@@ -2,4 +2,4 @@
  (name protocol)
  (libraries helpers tezos_interop core)
  (preprocess
-  (pps ppx_deriving_yojson ppx_deriving.ord)))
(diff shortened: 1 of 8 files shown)


protocol/dune
```diff
@@ -2,4 +2,4 @@
  (name protocol)
  (libraries helpers tezos_interop core)
  (preprocess
-  (pps ppx_deriving_yojson ppx_deriving.ord)))
```
EduardoRFS, line 5: this change is not exactly related but it's a needed change for future work

rosalogia (approved):

anchpop (approved):

PR #395, opened 2022-01-11
core: User_operation, Tezos_operation, State

EduardoRFS:
## Depends

- [x] #394 

## Problem

Same as #390, additionally our current state machine is very badly defined in the Protocol, mixing things like User_operations, Tezos_operations and Consensus_operations.

## Solution

Here I re-implement the core state machine, defining it's operations and state transition functions, this should in the future be consumed by the Protocol.

## Related

- #390

### DIFF
--- a/core/tezos_operation.re
+++ b/core/tezos_operation.re
@@ -1,0 +1,48 @@
+open Helpers;
+open Crypto;
+
+[@deriving yojson]
+type internal_operation =
+  | Tezos_deposit({
+      destination: Tezos.Address.t,
+      amount: Amount.t,
+      ticket: Tezos.Ticket_id.t,
+    });
+[@deriving yojson]
+type payload = {
+  tezos_operation_hash: Tezos.Operation_hash.t,
+  internal_operations: list(internal_operation),
+};
+
+[@deriving yojson]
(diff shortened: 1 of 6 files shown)


anchpop (approved):

rosalogia (approved):

core/tezos_operation.re
```diff
@@ -0,0 +1,48 @@
+open Helpers;
+open Crypto;
+
+[@deriving yojson]
+type internal_operation =
+  | Tezos_deposit({
+      destination: Tezos.Address.t,
```
Zett98, line 7: shouldn't this be just Address.t? we don't have anything else besides tz1 addresses on deku. But anyways to make it Address.t we would also need to modify tezos_interop.

Zett98 (approved):

core/tezos_operation.re
EduardoRFS: This is a Tezos_deposit, it comes from consensus.mligo and consensus.mligo accepts a KT1 address.

EduardoRFS: Merged as this was a non blocking comment

PR #396, opened 2022-01-11
core: use State module

EduardoRFS:
## Depends

- [x] #395 

## Problem

Mostly the closing of #390, the current core state machinery lives inside of the Protocol in a loosely defined manner.

## Solution

Before at #395 the core state machine was re-implemented inside of the Core library, but was not plugged in the Protocol, this PR connects the modules and rewrite the Protocol_operation.

This is a big PR because I didn't found a good way to split the operations and state machine changes. So please **caution**.

## Related

- #390


anchpop (approved):

protocol/protocol_state.re
```diff
@@ -1,16 +1,17 @@
...
 
 type t = {
-  // state machine data
-  ledger: Ledger.t,
+  core_state: Core.State.t,
   // TODO: more efficient lookup on included_operations
   // TODO: is this part of the protocol?
-  included_operations: Operation_side_chain_set.t,
-  included_main_operations: Operation_main_chain_set.t,
+  included_tezos_operations: Tezos_operation_set.t,
```
Zett98, line 13: Why reorder fields?

EduardoRFS: As I was touching the lines I reorder it to keep consistent with everywhere else. First we handle Tezos_operations then User_operations.

PR #397, opened 2022-01-11
Add logging of run_entrypoint.js output

d4hines:
## Problem

There is currently no easy way to tell whether the block producer is actually communicating with the main chain or not.

## Solution

This PR adds a log to make it clear.


EduardoRFS (approved):
all might saying good job for this PR

PR #398, opened 2022-01-11
Fix main-chain communication

d4hines:
## Problem

Deku communication with the main chain is broken in two different ways on main:
1 The `current_validator_keys` introduced to consensus.mligo in #208 is not being sent by the Deku node
2 Ignoring that problem, Deku signatures are not recognized as valid by the main chain.
  - This latter point bears explanation. Conceptually, validators sign block hashes to show their approval of that block. But starting in #302, we replaced block hashing with a function that hashed the block twice. Thus validators started signing not the hash(block), but the hash(hash(block)). This was done to match the behavior of `CHECK_SIGNATURE` in Tezos, as mentioned in the PR. However, this introduced a bug - when sending the `block_hash` parameter, we were sending not the `hash(block)` but the `hash(hash(block))`. Tezos was then hashing this (to get `hash(hash(hash(block)))`) and checking the signature, which of course didn't match.  

## Solution

The solution 1 was to use the block pool to get the public keys associated with the signatures. The order is important, so we map the list of signatures directly to the list of keys asociated with those signatures to preserve this order.

The solution to 2 was to not send the block hash at all.  We are already calculating the block hash as part of the validation, so we just re-use that data and omit the block hash from our payload. We thus are signing the `hash(hash(block))` in deku, calculating `hash(block)` in the consensus contract, and then checking it with `CHECK_SIGNATURE`, which creates `hash(hash(block))` and checks that indeed signed it.

The two solutions are independent; however, testing them is not independent since the chain requires both to work, so I've included them both in this PR. I suggest reviewers look at 1 commit at a time. 

## Testing

Here's how I tested locally:
- I changed the `state_root_hash_epoch` in `protocol/block.re` from 60 seconds to 10 to speed things up.
- I started up the chain and let it produce a few blocks. I verified the transactions were applied in a local better-call.dev instance.
- I shut the chain down for 5 minutes
- I started the chain again and let it produce a few blocks, verifying them in bcd again.

Instead of using bcd, you can use the logging introduced in #397.
 
## Related
- #208 
- #302

### DIFF
--- a/node/state.re
+++ b/node/state.re
@@ -111,6 +111,26 @@ let try_to_commit_state_hash = (~old_state, state, block, signatures) => {
            Address_map.find_opt(wallet, signatures_map),
          )
        );
+  let signatures_of_block =
+    Block_pool.find_signatures(~hash=block.Block.hash, state.block_pool)
+    |> Option.map(Signatures.to_list);
+
+  let current_validator_keys =
+    List.map(
+      ((signer_key, _signature_opt)) => {
+        let.some signatures_of_block = signatures_of_block;
+        let.some validator_signature =
+          List.find_opt(
+            s => {
+              let key_hash = Signature.address(s) |> Address.to_key_hash;
+              Key_hash.equal(signer_key, key_hash);
+            },
+            signatures_of_block,
+          );
+        some(Signature.public_key(validator_signature));
+      },
+      signatures,
+    );
 
   Lwt.async(() => {
     /* TODO: solve this magic number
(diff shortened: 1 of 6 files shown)


d4hines:
I re-tested with @EduardoRFS's refactor and it works.

Zett98 (approved):

EduardoRFS (approved):
anime girl waving so that you don't notice that I was the one who broke the main-chain communication

PR #400, opened 2022-01-13
Exception for negative amounts (issue #399)

### LINKED ISSUE #399: Proper Amount exception
As suggested by @aguillon at #390  we should have a proper Amount exception when someone tries to create a negative Amount, while this exception should never be handled inside of the Core code, it may be useful for user facing features such as the CLI.

aguillon:
## Problem

We shouldn't be able to represent negative amounts. The previous code contained an assertion, a more explicit exception is probably better. Also, tests.

## Solution

Raise an Invalid_argument when trying to represent a negative amount.
 
 ## Related
 
 - Closes #399

### DIFF
--- a/core/amount.re
+++ b/core/amount.re
@@ -6,12 +6,15 @@ let zero = 0;
 let (+) = (+);
 let (-) = (a, b) => {
   let t = a - b;
-  assert(t >= 0);
+  if (t < 0) {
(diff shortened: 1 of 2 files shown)


aguillon:
Ah, failing tests. Interesting.

EduardoRFS (approved):
anime girl jumping of happiness with small PR

PR #401, opened 2022-01-13
Push docker image to correct registry in CI

ulrikstrid:
## Problem

We are trying to push images to the hub.docker.io registry since that is the default but we're not authenticated with it

## Solution

We already authenticate with the github registry so we just had to ad `ghcr.io/` to the beginning of the tags.


EduardoRFS (approved):
Thank you sir.
anime girl saying yes sir to the PR

Issue #374, opened 2021-12-17
update deku compiler version for algebraic effects


anchpop:

Zett98:
postponed for now

EduardoRFS:
Last try a month ago there was problems with @opam/fix but maybe it it's fixed as there is a new release.

anchpop:
I think algebraic effects may be a good fit for gasification of the zinc interpreter

d4hines:
We have OCaml5, yay. But effects are still untyped 😢

PR #402, opened 2022-01-14
Delete tmuxinator script

anchpop:
## Problem

As far as I know, nobody uses the tmuxinator script, and that leads to it being broken all the time. (It's broken right now.) If you use it, please say so and I'll close that PR! That script is the first thing I ever did at marigold so it makes me kind of sad to delete it :(

## Solution

Remove it so at least we don't have broken code lying around

# Related

#374 would be awesome because (as I understand it) it would make it simpler to abstract the network layer of the node as an effect. Once we have that, we could easily run multiple nodes in the same process, run automated tests on the consensus, etc. Then we wouldn't depend on start.sh as much.


start.sh
```diff
@@ -1,4 +1,4 @@
-#! /bin/sh
+#! /bin/bash
```
anchpop, line 1: Sneaking this into this PR haha, because it bugs me every time I accidentally write `./start.sh` and get `./start.sh: 16: Syntax error: "(" unexpected`. `sandbox.sh` is already using `/bin/bash` so at the very least we should be consistent

d4hines: It turns out that on arch, `sh` is just a link to `bash`, so that's why this has been working for me and Edu but not you. Good catch.

d4hines (approved):
LGTM

PR #407, opened 2022-01-21
@rosalogia/pack unpack zinc

rosalogia:
## Depends

- [ ] #100

## Problem

We need proper pack/unpack operations in Zinc that correspond to the data encodings in Tezos.Pack in order to communicate with Tezos.

## Solution

Add a Pack module to the Executor module-type so that we can pack/unpack in the interpreter, then plug in Tezos.Pack as the Pack module in Deku.

This is a WIP solution, it seems like some implementation details will need to be exposed in order to unpack keys/key_hashes/addresses; alternatively we can just try and directly decode the byte-string to see if we have a key/key_hash/address and extract the necessary info. Waiting for input from @anchpop in particular.

## Related

- #378 (Similar but uses Yojson instead of micheline binary encoding)


deku/zinc/lib/zinc_interpreter/Zinc_interpreter.ml
```diff
@@ -52,6 +52,40 @@ module Make (D : Domain_types) = struct
...
+              match d with
+              | Bytes b -> b
+              | _ -> failwith "Can only unpack bytes"
+            in
+            let unpacked_data =
+              match Pack.of_bytes packed_data with
+              | Pack.Int z -> Num z
+              | Pack.String s -> String s
+              | Pack.Bytes b -> Bytes b
+              | _ -> failwith "Whatever"
```
anchpop, line 85: gotta fix this of course but should be easy

anchpop (approved):

rosalogia:
Improved so that we can encode/decode keys, key_hashes and addresses. Proper list support is still a WIP but shouldn't be too much work. Just need to add some tests after that. @anchpop and @Zett98 can review at their own pace as I finish those things up.

rosalogia:
By adding conversions from Tezos.Addres to the Protocol.Address module, packing and unpacking addresses now works.

d4hines:
Can we close this PR @rosalogia ?

rosalogia:
Yep

PR #411, opened 2022-01-26
Convert to OCaml syntax

d4hines:
## Problem

We all knew this day would come... it's time to switch from Reason to OCaml syntax. Besides creating less mental overhead for switching back and forth, we want to move to OCaml 5.0, which isn't supported by Reason yet.

At this point, the diff 
## Solution

This PR was done in multiple steps:
- Removed let.await_ok - it's not really needed (only used in one place)
- Replaced all `let.x` with `let%x`, since this now supported with ppx_let_binding
- Use a script called `reason-to-ocaml.sh` to do the converstion. It use `refmt` do most of it, and then cleans up using comby.
- Did the actual rename as a separate step to preserve history
- Ran OCamlformat
- Add .git-blame-ignore-revs according to https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame

Devs should run the following to fix their git blame

```
git config blame.ignoreRevsFile .git-blame-ignore-revs
```

## TL;DR - How to fix your PR

Suppose you changed ./file1.re, etc.
When rebasing, you'll want to do something like this:

```
# Install comby
bash <(curl -sL get.comby.dev)
# On Mac
brew install comby
# With Nix
nix-shell -p comby

# Remove the current ML file and replace it with your Reason files
rm file1.re file1.rei file2.re ...
git co your-branch -- ./file1.re file1.rei ...

# Do the conversion
./reason-to-ocaml.sh ./file1.re ./file1.rei ./file2.re ...

# Rename the files with the correct extension
find ./src -name "*.rei" -exec bash -c 'mv "$1" "${1%.rei}".mli' - '{}' \;
find ./src -name "*.re" -exec bash -c 'mv "$1" "${1%.re}".ml' - '{}' \;

# Run the formatter
esy format
```
At this point, the only diff should be the diff you want.

Be sure to fix your config:
```
git config blame.ignoreRevsFile .git-blame-ignore-revs
```


EduardoRFS (approved):
This PR will be rebased so that git blame works in the future.

anime girl shaking off a massive PR away

PR #412, opened 2022-01-27
Store a list of snapshots

d4hines:
## Problem

This PR is another step towards #147. 

The problem is that we currently only store two snapshots: the latest in `state.snapshot` and the next one in `state.next_state_root`. With async state hashing we may need to store more. 
## Solution

To accomplish this, we push all snapshot logic into the snapshots module, removing the `next_state_root` property from the node state. We store all snapshots in a map, along with pointers to the current and next snapshot. We then move those pointers on each epoch.

### DIFF
--- a/bin/deku_node.re
+++ b/bin/deku_node.re
@@ -138,7 +138,7 @@ let handle_protocol_snapshot =
     (_update_state, ()) => {
       let State.{snapshots, _} = Server.get_state();
       Ok({
-        snapshot: snapshots.last_snapshot,
+        snapshot: snapshots.current_snapshot,
         additional_blocks: snapshots.additional_blocks,
         last_block: snapshots.last_block,
         last_block_signatures:
--- a/bin/deku_node.re
+++ b/bin/deku_node.re
@@ -138,8 +138,7 @@ let handle_protocol_snapshot =
     (_update_state, ()) => {
       let State.{snapshots, _} = Server.get_state();
       Ok({
-        snapshot: snd(snapshots.Snapshots.last_snapshot),
-        snapshot_hash: fst(snapshots.last_snapshot),
+        snapshot: snapshots.last_snapshot,
         additional_blocks: snapshots.additional_blocks,
         last_block: snapshots.last_block,
         last_block_signatures:
--- a/bin/node_state.re
+++ b/bin/node_state.re
@@ -75,8 +75,8 @@ let get_initial_state = (~folder) => {
       let prev_epoch_state_bin = folder ++ "/prev_epoch_state.bin";
       let.await prev_protocol =
         Files.State_bin.read(~file=prev_epoch_state_bin);
-      let next_state_root = Protocol.hash(prev_protocol);
-      await((protocol, next_state_root));
+      let (hash, data) = Protocol.hash(prev_protocol);
+      await((protocol, Snapshots.{hash, data}));
     } else {
       await((node.protocol, node.next_state_root));
     };
--- a/bin/node_state.re
+++ b/bin/node_state.re
@@ -69,18 +69,35 @@ let get_initial_state = (~folder) => {
   };
   let state_bin = folder ++ "/state.bin";
   let.await state_bin_exists = Lwt_unix.file_exists(state_bin);
-  let.await (protocol, next_state_root) =
+
+  let.await protocol =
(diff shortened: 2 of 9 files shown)


node/snapshots.rei
```diff
@@ -20,5 +22,8 @@ let make:
 
 /** this should be called when a block is received */
 let append_block: (~pool: Block_pool.t, (Block.t, Signatures.t), t) => t;
-let update:
-  (~new_snapshot: (BLAKE2B.t, string), ~applied_block_height: int64, t) => t;
+let add_snapshot:
+  (~new_snapshot: (BLAKE2B.t, string), ~block_height: int64, t) => t;
+let start_new_epoch: (~block_height: int64, t) => t;
+let get_current_snapshot: t => (BLAKE2B.t, string);
+let get_next_snapshot_opt: t => option((BLAKE2B.t, string));
```
EduardoRFS, line 29: Don't suffix with $1_opt unless you already have $1 which throws an exception.

```suggestion
let get_next_snapshot t => option((BLAKE2B.t, string));
```

node/snapshots.re
```diff
@@ -1,25 +1,36 @@
...
 let make = (~initial_snapshot, ~initial_block, ~initial_signatures) => {
-  last_snapshot: initial_snapshot,
-  last_snapshot_height: 0L,
-  last_block: initial_block,
-  last_block_signatures: initial_signatures,
-  additional_blocks: [],
+  {
+    current_snapshot: {
+      hash: initial_block.Block.state_root_hash,
+      data: "",
```
EduardoRFS, line 27:
```suggestion
      // TODO: if a snapshot is requested before first epoch starts, we will send meaningless data
      data: "",
```

node/snapshots.rei
```diff
@@ -1,24 +1,30 @@
...
     ~initial_signatures: Signatures.t
   ) =>
   t;
 
 /** this should be called when a block is received */
 let append_block: (~pool: Block_pool.t, (Block.t, Signatures.t), t) => t;
-let update:
-  (~new_snapshot: (BLAKE2B.t, string), ~applied_block_height: int64, t) => t;
+let add_snapshot: (~new_snapshot: snapshot, ~block_height: int64, t) => t;
+let start_new_epoch: t => t;
```
EduardoRFS, line 29:
```suggestion
// TODO: should those two APIs be fused? I think they happen always at same time
let add_snapshot: (~new_snapshot: snapshot, ~block_height: int64, t) => t;
let start_new_epoch: t => t;
```

bin/deku_node.re
```diff
@@ -137,9 +137,11 @@ let handle_protocol_snapshot =
     (module Networking.Protocol_snapshot),
     (_update_state, ()) => {
       let State.{snapshots, _} = Server.get_state();
+      let Snapshots.{hash: snapshot_hash, data: snapshot} =
+        snapshots.Snapshots.current_snapshot;
       Ok({
-        snapshot: snd(snapshots.Snapshots.last_snapshot),
-        snapshot_hash: fst(snapshots.last_snapshot),
+        snapshot,
+        snapshot_hash,
```
EduardoRFS, line 144:
```suggestion
        // TODO: likely this interface should fuse both of the fields below
        snapshot,
        snapshot_hash,
```

bin/node_state.re
```diff
@@ -69,18 +69,36 @@ let get_initial_state = (~folder) => {
...
       let.await prev_protocol =
         Files.State_bin.read(~file=prev_epoch_state_bin);
-      let next_state_root = Protocol.hash(prev_protocol);
-      await((protocol, next_state_root));
+      let (hash, data) = Protocol.hash(prev_protocol);
+      let block_height = prev_protocol.Protocol.block_height;
+      let snapshots =
+        Snapshots.add_snapshot(
+          ~new_snapshot=Snapshots.{hash, data},
+          ~block_height,
```
EduardoRFS, line 92:
```suggestion
      let snapshots =
        Snapshots.add_snapshot(
          ~new_snapshot=Snapshots.{hash, data},
          ~block_height=prev_protocol.block_height,
```

EduardoRFS (changes requested):

EduardoRFS (approved):
konata saying good job for Daniel's PR

PR #413, opened 2022-01-28
Remove lwt multicore resolution

d4hines:
## Problem

We no longer need the custom lwt resoultion in package.json. We aren't using `Lwt_domain` for #147, and even if we did, `Lwt_domain` is on opam now.

## Solution

Remove it.


d4hines:
Actually I think that when we do #250 we may want Lwt_domain after all so let's hold off on merging this

EduardoRFS (approved):
As discussed in private, it will not be needed in the immediate future.

anime girl dancing for a straightforward PR

PR #414, opened 2022-01-28
Async State Hashing

d4hines:
## Depends

- [x] #413
- [x] #412
- [x] #501 

## Problem

Adds full async hashing per #147 . 

## Solution

In #412, we started storing a list of snapshots. In this PR, we instead store a list of mutable references to `option(snapshot)`'s. 
We hash the protocol's on different threads using `Domainslib.Task.async`,and then updates the references when done.

**This assumes Atomic.t works out of the box with multicore! Is this true?**

Reviewers probably want to check that I'm setting up `Domainslib.Task.setup_pool` correctly. I discovered that unless I call it after `Lwt_main.run`, `Lwt_main.run` fails with:
```
Unix.fork may not be called while other domains were created
```

There is some new complexity in `deku_node.ml`. Specifically, because the `snapshot.current_snapshot` may be `None`, we keep a ref of the last snapshot available, so we can always return a snapshot to those who ask. I believe it's important to always have the latest snapshot available to other nodes to minimize synchronization time. But if you can think of a cleaner way to do it than what I've done, please let me know.

This is still missing the recovery mechanism from #250 - I'll rebase that PR to this one.

## Related
- Supersedes #243

### DIFF
--- a/src/bin/deku_node.ml
+++ b/src/bin/deku_node.ml
@@ -58,19 +58,30 @@ let handle_block_level =
     (module Networking.Block_level)
     (fun _update_state _request ->
       Ok { level = Flows.find_block_level (Server.get_state ()) })
+let latest_snapshot = ref None
 let handle_protocol_snapshot =
   handle_request
     (module Networking.Protocol_snapshot)
     (fun _update_state () ->
       let State.{ snapshots; _ } = Server.get_state () in
+      let%ok snapshot =
+        match Snapshots.get_current_snapshot snapshots with
+        | Some snapshot ->
+          latest_snapshot := Some snapshot;
+          Ok snapshot
+        | None ->
+        match !latest_snapshot with
+        | Some latest_snapshot -> Ok latest_snapshot
+        | None -> Error `Node_not_yet_initialized in
       Ok
-        {
-          snapshot = snapshots.current_snapshot;
-          additional_blocks = snapshots.additional_blocks;
-          last_block = snapshots.last_block;
-          last_block_signatures =
-            Signatures.to_list snapshots.last_block_signatures;
-        })
+        Networking.Protocol_snapshot.
+          {
+            snapshot;
+            additional_blocks = snapshots.additional_blocks;
+            last_block = snapshots.last_block;
+            last_block_signatures =
+              Signatures.to_list snapshots.last_block_signatures;
+          })
 let handle_request_nonce =
   handle_request
     (module Networking.Request_nonce)
--- a/src/bin/deku_node.ml
+++ b/src/bin/deku_node.ml
@@ -58,21 +58,12 @@ let handle_block_level =
     (module Networking.Block_level)
     (fun _update_state _request ->
       Ok { level = Flows.find_block_level (Server.get_state ()) })
-let latest_snapshot = ref None
 let handle_protocol_snapshot =
   handle_request
     (module Networking.Protocol_snapshot)
     (fun _update_state () ->
       let State.{ snapshots; _ } = Server.get_state () in
-      let%ok snapshot =
-        match Snapshots.get_current_snapshot snapshots with
-        | Some snapshot ->
-          latest_snapshot := Some snapshot;
-          Ok snapshot
-        | None ->
-        match !latest_snapshot with
-        | Some latest_snapshot -> Ok latest_snapshot
-        | None -> Error `Node_not_yet_initialized in
+      let%ok snapshot = Snapshots.get_most_recent_snapshot snapshots in
       Ok
         Networking.Protocol_snapshot.
           {
--- a/src/node/flows.ml
+++ b/src/node/flows.ml
@@ -2,6 +2,7 @@ open Helpers
 open Crypto
 open Protocol
 open Building_blocks
+open Domainslib
 module Node = State
 let write_state_to_file path protocol =
   let protocol_bin = Marshal.to_string protocol [] in
@@ -39,6 +40,7 @@ let string_of_error = function
   | `Not_all_blocks_are_signed -> "Not_all_blocks_are_signed"
   | `State_root_not_the_expected -> "State_root_not_the_expected"
   | `Snapshots_with_invalid_hash -> "Snapshots_with_invalid_hash"
+  | `Node_not_yet_initialized -> "Node_not_yet_initialized"
 
 let print_error err =
   Format.eprintf "\027[31mError: %s\027[m\n%!" (string_of_error err)
@@ -58,6 +60,7 @@ type ignore =
 let reset_timeout = (ref (fun () -> assert false) : (unit -> unit) ref)
 let get_state = (ref (fun () -> assert false) : (unit -> State.t) ref)
 let set_state = (ref (fun _ -> assert false) : (State.t -> unit) ref)
+let get_task_pool = (ref (fun () -> assert false) : (unit -> Task.pool) ref)
 let received_block' =
   (ref (fun _ -> assert false)
     : (Node.t ->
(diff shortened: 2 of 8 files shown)


bin/deku_node.re
```diff
@@ -132,19 +134,34 @@ let handle_block_level =
     Ok({level: Flows.find_block_level(Server.get_state())})
   });
 
+let empty_hash = Snapshots.{hash: BLAKE2B.hash(""), data: ""};
```
EduardoRFS, line 137:
```suggestion
// TODO: this clearly should not be here
let empty_hash = Snapshots.{hash: BLAKE2B.hash(""), data: ""};
```

bin/deku_node.re
```diff
@@ -132,19 +134,34 @@ let handle_block_level =
...
-      });
+      let.ok Snapshots.{hash: snapshot_hash, data: snapshot} =
+        switch (Snapshots.get_current_snapshot(snapshots)) {
+        | Some(snapshot) =>
+          latest_snapshot := (true, snapshot);
+          Ok(snapshot);
+        | None =>
+          let (initialized, latest_snapshot) = latest_snapshot^;
+          let.assert () = (`Node_not_yet_initialized, initialized);
+          Ok(latest_snapshot);
```
EduardoRFS, line 153: This is a problem because of

https://github.com/marigold-dev/deku/blob/68170bf7315bb88493b86e6d7314a6047c495468/node/state.re#L191

node/snapshots.rei
```diff
@@ -1,24 +1,43 @@
...
-let update:
-  (~new_snapshot: (BLAKE2B.t, string), ~applied_block_height: int64, t) => t;
+let add_snapshot_ref: (~block_height: int64, t) => (snapshot_ref, t);
+
+let set_snapshot_ref: (snapshot_ref, snapshot) => unit;
+
+/** Combines add_snapshot_ref and set_snapshot_ref in a single operation.
+    Used when synchronously hashing a protocol state.
+*/
+let add_snapshot: (~block_height: int64, snapshot, t) => t;
```
EduardoRFS, line 37: I would prefer if this function was removed, it is effectively only an util as it doesn't wrap considerable logic

node/state.re
```diff
@@ -243,7 +200,36 @@ let load_snapshot =
...
+  List.fold_left_ok(
+    (prev_state, block) => {
+      let.ok state = apply_block(prev_state, block);
+      if (BLAKE2B.equal(
+            prev_state.protocol.state_root_hash,
+            state.protocol.state_root_hash,
+          )) {
+        Ok(state);
+      } else {
+        let (hash, data) = Protocol.hash(prev_state.protocol);
```
EduardoRFS, line 222: This should definitely be in parallel. Otherwise in a tight situation the node will not be able to load the snapshot fast enough.

node/flows.re
```diff
@@ -246,19 +248,32 @@ let try_to_commit_state_hash = (~prev_validators, state, block, signatures) => {
...
 };
 
+let hash_new_state_root = (state, protocol, update_state) => {
+  let task_pool = get_task_pool^();
+  let (snapshot_ref, snapshots) =
+    Snapshots.add_snapshot_ref(
+      ~block_height=protocol.block_height,
+      state.Node.snapshots,
+    );
+  let _ =
```
EduardoRFS, line 258: with async stuff is very easy to miss a promise or a domain by accident, at least do `(_task: type)`

d4hines: Done

node/state.re
d4hines: I added a `TODO:`

node/snapshots.rei
d4hines: Done

d4hines:
Rebased with main and tested this branch today.

d4hines:
A manual test I did: inject a `Unix.sleep 5` in the function executed on the other core with `Task.async` in the `hash_new_state_root` function of `flows.ml`. I then observed that this sleep did not cause any visible slow down in block production.

src/node/flows.ml
```diff
@@ -206,14 +209,21 @@ let try_to_commit_state_hash ~prev_validators state block signatures =
...
+  update_state { state with snapshots }
+
 let rec try_to_apply_block state update_state block =
   let%assert () =
     ( `Block_not_signed_enough_to_apply,
       Block_pool.is_signed ~hash:block.Block.hash state.Node.block_pool ) in
-  let%assert () =
-    ( `Invalid_state_root_hash,
-      block_matches_current_state_root_hash state block
-      || block_matches_next_state_root_hash state block ) in
```
EduardoRFS, line 216: Sorry, why is this removed again? Shouldn't we always have the next state root hash? And if it's late it should hash it synchronously IIRC.

src/bin/deku_node.ml
```diff
@@ -58,19 +58,30 @@ let handle_block_level =
...
-            Signatures.to_list snapshots.last_block_signatures;
-        })
+        Networking.Protocol_snapshot.
+          {
+            snapshot;
+            additional_blocks = snapshots.additional_blocks;
+            last_block = snapshots.last_block;
+            last_block_signatures =
+              Signatures.to_list snapshots.last_block_signatures;
+          })
```
EduardoRFS, line 84:
```suggestion
          {
            snapshot;
            additional_blocks = snapshots.additional_blocks;
            last_block = snapshots.last_block;
            last_block_signatures =
              Signatures.to_list snapshots.last_block_signatures;
          })
```

src/bin/deku_node.ml
```diff
@@ -58,19 +58,30 @@ let handle_block_level =
     (module Networking.Block_level)
     (fun _update_state _request ->
       Ok { level = Flows.find_block_level (Server.get_state ()) })
+let latest_snapshot = ref None
```
EduardoRFS, line 61: Why isn't this on node state?

EduardoRFS (changes requested):

src/node/flows.ml
d4hines: If it's late, it does not start hashing asynchronously - this PR is fully async. 

The way it currently works is that if you miss a state root hash, we apply it anyway (because this check is removed), but you won't sign blocks. In #250, we'll add a way to sign those blocks if you later finish the state root hash. 

If we add a synchronous hashing mechanism, then there's no need to remove this check, and also no need for #250, but it may cause nodes to hash the same state on multiple cores. I think we could implement something more complex like "if you don't have the hash you need but it's being hashed on another core, wait until it's finished" - I'd have to think about it though.

src/node/server.ml
```diff
@@ -29,6 +29,18 @@ let rec reset_timeout server =
...
      reset_timeout server;
      Lwt.return_unit)
+
+let task_pool = ref None
+
+let get_task_pool () =
+  match !task_pool with
+  | Some pool -> pool
+  | None ->
+    let pool = Domainslib.Task.setup_pool ~num_additional_domains:4 () in
```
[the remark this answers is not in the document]
aguillon: @EduardoRFS should we explicitly require (and check for) a minimal number of cores in the future?

src/node/server.ml
```diff
@@ -29,6 +29,19 @@ let rec reset_timeout server =
      | Error `Not_current_block_producer -> ());
      reset_timeout server;
      Lwt.return_unit)
+
+let task_pool = ref None
```
aguillon, line 33: Not a fan, but I understand the why. I think the different parts of Deku's initialization should appear all together somewhere in the future (probably bin/node_state) to be more explicit. What do you think? @EduardoRFS @d4hines

aguillon (approved):
I have nothing against this PR. We'll discuss merging later via Slack.

src/node/flows.ml
EduardoRFS: The check can be removed but we should add a comment here, this should make you go out of sync so that you don't sign blocks and then if you finish hashing it late and it matches you can go back to being in sync(as your hypothesis was correct).

src/node/server.ml
EduardoRFS: Agreed with Daniel. Almost all global state is initialized at server.ml, the remaining are state that are not relevant to us, such as the RNG.

src/bin/deku_node.ml
EduardoRFS: Same idea as above

src/node/server.ml
aguillon: Agreed.

aguillon (approved):

src/node/snapshots.ml
```diff
@@ -73,7 +73,19 @@ let start_new_epoch t =
       additional_blocks;
     }
   | [] -> failwith "You must add a snapshot before you can start a new epoch"
-let get_current_snapshot t = Atomic.get t.current_snapshot
+
+let latest_finished_snapshot = ref None
```
EduardoRFS, line 77:
```ocaml
(* GLOBAL STATE *)
```
Please add a comment here as it's not in server.ml

src/node/snapshots.ml
```diff
@@ -34,16 +41,20 @@ let append_block ~pool (block, signatures) t =
...
+  ( atom,
+    {
+      next_snapshots = t.next_snapshots @ [(block_height, atom)];
+      current_snapshot = t.current_snapshot;
+      last_block = t.last_block;
+      last_block_signatures = t.last_block_signatures;
+      additional_blocks = t.additional_blocks;
+    } )
+let set_snapshot_ref ref_ snapshot =
+  Format.printf "\027[36m New protocol snapshot hash: %s\027[m\n%!"
```
EduardoRFS, line 55:
```ocaml
Format.eprintf
```

EduardoRFS (approved):
anime girl with a tambourine approves this PR

src/node/flows.ml
d4hines: added a comment in latest commit

PR #416, opened 2022-01-28
Add submit_user_operation helper function

anchpop:
## Problem

We want to show some information to the user before they submit an operation to the chain through the CLI - for instance, we should show the expected gas cost and whether or not the operation is expected to succeed. But that's difficult right now because the code to submit an operation is duplicated every time it appears.

## Solution

Deduplicate it. 

## Related

- #415


d4hines:
Can you rebase this with main @anchpop ?

PR #417, opened 2022-01-28
@anchpop/copy paste typos correction

anchpop:
## Depends

- [ ] #416 (not for any good reason, just noticed these while working on #415)

## Problem

There are a few typos in the codebase caused by copy-pasting. 

## Solution

Fix them

PR #418, opened 2022-01-28
Add endpoint to give information about expected gas cost of operation

anchpop:
## Depends

- [x] #417

## Problem

We want to have a way to show the user in the CLI whether or not the operation they're submitting will succeed. The CLI can't compute this itself, so we need to add an API endpoint to the node to communicate this information.

## Solution

Add an API endpoint that in the future can be extended to return information about gas cost and whether or not an operation was successful. 

## Related

- #415

### DIFF
--- a/bin/deku_node.re
+++ b/bin/deku_node.re
@@ -179,6 +179,18 @@ let handle_receive_consensus_operation =
       request.signature,
     )
   });
+let handle_operation_mock =
+  handle_request(
+    (module Networking.User_operation_mock),
+    (_, request) => {
+      let (gas_consumed, result) =
+        Flows.received_operation_mock_request(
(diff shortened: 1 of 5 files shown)


AngryStitch (approved):
Thanks for splitting commits, it makes things easier to review and follow.

Thank you for renaming the stuff you rename, it's "cosmetic" but very nice for people who'll read the code after you :D 

Like the rest of deku code, there isn't any docstring so I guess that's ok even if it makes be sad.

d4hines (changes requested):
This needs to be rebased with main.

I'd prefer if could substitute `mock` with `simulation`. This is terminology used in tezos, and I think it's more accurate (mocking conjures up thoughts about testing for me).

PR #419, opened 2022-02-01
Ppx let binding

d4hines:
## Problem
We want to migrate to OCaml syntax, but OCaml syntax does not support the custom let bindings that Reason does.

## Solution
This PR adds a PPX that allows users to define custom let expressions a la Reason. The only difference is that you have to replace the `.` with a `%` when you use the let expression

One issue is that we couldn't figure out a simple way to dynamically register the syntax extensions. Thus, we simply hard-coded a list of extensions to check for (the ones we currently use). This list can easily be extended, or if someone wants to put in the work to make it cleaner/more dynamic, they certainly can. But we felt this was good enough for a first pass.

### DIFF
--- a/ppx_let_binding/.ocamlformat
+++ b/ppx_let_binding/.ocamlformat
@@ -1,0 +1,21 @@
+version = 0.19.0
+profile = conventional
+
+leading-nested-match-parens = false
+align-constructors-decl = true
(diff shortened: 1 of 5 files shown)


EduardoRFS (approved):
anime girl dancing to the sound of another ppx

PR #423, opened 2022-02-09
Move code to ./src

d4hines:
## Problem

Having all source code in `./src` directory is a nice convention when you want to do scripting over your code (linting, etc.)

## Solution

Moves all code to ./src


EduardoRFS (approved):
konata saying good job

PR #425, opened 2022-02-10
Refactor: move print_error from deku_node to flows

d4hines:
## Problem

1. `print_error` should be re-usable in several modules
1. `print_error` is missing a few kinds of errors producable by the node
1. When we load snapshot, we supress errors, which is bad when debugging/troubleshooting.
1. Our error messages are just plain white text, no formatting. Lame.

## Solution

1.  Moved  function.
1.  Added missing errors
1. Use `print_error` to print errors when loading a snapshot.
1.  Add some formatting to the error message to make it stand out.


EduardoRFS (approved):
anime girl dancing to an approved PR

PR #426, opened 2022-02-10
add .git-blame-ignore-revs

d4hines:
## Problem

Migrating to OCaml sytnax changed every file.

## Solution

Add a file listing which commits should be ignored in
git-blame. You must configure your local git to use
it with:

git config blame.ignoreRevsFile .git-blame-ignore-revs

See https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame


EduardoRFS (approved):
konata saying good job

PR #427, opened 2022-02-10
D4hines/clean-up-ocaml-syntax

d4hines:
## Problem

Not sure how, but when we migrated ocaml syntax in 2b69a2d, I missed a few let sytnax transformations. 

## Solution

I modified the script to work on ocaml files as well and reapplied it. I checked this time to see that all the `[%custom_thing let ...`'s were gone this time.


EduardoRFS (approved):
haruhi approves this PR

PR #428, opened 2022-02-10
feat: add docker-compose.yml and improve sandbox

### LINKED ISSUE #422: Sandbox setup scripts improvements
I have a sandbox.sh script locally that is a significant improvement over the current one because it doesn't tear down the entire flextesa network every time - rather, it re-uses it and just deploys a new consensus contract.

I also regularly use the docker-compose setup from https://github.com/baking-bad/bcdhub

I need to make a proper PR that adds:
- appropriate docker-compose files
-  the improvments to the script.

d4hines:
## Problem

We currently create a new flextesa sandbox every time we run `./sandbox.sh setup`. This is unnecessarily slow.

## Solution

I've added a docker-compose file that creates a flextesa sandbox,
and also brings up a better-call-dev instance available at localhost:8000.
This is really nice for troubleshooting deku - i've been using it for a month or two and just didn't make the PR until now.

So the new workflow is:
- bring up a tezos network once with `docker-compose up -d`.
- Deploy as many deku cluseters as you want with `./sandbox.sh setup`
- If encouter issues with the better-call-dev database (which sometimes happesn when I restart my computer), tear down with `docker-compose down -v` (`-v` is important to stop the network as well).

Closes #422

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ esy test
 
 ### Running a sidechain
 
-For convenient local development, we have a included two components:
+For convenient local development, we have included two components:
 - A `docker-compose.yml` file that setups a local Tezos network
   using [Flextesa](https://tezos.gitlab.io/flextesa/) running in Docker.
   Additionally, the network sets up a [Better Call Dev](https://github.com/baking-bad/bcdhub) instance
--- a/README.md
+++ b/README.md
@@ -38,23 +38,31 @@ esy test
 
 ### Running a sidechain
 
-For convenient local development, we have a included a script `sandbox.sh` that
-automates the setup of a local Tezos network using [Flextesa](https://tezos.gitlab.io/flextesa/)
-running in Docker. Additionally, the script sets up a identities for a local Deku cluster.
-This script is the easiest way to get started with Deku; however, it uses unsafe
-configuration options to lower the required Tezos confirmations to 1. This setting greatly
-speeds up local development, **but is not at all safe for production**!
+For convenient local development, we have a included two components:
+- A `docker-compose.yml` file that setups a local Tezos network
+  using [Flextesa](https://tezos.gitlab.io/flextesa/) running in Docker.
+  Additionally, the network sets up a [Better Call Dev](https://github.com/baking-bad/bcdhub) instance
+  for introspection of the deployed contract and its operations. The BCD interace is available in
+  your browser at http://localhost:8000.
+- A script `./sandbox.sh` that sets up a identities for a local Deku cluster.
+  This script is the easiest way to get started with Deku; however, it uses unsafe
+  configuration options to lower the required Tezos confirmations to 1. This setting greatly
+  speeds up local development, **but is not at all safe for production**!
 
 #### Requirements
 
-The sandbox requires only Bash and [Docker](https://docs.docker.com/get-docker/) to be installed.
-
+The sandbox requires Bash, [Docker](https://docs.docker.com/get-docker/), and docker-compose to be installed,
+in addition to the usual Deku pre-requisites.
 #### Setup
+
+Run `docker-compose up -d` to start a Tezos network and the BCD interface. 
 
 Run `./sandbox.sh setup` to start a local Tezos sandbox network, setup three Deku validator node identities, and deploy
 a Deku consensus contract configured for these validators to the local sandbox.
 
 Run `./sandbox.sh tear-down` to kill the Tezos sandbox network and wipe the Deku state.
+
+Run `docker-compose down -v` to stop the Tezos network and destroy the BCD database.
 
 #### Start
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -44,24 +44,32 @@ esy test
 
 ### Running a sidechain
 
-For convenient local development, we have a included a script `sandbox.sh` that
-automates the setup of a local Tezos network using [Flextesa](https://tezos.gitlab.io/flextesa/)
-running in Docker. Additionally, the script sets up a identities for a local Deku cluster.
-This script is the easiest way to get started with Deku; however, it uses unsafe
-configuration options to lower the required Tezos confirmations to 1. This setting greatly
-speeds up local development, **but is not at all safe for production**!
+For convenient local development, we have a included two components:
```
gsebil08, line 47:
```suggestion
For convenient local development, we have included two components:
```

README.md
```diff
@@ -44,24 +44,32 @@ esy test
...
+- A script `./sandbox.sh` that sets up a identities for a local Deku cluster.
+  This script is the easiest way to get started with Deku; however, it uses unsafe
+  configuration options to lower the required Tezos confirmations to 1. This setting greatly
+  speeds up local development, **but is not at all safe for production**!
 
 #### Requirements
 
-The sandbox requires only Bash and [Docker](https://docs.docker.com/get-docker/) to be installed.
-
+The sandbox requires Bash, [Docker](https://docs.docker.com/get-docker/), and docker-compose to be installed,
```
gsebil08, line 60: What about a `nix-shell` to provide all these binaries?

docker-compose.yml
```diff
@@ -0,0 +1,138 @@
...
+    environment:
+      - block_time=4
+    ports:
+      - "20000:20000"
+    expose:
+      - 20000/tcp
+    logging: *bcd-logging
+
+  gui:
+    container_name: sandbox-gui
```
gsebil08, line 125: What about adding a fixed `container_name` for every container? This would avoid some troubles between this file and `sandbox.sh`.

README.md
d4hines: Hello fellow Nix friend! :raised_hands: 

I'm not sure that, e.g. `nix-shell -p docker` is enough to get a working docker setup, since it requires the docker daemon to be configured and running, etc. As for bash, I think this is sufficiently ubiquitous as to not need a Nix. In fact, I think people might be annoyed if they had to install Nix and only used it to install bash :sweat_smile: 

We could create a pure Nix setup that removes Docker entirely, but then we lose the advantage of better-call.dev maintaining and supporting this docker-compose setup for us.

In short, I'm definitely in favor of building Deku with Nix (#438 is great, and I'm going to test it shortly), but I think Docker is a reasonable system requirement for running a Deku cluster in development mode.

docker-compose.yml
d4hines: Can you clarify?

For reference, the only container we should need to reference by name is the flextesa container. The purpose is so we can use the tezos client (which is already configured correclty in the flextesa container) to deploy the contract, etc.

README.md
gsebil08: Thanks for this detailed explanation dear Nix friend 😉
I was not going so deep, but just thinking about providing the necessary binaries to run our project. I am tired of pre-requisites list with tons of tools to install.
For now, I think [`esy`](https://search.nixos.org/packages?channel=unstable&show=nodePackages.esy&from=0&size=50&sort=relevance&type=packages&query=esy), [`docker`](https://search.nixos.org/packages?channel=unstable&show=docker&from=0&size=50&sort=relevance&type=packages&query=docker), [`docker-compose`](https://search.nixos.org/packages?channel=unstable&show=docker-compose&from=0&size=50&sort=relevance&type=packages&query=docker-compose) and maybe [`opam`](https://search.nixos.org/packages?channel=unstable&show=opam&from=0&size=50&sort=relevance&type=packages&query=opam).
Providing them within a nix-shell, would allow any new people to simply do:
```shell script
$ cd deku
$ nix-shell
$ docker-compose up -d
```
without saying 
> if you want to contribute, please install .......

You can have a look at this [shell.nix](https://github.com/gsebil08/merkle-tree/blob/master/shell.nix) I provided on my merkle-tree test fr marigold 😄

docker-compose.yml
gsebil08: Sorry for this newbie comment, still discovering a lot of things!...

My question was about hard write a `container_name` for every container we use anywhere else. Hence, if I understand correctly, giving a `container_name` to the `flextesa` container should be enough!

README.md
EduardoRFS: @gsebil08 the only requirement is `esy`, `opam` is not used anywhere, for a long time I didn't have installed nor `docker` or `docker-compose either`.

EduardoRFS (approved):
anime girl dancing to finally having something reproducible

PR #429, opened 2022-02-10
tezos: Operation.forge

EduardoRFS:
## Problem

Currently Deku uses Taquito to communicate to Tezos, by calling a Node.JS script, this has been shown to not be very reliable and predictable. Alternatively we could be using the `tezos-*` libraries on opam, but by adding them as a dependency essentially the entire Tezos blockchain becomes a dependency, which makes build times worse and the dependency graph much bigger.

## Solution

In this PR I start integrating directly on the Tezos RPC, first we need to be able to forge and sign operations locally, this is done by copying the encoding from Tezos like it was done for `Pack` and combining everything in the function `Operation.forge`.

## Details

Bumping the dependencies was needed, as we were using `data-encoding.0.2`. The test was generated by dumping the communication from Taquito <-> Tezos Node.


src/tezos/gas.ml
```diff
@@ -0,0 +1,8 @@
+let () = assert (Sys.int_size = 63)
+
+(* TODO: properly implement this, saturation arithmetic *)
```
d4hines, line 3: I know Tezos has a whole saturation arithmetic module. Without copying this, can someone do something unsafe?

EduardoRFS: nope, it's still limited to (2 ** 62 - 1) because we don't provide the APIs for it, but yes if we added `( + )` here it would be unsafe.

d4hines (commented):
Looks good. I left a few questions.

tests/test_tezos_interop.re
```diff
@@ -576,6 +576,49 @@ describe("operation_hash", ({test, _}) => {
...
+      let gas_limit = Gas.of_int(1520) |> Option.get;
+      let storage_limit = Z.of_int(0);
+      let amount = Tez.of_mutez(2000000L) |> Option.get;
+      let destination =
+        Address.of_string("tz1ULf5uGJXefx8c8iLfHfuW1doMPpVicg7u")
+        |> Option.get;
+      let source =
+        Key_hash.of_string("tz1M6iKVFN8RhHjVSL3oF75nF2FJ1yMkrk5t")
+        |> Option.get;
+      /* TODO: attack, what if we put a gigantic counter */
```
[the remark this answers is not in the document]
EduardoRFS: not exactly, in general external encodings they have a check_size to limit it, and this is true even for the counter, so let me remove this comment

d4hines (approved):
LGTM
Uncle Iro approves

Zett98 (approved):

PR #430, opened 2022-02-10
Migrate to Dream

d4hines:
## Depends

- [x] #425

## Problem

We use the `opium` library for defining our http routes. We want to use `dream`. There are two reasons:
- Dream has more modern interface
- Dream's dependencies are more comatible with 5.0

## Solution

This PR does a straightforward swap of opium for dream.

~**Warning!:** There is something wrong the esy setup in this PR. deku wasn't working on my machine except if I installed `libev` with my local package manager, which doesn't seem right. I'd appreciate help debugging why.~

Edit: we patched the development scripts to work by exporting the right LD_LIBRARY_PATH. And for production it won't matter because it's statically linked.


src/bin/deku_node.ml
```diff
@@ -155,26 +119,28 @@ let node folder =
...
-    |> Lwt_main.run in
-  let forever, _ = Lwt.wait () in
-  Lwt_main.run forever
+  Node.Server.start ~initial:node;
+  Dream.initialize_log ~level:`Warning ();
+  let port = Node.Server.get_port () |> Option.get in
+  Dream.run ~port
+  @@ Dream.router
+       [
+         handle_received_block_and_signature;
```
gsebil08, line 128: This is one is present two times, is it normal?

src/node/flows.ml
```diff
@@ -9,6 +9,40 @@ let write_state_to_file path protocol =
       Lwt_io.with_file ~mode:Output path (fun oc ->
           let%await () = Lwt_io.write oc protocol_bin in
           Lwt_io.flush oc))
+
+let string_of_error = function
+  | `Added_block_has_lower_block_height -> "Added block has lower block height"
```
gsebil08, line 14: Should the error printed be more `human friendly`?

EduardoRFS: Resolving this, you're right, but there is an open issue for it. https://github.com/marigold-dev/deku/issues/64

src/bin/deku_node.ml
EduardoRFS: Nice catch, fixed on a future commit, no, it's not normal but should not be a problem either.

EduardoRFS (approved):
anime girl thumbs up to a straightforward PR

PR #431, opened 2022-02-11
Rename `handle` to `withdrawal_handle`

anchpop:
## Problem

Soon, we'll add a ticket table (#388), which will create a notion of "ticket handles". This conflicts with the current usage of "handle", which currently refers to the handles used for withdrawal.

## Solution

Rename all current usages of "handle" to "withdrawal_handle"


EduardoRFS (approved):
anime girl thumbs up this PR

PR #432, opened 2022-02-16
Defining data structures for a new consensus algorithm

AngryStitch:
(Edit from April 5 by @aguillon. This PR and the following ones are all joint work with @AngryStitch.)

## Problem

Change Deku's consensus algorithm.

## Solution

This is the first of 4 PRs aiming to implement Tendermint into Deku. Here we mostly give basic definitions that are then going to be used in the algorithm.

Please note that we don't have tests, we know that and @AngryStitch is in the process of fixing it. Moreover, this PR is not directly testable, but I'll provide instructions to help with testing in a later PR.

### Chosen consensus algorithm

The chosen algorithm is Tendermint's consensus as defined by [Algorithm 1, page 6](https://arxiv.org/pdf/1807.04938.pdf). We will refer to Tendermint's consensus as "Tendermint".

Tendermint achieves consensus through a 3 step process (Proposal, Prevote, Precommit) and works in rounds. It uses locking rules to avoid committing 2 different blocks at a same height. (See [here](https://marigold.gitbook.io/marigold/consensus/details-on-algorithms/tendermint))

### Top level types used by Tendermint

As stated in  [Algorithm 1, page 6](https://arxiv.org/pdf/1807.04938.pdf), Tendermint decides on a `value` (`src/node/tendermint_internals.ml`) at a `height`  and a `round`. The height is the chain height (called `level` in Tezos). Tendermint uses rounds to know who is the block producer. If for some reason the consensus fails at a given round, then a new round should start, with a different block producer.

A `value` can be `Nil` sometimes. It's Tendermint way of not deciding, for instance not to break lock rules.

Although the round is only important for Tendermint and probably shouldn't appear in the definition of the blocks, this makes it easier to compute the hash of a block. Why, may you ask, would you need the round to compute the hash? Because it is important that each validator signs the right block at the right round (when the algorithm succeeds), as it is linked to who is producing the block (and thus potentially rewards, security, etc.).

### `Consensus_state` (`src/node/tendermint_internals.ml`)

Every Tendermint instance has a `consensus_state`: 
```
type consensus_state = {
  mutable height : height;
  mutable round : round;
  mutable step : consensus_step;
  mutable locked_value : value;
  mutable locked_round : round;
  mutable valid_value : value;
  mutable valid_round : round;
}
```

that will allow it to enforce locking rules.

@EduardoRFS suggested making these fields non-mutable: this is delayed to a later PR, as some other things.

### DIFF
--- a/src/protocol/block.ml
+++ b/src/protocol/block.ml
@@ -10,12 +10,13 @@ type t = {
   previous_hash : BLAKE2B.t;
   author : Key_hash.t;
   block_height : int64;
+  consensus_round : int;
   operations : Protocol_operation.t list;
 }
 [@@deriving yojson]
 let hash, verify =
   let apply f ~state_root_hash ~withdrawal_handles_hash ~validators_hash
-      ~previous_hash ~author ~block_height ~operations =
+      ~previous_hash ~author ~block_height ~consensus_round ~operations =
     let to_yojson =
       [%to_yojson:
         BLAKE2B.t
@@ -24,6 +25,7 @@ let hash, verify =
         * BLAKE2B.t
         * Key_hash.t
         * int64
+        * int
         * Protocol_operation.t list] in
     let json =
       to_yojson
@@ -33,12 +35,14 @@ let hash, verify =
           previous_hash,
           author,
           block_height,
+          consensus_round,
           operations ) in
     let payload = Yojson.Safe.to_string json in
     let block_payload_hash = BLAKE2B.hash payload in
     let hash =
-      Tezos.Deku.Consensus.hash_block ~block_height ~block_payload_hash
-        ~state_root_hash ~withdrawal_handles_hash ~validators_hash in
+      Tezos.Deku.Consensus.hash_block ~block_height ~consensus_round
+        ~block_payload_hash ~state_root_hash ~withdrawal_handles_hash
+        ~validators_hash in
     let hash = BLAKE2B.hash (BLAKE2B.to_raw_string hash) in
     f (hash, block_payload_hash) in
   let hash = apply Fun.id in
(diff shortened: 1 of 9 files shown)


EduardoRFS:
Some general coding comments.

- make the mli's
- remove the mutation
- stop using Hashtbl
- remove the module alias
- make a library for the consensus

AngryStitch:
>  make the mli's

Will do.

>    remove the mutation

Will remove the mutable in `consensus_state`, if at all possible in the very short amount of time left.

>  stop using Hashtbl

Why? What do you want us to use instead?

 > remove the module alias

Ok

>  make a library for the consensus

Please clarify (why, what, where, etc)

EduardoRFS:
@AngryStitch

> Why? What do you want us to use instead?

Maps, Hashtbl's are mutable but they also have collision, which makes them quite tricky to analyze, especially in a critical environment such as a blockchain, so don't use them unless you really need that additional bit of performance, which here you don't, this lookup will not be your bottleneck.

> Please clarify (why, what, where, etc)

I'm talking about a dune library, in the same way as `core`, so that we can depend on the consensus without depending on the entire node it also defines the mental boundaries on what is this piece of code interacting with, which is very important for a review process.

d4hines:
@aguillon, what's the status of this PR? Does your work depend on or supersede it?

src/protocol/block.ml
```diff
@@ -10,12 +10,13 @@ type t = {
...
   operations : Protocol_operation.t list;
 }
 [@@deriving yojson]
 let hash, verify =
   let apply f ~state_root_hash ~withdrawal_handles_hash ~validators_hash
-      ~previous_hash ~author ~block_height ~operations =
+      ~previous_hash ~author ~block_height ~consensus_round ~operations =
     let to_yojson =
       [%to_yojson:
         BLAKE2B.t
```
rosalogia, line 22: Excited for when we switch to Bin_prot and we can just serialize record types directly

rosalogia: Actually, is this even the best we can do now? Why not use a JSON object or something? Large tuples are a little unwieldy right, since it's somewhat unclear without names/labels what each value is supposed to be?

aguillon: What's a JSON object in this context?

PR #433, opened 2022-02-16
Implementing Tendermint algorithm

AngryStitch:
## Depends

- [ ] #432 

## Problem

Deku needs a new consensus so Deku needs to implement Tendermint.

### Consensus as seen by Tendermint

Tendermint is defined at  [Algorithm 1, page 6](https://arxiv.org/pdf/1807.04938.pdf) by a series of process that are executed.

A process has type:
    ```
    type process =  CI.height -> CI.round -> CI.consensus_state -> CD.input_log ->  CD.output_log -> State.t ->  consensus_action option
    ```

### `Consensus_actions` (`src/node/tendermint_processes.ml`)
Processes trigger consensus actions:
```
type consensus_action = 
          | Broadcast         of CI.sidechain_consensus_op 
          | Schedule          of consensus_action option Clock.t
          | RestartTendermint of (CI.height * CI.round)
          | DoNothing
```

`Broadcast` refers to sending consensus messages to the network (other node and oneself) such as proposals, prevotes, precommits.

`Schedule` enables Tendermint fail-safe mecanisms: a step must be performed before the clock rings otherwise an on-timeout process will be triggered.

`RestartTendermint` deals with round (going to next round if no value was decided on) or height (going to next height when the current value is accepted) progress.

`DoNothing` when no actions on consensus state are triggered (some ontimeout actins or locking rules).

### Clocks

`Clocks` are required to trigger specific actions when a timeout is reached.

### Validators (`src/node/tendermint_internals.ml` and `src/protocol/validators.mli`)

To function properly, Tendermint needs to perform some checks:
    - answer the question "am I the current block proposer?" (propose a block/wait for proposal)
    - answer the question "is X, who send me a block proposal the valid proposer for the current round?1 (handle proposal/ignore message)
    - answer the question "what weight should I give to this proposal?"


src/node/tendermint_data.ml
```diff
@@ -0,0 +1,302 @@
...
+(* FIXME: Tendermint we could do better *)
+type node_identifier = Key_hash.t
+
+type index_step = CI.consensus_step
+
+type index = height * index_step
+(** Tendermint input is indexed by height and consensus step. *)
+
+type proposal_content = {
+  process_round : round;  (** Round of the process that sent the PROPOSAL *)
```
lykimq, line 31: For a constructor, the comment normally is a normal comment: `(* *)`

lykimq:
General comment: There is only one `.mli`, think for each file, you can add the `mli` to declare types, signatures, etc., and then do comments on these files instead of comment directly on the `ml` files.

lykimq:
General comment: There is only one `.mli`, think for each file, you can add the `mli` to declare types, signatures, etc., and then do comments on these files instead of comment directly on the `ml` files.

src/node/tendermint_data.ml
```diff
@@ -0,0 +1,302 @@
...
+  proposal : CI.value;
+  process_valid_round : round;
+      (** "Locked round" (see paper) of the PRECOMMIT of this value; -1 when the value has notbeen locked before *)
```
lykimq, line 34: You can use the markup for odoc: 
`@see <url> message`:	add a reference to the URL with a message as a comment

lykimq, line 34: has not been

src/node/tendermint_data.ml
```diff
@@ -0,0 +1,302 @@
...
+  sender : node_identifier;
+}
+(** Proposal messages carry more relevant information: the process_round, the
+   proposed value, the last known valid round and, the sender *)
```
lykimq, line 38: Normally it is: `the last known valid round and, the sender`

d4hines:
@aguillon is this ready for review? If not, please mark as draft.

PR #436, opened 2022-02-16
Committing the block decided by Tendermint

AngryStitch:
## Depends 

- [ ] #432 
- [ ] #433 
- [ ] #434 

## Problem 

We now need to commit/apply the block. Although Tendermint commits block using the output log, this should not be visible from outside of the node. Moreover, side effects (writing, broadcasting…) should be done outside of the module.

Furthermore, Deku is a sidechain and needs to be witnessed by its mainchain. To do so, we need to commit proofs that validators decided on the agreed-upon value through actual signatures of block hashes: these proofs coincide with the PrecommitOp, and have to be broadcasted to the rest of the node module as well.

## Solution

Change a lot of things.

### Removing `block_pool`

The `block_pool` was strongly linked to the previous consensus. We decided to remove it and replace the tiny bit still required by `staging_area`. In `staging_area` we keep the signatures linked to a `block_hash` that are required in `tezos_interop`.

### Broacasting

Tendermint is able of 2 categories of broadcasts:
    - `Networking.broadcast_signature` when the consensus operation is a Precommit of a block. This broadcast does 2 things: broadcast the consensus operation *and* broadcast the block hash and a signature of this block hash which will serve as proof in the future commit phase.
    - `Networking.broadcast_consensus_op` for every other consensus operation (proposal, precommit of Nil, etc) that broadcasts the consensus operation.

The broadcast always contains a crytpographic signature that can be verified to prevent man-in-the-middle issues and such.

## Known issues

- mligo: contract init is probably wrong, need another set of exes
- `Failure "Failed to  parse storage micheline expression"` after #429, need Eduardo.


aguillon:
@AngryStitch should we delete this one?

PR #437, opened 2022-02-17
chore: migrate to OCaml 5.00

EduardoRFS:
## Problem

Currently we're using an old version of OCaml multicore based on OCaml 4.12, this version is not supported at all and it's partially broken as noticed by @aguillon.

## Solution

This PR migrates to OCaml 5.00 and pin all the needed versions.


aguillon:
At first glance, it works much better on my system, thanks a lot! I'll continue working on consensus over this branch and we'll decide if everyone migrates after that or not.

I don't understand why commit 9a310d285ea99d9e891f8b5e28fbe3dc5f3498d5 removes some useful stuff in sandbox.sh (`docker run`, etc.). Did @d4hines forget to add some stuff?

PR #438, opened 2022-02-18
Pure nix setup

ulrikstrid:
## Problem

We have a nix setup currently that doesn't really work and it depends on using esy which is not ideal.

## Solution

This PR is mostly proposed as a discussion piece and I don't expect it to be merged as is.
I managed to debug the segfault and find a solution to it while doing this work.


flake.nix
```diff
@@ -1,19 +1,53 @@
...
+            dune_2
+            findlib
+            ocaml-lsp
+            ocamlformat-rpc
+            pkgs.nodejs
+            pkgs.nodePackages.npm
           ];
-        };
+
+          propagatedBuildInputs = with pkgs.ocaml-ng.ocamlPackages_multicore; [
```
gsebil08, line 29: I think that is awesome! If I understand correctly, everyone using `nix` will use the exact same version of package, and these versions will also be the same on our CI?

flake.nix
```diff
@@ -1,19 +1,53 @@
...
-          inherit system;
-          extraPackages = with pkgs; [
-            ligo
+        devShell = (pkgs.mkShell {
+          shellHook = ''
+            npm install
+            export PATH="node_modules/.bin:_build/default/src/bin:$PATH"
+          '';
+
+          packages = with pkgs.ocaml-ng.ocamlPackages_multicore; [
```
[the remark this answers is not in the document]
ulrikstrid: I'm pretty sure `nix flake update` would do the same thing in this case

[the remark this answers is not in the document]
d4hines: Indeed, nix flakes pretty much replace niv. Some people use niv from within flakes for legacy, but I don't believe we have any need to do so.

EduardoRFS:
@d4hines or @gsebil08 this looks good to any of you? If so please approve the PR

ulrikstrid:
@EduardoRFS there is one change for the esy users, I've split `package.json` into `esy.json` for esy dependencies and `package.json` for node dependencies. So any JS dependency (which you are removing anyway) that is added needs to be added to both files.

But if people are fine with this I can fixup the PR so that we can merge it. We can keep  @d4hines's esy-fhs thing in it as well since esy is the main tool for the repo.

flake.nix
ulrikstrid: Yes, that's the goal, fully reproducible development environments :)

.gitignore
```diff
@@ -1,6 +1,20 @@
+# Esy things
+esy.lock/
```
gsebil08, line 2: Should we?

.envrc
```diff
@@ -0,0 +1 @@
+use flake
```
gsebil08, line 1: Once you have installed `nix`, `nix-direnv` and `direnv` this will automatically run `use_flake` for you!

flake.nix
```diff
@@ -7,14 +7,30 @@
...
             export PATH="node_modules/.bin:_build/default/src/bin:$PATH"
           '';
+          buildInputs = with pkgs; [
+            # Make develop life easier
+            # General tooling
+            docker
+            docker-compose
+
+            # OCaml developer tooling
+            opam
```
ulrikstrid, line 28: I don't think we need opam in this project since people are either using `esy` or `nix` (if this gets merged)

flake.nix
```diff
@@ -7,14 +7,30 @@
...
+            pkgs.nodePackages.esy
```
ulrikstrid, line 29: We probably want to add @d4hines `esy-fhs` for greater compatibility

.gitignore
ulrikstrid: Probably not :sweat_smile:

flake.nix
```diff
@@ -7,14 +7,30 @@
...
+      in {
         devShell = (pkgs.mkShell {
           shellHook = ''
             npm install
             export PATH="node_modules/.bin:_build/default/src/bin:$PATH"
           '';
+          buildInputs = with pkgs; [
+            # Make develop life easier
+            # General tooling
+            docker
```
ulrikstrid, line 24: I'm not sure about docker, it's good to have it scoped to the project but I'm not sure how this will interact with globally installed docker which is probably the norm.

gsebil08: >how this will interact with globally installed docker

The one provided in this flake will not be used, and `which docker` will result to use the globally installed one.

>[...] which is probably the norm

I don't think relying on "_what is probably the norm_" is a good thing for the future.
From my personal experience, I already gave up to participate to some open-source projects because of a prerequisite list as long as my arm is... Hence, if we can provide a "_nix, it's all it takes to participate_" I think this is a good idea 😄 

This could also avoid us trouble like:
> Oh, this won't work with `whateverbin 2.X`, so please find a way to install the `1.9.x` version

since we directly provide the working ones ^^

Anyway, this is just a suggestion, since our prerequisite isn't long, I don't see a real trouble here.

flake.nix
gsebil08: What is that?

flake.nix
ulrikstrid: > The one provided in this flake will not be used, and which docker will result to use the globally installed one.

If this is true I don't think it's a problem to do this :)

flake.nix
ulrikstrid: He has a flake that constructs a environment that works for `esy` here: https://github.com/d4hines/esy-fhs

ulrikstrid: I think it's used in `main` so we can just re-add it here

.envrc
ulrikstrid: Do you know if there's a fallback that only needs `nix` and `direnv`?

gsebil08: Nope, would you mind share? 😄

.gitignore
gsebil08: Removed 😉

flake.nix
gsebil08: Maybe, I don't have any idea on "how to add it" would you mind git it a try?

ulrikstrid: Done :)

ulrikstrid: Took the freedom to remove opam as well since it's not used in Deku

.envrc
ulrikstrid: According to this we can add it ourselves: https://nixos.wiki/wiki/Flakes

```sh
use_flake() {
  watch_file flake.nix
  watch_file flake.lock
  eval "$(nix print-dev-env --profile "$(direnv_layout_dir)/flake-profile")"
}
```

ulrikstrid:
This PR is less invasive now, but we need to figure out how to handle the changes in `sandbox.sh` and `start.sh`

.gitignore
```diff
@@ -1,6 +1,19 @@
...
+# nix-direnv cache
+.direnv
+
+# Deku nodes folder
+data/
+
+# Others
 node_modules
-data
+package-lock.json
```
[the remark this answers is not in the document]
ulrikstrid: I think it's fine to ignore it, we're moving away from node anyway I think

nix/shell.nix
```diff
@@ -0,0 +1,26 @@
+{ pkgs, esy ? pkgs.nodePackages.esy, deku, npmPackages }:
+
+pkgs.mkShell {
+  shellHook = deku.preBuild;
```
d4hines, line 4: I think this should be:
```suggestion
  shellHook = deku.configurePhase;
```
It doesn't build as-is for me.

nix/shell.nix
```diff
@@ -0,0 +1,26 @@
...
+
+  nativeBuildInputs = (with pkgs; [
+    # Make developer life easier
+    ## General tooling
+    docker
+    docker-compose
+    nodejs-12_x
```
d4hines, line 11:
```suggestion
    nodejs-12_x
    ligo
```
I think we should have Ligo here too

nix/shell.nix
```diff
@@ -0,0 +1,26 @@
...
+
+    # Nix files formatter
+    nixfmt
```
d4hines, line 14:
```suggestion
    nixfmt
    nodePackages.prettier
```
We also use prettier in the `dune build @fmt` target.

d4hines: Hmm, I added it this way and that wasn't sufficient to have `prettier` on the path with direnv. `nix develop` works, but direnv doesn't. :thinking:

nix/deku.nix
```diff
@@ -0,0 +1,64 @@
...
+    src = ./..;
+    dirs = [ "src" "ppx_let_binding" "tests" ];
+    files = [ "dune-project" "sidechain.opam" "package.json" "package-lock.json" ];
+  };
+
+  configurePhase = ''
+    export NODE_PATH=${npmPackages}/node_modules
+    export PATH=${npmPackages}/node_modules/.bin:$PATH
+
+    ln -s ${npmPackages}/node_modules ./node_modules
```
d4hines, line 25: I don't think we need both `export NODE_PATH...` and `ln -s ...`.

For what it's worth, I'm getting `failed to create symbol link `./node_modules/node_modules` whenever node_modules is already tehre.

nix/shell.nix
```diff
@@ -0,0 +1,26 @@
...
+  ]) ++ (with pkgs.ocaml-ng.ocamlPackages_5_00; [
+    # OCaml developer tooling
+    ocaml
+    findlib
+    dune_2
+    # esy
+    ocaml-lsp
```
d4hines, line 22:
```suggestion
    ocaml-lsp
    ocamlformat
```
We're missing ocamlformat

d4hines: Hmm, that doesn't work - not defined.

d4hines: Oh, we have to use do something like:
```suggestion
    ocaml-lsp
    pkgs.ocamlformat
```

d4hines: Darn, the version isn't correct.

flake.nix
```diff
@@ -1,19 +1,47 @@
...
+    ocaml-overlays.inputs.nixpkgs.follows = "nixpkgs";
+
+    nix-npm-buildpackage.url = "github:serokell/nix-npm-buildpackage";
+  };
+
+  outputs = { self, nixpkgs, flake-utils, nix-npm-buildpackage, ocaml-overlays }:
     flake-utils.lib.eachDefaultSystem (system:
-      let pkgs = import nixpkgs { inherit system; }; in
+      let        
+        pkgs = ocaml-overlays.legacyPackages."${system}";
```
d4hines, line 25: What about something like this?
```suggestion
       pkgs = import nixpkgs {
           inherit system;
           overlays = [ocaml-overlays.overlay]
        };
```
This makes it more clear where to add other overlays as we need them.

nix/deku.nix
ulrikstrid: We now do different things in the shell and the build which fixes this. Also remoed the `NODE_PATH`.

nix/shell.nix
ulrikstrid: I think I solved this, but I had to set a ocamlformat version

.gitignore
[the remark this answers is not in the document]
ulrikstrid: Seems like we need package-lock.json afterall

.envrc
ulrikstrid: @d4hines do we want this snippet instead of `use flakes`?

d4hines: Yeah, I'd prefer the one with less dependencies if it works equally well.

ulrikstrid: Done

d4hines:
Superceded by #491. Thanks a ton to @ulrikstrid for doing most of the legwork!

PR #439, opened 2022-02-18
Fix docker CI

fraidev:
## Problem

The source path has been changed, and the reference to it has not changed in the dockerfile.

## Solution

Change the path used in the dockerfile.

## Related

- #423


EduardoRFS (approved):
small anime girl dancing for a PR that fixes the CI

PR #444, opened 2022-02-22
tezos_rpc: library and /inject-operations

EduardoRFS:
## Depends

- [x] #445

## Problem

As mentioned at #429 ideally we should move away from Taquito, for this one of the requirements is that we need to be able to inject operations on Tezos.

## Solution

This PR exposes `Tezos_rpc.Inject_operations.execute` to actually inject the operations on the chain, under the hood it uses #429 work to make the HTTP requests to a Tezos node.

This also defines the `tezos_rpc` library to be extended in the future.

### DIFF
--- a/src/tezos_rpc/http.ml
+++ b/src/tezos_rpc/http.ml
@@ -1,0 +1,41 @@
+open Helpers
+open Error
+
+type 'a method_ =
+  | GET : unit method_
+  | POST : string method_
+let http_request (type a) ~uri ~(method_ : a method_) (data : a) =
+  let open Piaf in
+  let%await response =
+    match method_ with
+    | GET -> Client.Oneshot.get uri
+    | POST ->
+      let body = Body.of_string data in
+      Client.Oneshot.post ~body uri in
+  match response with
+  | Ok response -> (
+    let%await body = Piaf.Body.to_string response.body in
+    match body with
+    | Ok body -> await (Ok body)
+    | Error err -> await (Error (Piaf_body err)))
+  | Error err -> await (Error (Piaf_request err))
+
+let http_request ~node_uri ~path ~method_ response_of_yojson data =
+  let uri = Uri.with_path node_uri path in
+  let%await body = http_request ~uri ~method_ data in
+  match body with
+  | Ok body -> (
+    try
+      let json = Yojson.Safe.from_string body in
+      match response_of_yojson json with
+      | Ok response -> await (Ok response)
+      | Error err -> await (Error (Response_of_yojson err))
+    with
+    | Yojson.Json_error err -> await (Error (Json_error err)))
+  | Error err -> await (Error err)
+
+let http_get ~node_uri ~path ~of_yojson =
+  http_request ~node_uri ~path ~method_:GET of_yojson ()
+let http_post ~node_uri ~path ~of_yojson ~data =
+  let data = Yojson.Safe.to_string data in
+  http_request ~node_uri ~path ~method_:POST of_yojson data
--- a/src/tezos_rpc/http.mli
+++ b/src/tezos_rpc/http.mli
@@ -1,0 +1,11 @@
+val http_get :
+  node_uri:Uri.t ->
+  path:string ->
+  of_yojson:(Yojson.Safe.t -> ('a, string) result) ->
+  ('a, Error.error) result Lwt.t
+val http_post :
+  node_uri:Uri.t ->
+  path:string ->
+  of_yojson:(Yojson.Safe.t -> ('a, string) result) ->
+  data:Yojson.Safe.t ->
+  ('a, Error.error) result Lwt.t
--- a/src/tezos_rpc/dune
+++ b/src/tezos_rpc/dune
@@ -1,0 +1,6 @@
+(library
+ (name tezos_rpc)
+ (libraries helpers piaf crypto tezos uri)
+ (modules_without_implementation error tezos_rpc)
+ (preprocess
+  (pps ppx_deriving_yojson ppx_deriving.eq ppx_let_binding)))
--- a/src/tezos_rpc/error.mli
+++ b/src/tezos_rpc/error.mli
@@ -1,0 +1,5 @@
+type error =
+  | Json_error         of string
+  | Piaf_body          of Piaf.Error.t
+  | Piaf_request       of Piaf.Error.t
+  | Response_of_yojson of string
--- a/src/tezos_rpc/inject_operations.ml
+++ b/src/tezos_rpc/inject_operations.ml
@@ -1,0 +1,12 @@
+open Tezos
+open Http
+
+type response = Operation_hash.t [@@deriving yojson]
+
+(* TODO: should I use ?async *)
+let path = "/injection/operation"
+
+let execute ~node_uri ~secret ~branch ~operations =
+  let signed_forged_operation = Operation.forge ~secret ~branch ~operations in
+  http_post ~node_uri ~path ~of_yojson:response_of_yojson
+    ~data:(`String signed_forged_operation)
--- a/src/tezos_rpc/inject_operations.mli
+++ b/src/tezos_rpc/inject_operations.mli
@@ -1,0 +1,10 @@
+open Crypto
+open Tezos
+
+type response = Operation_hash.t
+val execute :
+  node_uri:Uri.t ->
+  secret:Secret.t ->
+  branch:Block_hash.t ->
+  operations:Operation.t list ->
+  (response, Error.error) result Lwt.t
--- a/src/tezos_rpc/tezos_rpc.mli
+++ b/src/tezos_rpc/tezos_rpc.mli
@@ -1,0 +1,2 @@
+module Error = Error
+module Inject_operations = Inject_operations


src/tezos/operation.mli
```diff
@@ -2,14 +2,12 @@ open Crypto
 
 (* TODO: operation here means Manager_operation *)
 
-type parameters = {
-  entrypoint : string;
-  value : Michelson.t Data_encoding.lazy_t;
-}
 type transaction = {
```
gsebil08, line 5: [OCaml newbie question] why is this type defined twice (in .ml and .mli)? I find this pretty error-prone (like modifying in .ml and not here)

src/tezos/operation.ml
```diff
@@ -3,16 +3,11 @@ open Data_encoding
...
-  entrypoint : string;
-  value :
-    Michelson_v1_primitives.prim Tezos_micheline.Micheline.canonical
-    Data_encoding.lazy_t;
-}
 type transaction = {
   amount : Tez.t;
   destination : Address.t;
-  parameters : parameters option;
+  entrypoint : string;
```
gsebil08, line 9: At this point, I am not sure about what is _really_ the entrypoint (I think it is _the smart contract entrypoint_, am I right?), but should we go for a custom type instead of `string`?
I think going for a custom type could help us to get some domain specific: What is this string, functionally? => Can this string be empty? Can this string be only a space character? Does it have to match a given regex?

src/tezos/michelson.ml
```diff
@@ -7,4 +7,10 @@ type t = Michelson_v1_primitives.prim Tezos_micheline.Micheline.canonical
 let expr_encoding =
   Micheline_encoding.canonical_encoding ~variant:"michelson_v1"
     Michelson_v1_primitives.prim_encoding
-let lazy_expr_encoding = Data_encoding.lazy_encoding expr_encoding
+
+let unit =
+  Micheline.strip_locations (Prim (-1, Michelson_v1_primitives.D_Unit, [], []))
+let is_unit value =
```
gsebil08, line 13: Is there a point to define this function globally while it is used only once in `operation:encoding`?

tests/test_tezos_interop.re
```diff
@@ -577,46 +577,102 @@ describe("operation_hash", ({test, _}) => {
...
+        ~secret="edsk4RbgwutwsEdVNuJsE5JDsxeJ6qFcG8F5rKFGnj5finT6FV46sd",
+        ~entrypoint="default",
+        ~value=Michelson.unit,
+      );
 
     let taquito_forged_bytes = "3d95683f0d29a6deb044f4ecd86efd9cbae6b373b7b9d7c2db16456783e664566c001004051072b588b39e25b9f4dbf4673abcd63147bb03addfc901f00b0080897a00005f70062003e798791cb04f51bcec1d3358ac64a600468a797b657cff8b585c18cb599443d7c0982220cfa9231270138c7a3e0996d8ea42a2912d51c44681877b2d3a3e8051ebd1498305f8f73cf68967f60349fc00";
     expect.string(forged_bytes).toEqual(taquito_forged_bytes);
-  })
+  });
+  /*
```
gsebil08, line 635: Why do we keep these lines? If they are important they should not be comments but documentation.
If they are not important and it is just "a way to retrieve them if needed" this is the job of GitHub, what do you think?

tests/test_tezos_interop.re
```diff
@@ -577,46 +577,102 @@ describe("operation_hash", ({test, _}) => {
   });
 });
 describe("operation forging", ({test, _}) => {
```
gsebil08, line 579: What about a clear convention for tests, something like 
>When calling _{function}_, successfully _{expected action}_

Because with the actual description, I don't know what is the scope of this test. Maybe the case `when calling forging with bad parameters, fails with this error` is not covered...

src/tezos_rpc/http.ml
```diff
@@ -0,0 +1,41 @@
+open Helpers
+open Error
+
+type 'a method_ =
```
gsebil08, line 4: I don't see any test module, is it normal?

src/tezos_rpc/http.mli
```diff
@@ -0,0 +1,11 @@
+val http_get :
```
gsebil08, line 1: I feel like only the `http_post` is used, then should we already define it?

src/tezos_rpc/http.ml
```diff
@@ -0,0 +1,41 @@
...
+  | Ok response -> (
+    let%await body = Piaf.Body.to_string response.body in
+    match body with
+    | Ok body -> await (Ok body)
+    | Error err -> await (Error (Piaf_body err)))
+  | Error err -> await (Error (Piaf_request err))
+
+let http_request ~node_uri ~path ~method_ response_of_yojson data =
+  let uri = Uri.with_path node_uri path in
+  let%await body = http_request ~uri ~method_ data in
```
gsebil08, line 25:
```suggestion
  let%await body = http_request uri ~method_ data in
```

gsebil08: If I understand correctly, of course 😅

gsebil08 (commented):
First real review, hope my comments can be useful...

Zett98 (approved):

src/tezos/operation.mli
EduardoRFS: The mli is an interface for the ml(the implementation), so if you change one but not the other you're gonna get an error, think about mli in a similar fashion to a C header but checked against the implementation.

src/tezos/operation.ml
EduardoRFS: Yes, this is the contract entrypoint. About the functionality, it can be any string if I remember correctly up to 61 chars.

Your argument about enforcing it on the typer makes sense, just that it's not a considerable win here, but an considerable amount of complexity, especially because this entrypoint is likely gonna be a hard coded string.

I would say that this should not block this PR, but if you want to open another PR adding an `Entrypoint.t` feel free.

src/tezos/michelson.ml
EduardoRFS: Important thing, it's not "globally", it's inside of a module, which is inside of a library, while it is accessible globally it's not like it's polluting the scope. It's quite different from Haskell in this sense.

But this was on the other PR and the reasons why I added on Michelson instead of just doing it inline, is because I was following a similar style from Tezos and thinking now(after the fact, which is biased), it makes sense conceptually, I would prefer to avoid comparing Michelson outside of this module in general.

tests/test_tezos_interop.re
EduardoRFS: That was a mistake of mine, I forgot it during the development, but noticed before merging, it was on the previous PR tho.

tests/test_tezos_interop.re
EduardoRFS: Please open an issue for this, but those tests are mostly regression test, they're not trying to enforce failure properties, just to ensure that the cases that we actually care are covered here.

src/tezos_rpc/http.ml
EduardoRFS: Currently we don't have a lot of automated tests, and we're in crunch mode, but @d4hines is working on having more test infrastructure.

src/tezos_rpc/http.mli
EduardoRFS: It's gonna be used in a future PR(which is now already opened), also it's a 2 lines implementation, so it's quite reasonable.

src/tezos_rpc/http.ml
EduardoRFS: This typed, as you can see from the CI. But also nope, this implementation is correct.

https://github.com/marigold-dev/deku/pull/444/files#diff-698e4c768874ed323e94fa37737b6039c3ef8f7a6f5222ecd43cfaab42d59c3bR7

anchpop (approved):

src/tezos_rpc/http.ml
gsebil08: Thanks, I guess I had something weird with my environment, since it showed me `unused variable`. Working great now. Thanks!

src/tezos_rpc/http.ml
gsebil08: What about basic testing?
For instance, even if we can suppose that `yojson` is working, having tests like:
> when calling to_yojson on this object, it produces the following JSON

or also

> when calling to_yojson then from_yojson, it gives back the original object

If we always say "we can do that later" we will never do it... Moreover, some tests are just quick-win and ensure our domain is safe in the real world.

For instance, I think the function `http_request` isn't so difficult to test, am I wrong?
Or maybe we can start by only testing `http_post` function, which is exposed and used, and test the others later.
What do you think?

src/tezos/operation.ml
gsebil08: Opened #452

src/tezos/michelson.ml
gsebil08: Didn't see it was inside a module. Thanks.

tests/test_tezos_interop.re
gsebil08: Created #453

src/tezos_rpc/http.ml
gsebil08: Opened #454

PR #446, opened 2022-02-23
tezos_rpc: Fetch_block_operations

EduardoRFS:
## Depends

- [x] #444

## Problem

Similar to #444, another requirement to move away from Taquito is to be able to fetch the operations of a block, to detect interactions with `consensus.mligo`.

## Solution

Implements `Tezos_rpc.Fetch_block_operations`, this API provides a way to fetch the operations from a specific chain and block_hash.

**warning**: the implementation is not ideal as the JSON encoded by Tezos is quite complex, so this PR just abuses of `ppx_deriving_yojson` to generate some simplified decoders, in the current stage this only decode transactions and internal transactions.

### DIFF
--- a/src/crypto/base58.mli
+++ b/src/crypto/base58.mli
@@ -48,6 +48,8 @@ module Prefix : sig
   val ed25519_signature : string
   val secp256k1_signature : string
   val p256_signature : string
+
+  val chain_id : string
 end
 
 module Alphabet : sig
--- a/src/tezos_rpc/fetch_block_operations.ml
+++ b/src/tezos_rpc/fetch_block_operations.ml
@@ -1,0 +1,176 @@
+open Helpers
+open Crypto
+open Tezos
+open Http
+
+type parameters = {
+  entrypoint : string;
+  value : Michelson.t;
+}
+type internal_operation =
+  | Internal_transaction     of {
+      sender : Address.t;
+      destination : Address.t;
+      parameters : parameters option;
+    }
+  | Internal_non_transaction
+
+type operation_status =
+  (* if one of the internal operations the status will be "backtracked" *)
+  | Applied
+  | Other
+type operation =
+  | Transaction     of {
+      source : Key_hash.t;
+      status : operation_status;
+      internal_operations : internal_operation list;
+    }
+  | Non_transaction
+type block_operation = {
+  (* TODO: should protocl also be here? *)
+  (* TODO: should signature also be here? *)
+  hash : Operation_hash.t;
+  chain : Chain_id.t;
+  branch : Block_hash.t;
+  contents : operation list;
+}
+type response = block_operation list
+
+module Decoder = struct
+  (* TODO: what to do when some operations are not known?
+        Let's say by a bug or because the format changed? *)
+  (* TODO: also, how is the failure mode, exceptions or result???*)
+  type kind =
+    | Kind_transaction
+    | Kind_not_transaction
+  let kind_of_yojson json =
+    (* TODO: probably can be done better *)
+    let%ok kind = [%of_yojson: string] json in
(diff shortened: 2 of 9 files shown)


src/tezos_rpc/fetch_block_operations.ml
```diff
@@ -0,0 +1,176 @@
...
+      sender : Address.t;
+      destination : Address.t;
+      parameters : parameters option;
+    }
+  | Internal_non_transaction
+
+type operation_status =
+  (* if one of the internal operations the status will be "backtracked" *)
+  | Applied
+  | Other
```
anchpop, line 21: I don't understand this comment, why is "backtracked" not one of the variants if it's a possible status?

anchpop (approved):
I don't have the knowledge of tezos necessary to know whether or not this code is correct, if this is important let me know and I'll learn the tezos format so I'm able to review it more thoroughly 

btw, I'm surprised there's not some ocaml library maintained as part of tezos for doing this?

EduardoRFS:
@anchpop There is an OCaml library, that's the tezos-client, the problem is that it pulls the entire Tezos codebase with it, that's what we're trying to avoid.

src/tezos_rpc/fetch_block_operations.ml
EduardoRFS: This PR intentionally tries to over simplify the format, there is like 5 or 6 status, in the future we can improve this, maybe by removing it from Tezos directly.

src/crypto/base58.mli
```diff
@@ -48,6 +48,8 @@ module Prefix : sig
   val ed25519_signature : string
   val secp256k1_signature : string
   val p256_signature : string
+
+  val chain_id : string
```
gsebil08, line 52: Comment only about format. Since these types are grouped, is it possible to add one short comment line before each category?
Even if it is useless like:
```ocaml
  (* signatures definitions *)
  val ed25519_signature : string
  val secp256k1_signature : string
  val p256_signature : string
```
the goal is to add some guidelines for people coming from nowhere _(for example: me 🤣)_

src/tezos_rpc/fetch_block_operations.ml
```diff
@@ -0,0 +1,176 @@
...
+}
+type internal_operation =
+  | Internal_transaction     of {
+      sender : Address.t;
+      destination : Address.t;
+      parameters : parameters option;
+    }
+  | Internal_non_transaction
+
+type operation_status =
```
gsebil08, line 18: [Format question] Why is there a blank line between the definition of `type internal_operation` and this one, but not between this one and the following `type operation`? 🤔

EduardoRFS: Mostly this is a grouping cue, `internal_operation` and `parameters` make a single entity an `internal_operation`, where `operation_status`, `operation`(bad name) and `block_operation` makes an `block_operation`.

src/crypto/base58.mli
EduardoRFS: Hmm, this file is partially extracted from Tezos, as can be seen by the license on top of the file.

I would not agree that adding a comment would help, if someone doesn't notice the pattern at `_signature` they're even less likely to read the comment.

But please open an issue to discuss this,

aguillon (approved):

src/crypto/base58.mli
gsebil08: >Hmm, this file is partially extracted from Tezos

Ok, I didn't know that.

>But please open an issue to discuss this

Ok, I will open an issue when I will feel a need while coding to give concrete example.

gsebil08 (approved):

PR #447, opened 2022-02-23
tezos_rpc: listen to blocks on Tezos

EduardoRFS:
## Depends

- [x] #446 

## Problem

Following #446, in the mission of removing Taquito, being able to listen for blocks happening on Tezos is required.

## Solution

Because Tezos may have reorgs, this PR exposes two APIs, `Tezos_rpc.Listen_to_blocks` which allows to listen to all the blocks that were validated by the Tezos node and also `Tezos_rpc.Listen_to_chain_heads` which allows to listen to all blocks elected as the head of a chain.

With both of those APIs we can keep track of all the branches happening on Tezos and based on the last head, find the operations needed.

## Implementation

Because we use connections forever-lived to `/monitor` and Yojson doesn't seems to support incremental reading from an infinite stream, this PR abuses of OCaml 5.00 effects to workaround this.


anchpop (approved):

aguillon (approved):

src/tezos_rpc/http.ml
```diff
@@ -39,3 +39,109 @@ let http_get ~node_uri ~path ~of_yojson =
 let http_post ~node_uri ~path ~of_yojson ~data =
   let data = Yojson.Safe.to_string data in
   http_request ~node_uri ~path ~method_:POST of_yojson data
+
+(* The two functions below, make_lazy_lexbuf and lazy_json_from_stream
+   exists to provide a way to continuously transform an string Lwt_stream into
```
gsebil08, line 44:
```suggestion
   exist to provide a way to continuously transform a string Lwt_stream into
```

src/tezos_rpc/http.ml
```diff
@@ -39,3 +39,109 @@ let http_get ~node_uri ~path ~of_yojson =
...
+    | Poll_data ->
+      Some (fun (cont : (unit, unit) Deep.continuation) -> waiting := Some cont)
+    | _ -> None in
+
+  let () = Deep.try_with read lexbuf { effc = handler } in
+  feed
+
+(* TODO: what is the failure modes of this?
+         Any possible memory leak?
+         What happens if invalid JSON on the input?
```
gsebil08, line 98: I guess we could figure it out with a good testing strategy 😇 
Or maybe we would be like "it is impossible to provide an invalid JSON" 😇

src/tezos_rpc/http.ml
```diff
@@ -39,3 +39,109 @@ let http_get ~node_uri ~path ~of_yojson =
...
+    (fun str ->
+      feed str;
+
+      let results = List.rev !pending in
+      pending := [];
+      results)
+    string_stream
+
+let http_get_listen ~uri =
+  let open Piaf in
```
gsebil08, line 122: Oh! I didn't know it is possible to do that!
Useless comment, just to share I learned a new thing 😅

gsebil08 (approved):

src/tezos_rpc/http.ml
EduardoRFS: Both are feasible, but I will be revisiting this modules in the near future, so merging to simplify my life.

PR #448, opened 2022-02-23
tezos_rpc: add Fetch_block_header

EduardoRFS:
## Depends

- [x] #447

## Problem

Following #447, in the missiong of moving away from Taquito, sometimes we may not receive a block, because the connection with the Tezos RPC was temporarily broken, so we need to be able to fetch old blocks header.

## Solution

Implements `Tezos_rpc.Fetch_block_header`, this will call Tezos asking for an specific block header.

### DIFF
--- a/src/tezos_rpc/fetch_block_header.ml
+++ b/src/tezos_rpc/fetch_block_header.ml
@@ -1,0 +1,23 @@
+open Tezos
+open Http
+
+type response = Block_header.t
+
+let path ~chain ~block_hash =
+  (* TODO: I don't like this Format.sprintf *)
+  Format.sprintf "/chains/%s/blocks/%s/header" chain block_hash
+
+let execute ~node_uri ~chain ~block_hash =
+  let chain =
+    match chain with
+    | Some chain -> Chain_id.to_string chain
+    | None -> "main" in
+
+  let block_hash =
+    match block_hash with
+    | Some block_hash -> Block_hash.to_string block_hash
+    (* TODO: we could also query by height *)
+    | None -> "head" in
+
+  let path = path ~chain ~block_hash in
+  http_get ~node_uri ~path ~of_yojson:Block_header.of_yojson
--- a/src/tezos_rpc/fetch_block_header.mli
+++ b/src/tezos_rpc/fetch_block_header.mli
@@ -1,0 +1,8 @@
+open Tezos
+
+type response = Block_header.t
+val execute :
+  node_uri:Uri.t ->
+  chain:Chain_id.t option ->
+  block_hash:Block_hash.t option ->
+  (response, Error.error) result Lwt.t
--- a/src/tezos_rpc/tezos_rpc.mli
+++ b/src/tezos_rpc/tezos_rpc.mli
@@ -1,6 +1,7 @@
 module Error = Error
 module Inject_operations = Inject_operations
 module Fetch_block_operations = Fetch_block_operations
+module Fetch_block_header = Fetch_block_header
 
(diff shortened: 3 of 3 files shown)


src/tezos_rpc/fetch_block_header.ml
```diff
@@ -0,0 +1,23 @@
...
+open Http
+
+type response = Block_header.t
+
+let path ~chain ~block_hash =
+  (* TODO: I don't like this Format.sprintf *)
+  Format.sprintf "/chains/%s/blocks/%s/header" chain block_hash
+
+let execute ~node_uri ~chain ~block_hash =
+  let chain =
```
gsebil08, line 11: In Haskell, this would lead to `shadowing` how is it possible in OCaml? Just because of the `~`?
At first sight, I got confused, is there any OCaml (or Deku) convention to avoid this kind of misleading?
I am used to do something like
```ocaml
  let chain' =
    match chain with
    | Some chain -> Chain_id.to_string chain
    | None -> "main" in
```
Moreover, I am really curious to understand how:
```ocaml
let path = path ~chain ~block_hash in
```
is able to get this `chain` and not the `~chain` provided line 10.

Would you mind help me figure out how this is working? 😄

src/tezos_rpc/fetch_block_header.ml
```diff
@@ -0,0 +1,23 @@
...
+    match chain with
+    | Some chain -> Chain_id.to_string chain
```
gsebil08, line 13: Stupid question, does this "business" belongs to this function? Why don't change `chain_id.t` to a required field, and give the responsibility of giving `main` to the caller?
Please note, once again, I have no clue of the business behind this 😅

src/tezos_rpc/fetch_block_header.ml
EduardoRFS: OCaml values are not recursive by default, unlike Haskell(and most languages), so you can freely shadow variables.

It's the same as the following in Haskell.
```haskell
let x = 1 in
(\x -> print x) (x + 1)
```

Types are recursive by default but you can do non recursive types using `type nonrec t = t`.

src/tezos_rpc/fetch_block_header.ml
EduardoRFS: Chain_id.t cannot be main, this is a special property of this endpoint, it actually accepts `chain_id | "main"`, we could make a variant for that, and I'm still thinking over it, but the easier solution was to do `chain_id option`

aguillon (approved):

src/tezos_rpc/fetch_block_header.ml
gsebil08: Thank you for the explanation 🙏!

src/tezos_rpc/fetch_block_header.ml
gsebil08: Either I don't understand your answer, or you didn't understand my question 😅.
Here, in `execute` function, we do some basic `match` condition on the `~chain`. If `execute` had been called without `chain` (since it's optional) it will use `main`. Right?

My question is:
Should this function `execute` be responsible of this? In fine, it will be exactly the same. But having `execute ~node_uri chain ~block_hash` can be totally different of having `execute ~node_uri ~chain ~block_hash`.

In my mind, I  want, as fair as possible, my functions to be "functionally scoped". By that, I mean I don't want them to calculate things they are not responsible for.

gsebil08: Finally got it! Sorry...

PR #449, opened 2022-02-23
chore: lock dune version to 2.9.3

EduardoRFS:
## Problem

Dune 3.0 stops depending on ocamlfind, which makes bigarray not available, and that breaks our pipeline. That can be seen at the docker CI.

## Solution

For now I'm locking dune to 2.9.3.


package.json
```diff
@@ -20,7 +20,7 @@
   },
   "dependencies": {
     "ocaml": "4.12.",
-    "@opam/dune": "*",
+    "@opam/dune": "~2.9.3",
```
gsebil08, line 23: What about opening an issue to track this trouble? (with a dedicated label like `dependencies` or even `technical debt`)
From my point of view, this is bad habit to lock such a dependency. But this is understandable since it's a major release (2.9.2 to 3.X.X).
What do you think?

anchpop (approved):

ulrikstrid:
Another solution would be to depend on the newly released `mmap 1.2.0`: https://github.com/ocaml/opam-repository/pull/20791

package.json
EduardoRFS: #451

PR #456, opened 2022-02-24
tests: manual tests for Tezos_rpc

EduardoRFS:
## Depends

- [x] #448

## Problem

Currently Tezos_rpc has no tests as we don't have the testing infrastructure required.

## Solution

This PR doesn't really solve the problem, but it helps by adding some manual tests.

### DIFF
--- a/tests/test_tezos_rpc.ml
+++ b/tests/test_tezos_rpc.ml
@@ -1,0 +1,36 @@
+open Helpers
+open Crypto
+open Tezos
+open Tezos_rpc
+
+let failwith_err err =
+  let failwith s = Format.kasprintf failwith s in
+  match err with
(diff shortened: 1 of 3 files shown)


tests/test_tezos_rpc.ml
```diff
@@ -0,0 +1,78 @@
...
+  let%await result = execute ~node_uri ~chain:None ~block_hash in
+  (match result with
+  | Ok header ->
+    Format.printf "header.hash: %s, header.level: %ld\n%!"
+      (Block_hash.to_string header.hash)
+      header.level
+  | Error err -> failwith_err err);
+  Lwt.return_unit
+
+let tests =
```
gsebil08, line 72: ?

Issue #440, opened 2022-02-22
New Networking Layer


rosalogia:
We would like to improve Deku's networking layer such that it is simpler, more reliable, and maintains performance.

Currently a major issue with the networking layer is that if we have some peers, A, B, and C, such that:

* A and C are peers
* C and B are peers
* A and B are not peers

and A broadcasts a message, C will receive it, but B will not. This is incorrect behaviour. The networking layer should be reworked such that so long as there is a path between A and B, regardless of whether or not A and B are peers, any message broadcasted by A should be received by B and vice versa. At the same time, this should be done in a way that avoids sending O(n^2) messages.

In the process of achieving this, there are various sub-goals that will help generally improve the quality of the networking layer, including:

- [ ] #441 
- [x] #442
- [x] #443

d4hines:
We fixed the bugs in connectivity but still have O(n^2) communication. Closing this issue - we will address improving gossip separately.

PR #462, opened 2022-02-28
WIP networking(switchFromHttpToUdp

gsebil08:
## Related

- #440


src/bin/deku_node.ml
```diff
@@ -113,32 +113,74 @@ let handle_ticket_balance =
...
+          handle_receive_consensus_operation;
+          handle_withdraw_proof;
+          handle_ticket_balance;
+          handle_trusted_validators_membership; *)
+let handle_message buffer_size = failwith "undefined"
+
+(* udp_node should definitively be in a dedicated module *)
+let udp_node folder =
+  (* those first lines are copied/pasted from previous `node` function. I think they are in charge of getting the current node state from the files in folder data/ *)
   let node = Node_state.get_initial_state ~folder |> Lwt_main.run in
```
gsebil08, line 152: Quite not sure about these lines. They are important because I think it is the current state of the Node. But how can I ensure this is working? Any help would be welcome.

src/bin/deku_node.ml
```diff
@@ -113,32 +113,74 @@ let handle_ticket_balance =
...
+          handle_block_by_hash;
+          handle_protocol_snapshot;
+          handle_request_nonce;
+          handle_register_uri;
+          handle_receive_user_operation_gossip;
+          handle_receive_consensus_operation;
+          handle_withdraw_proof;
+          handle_ticket_balance;
+          handle_trusted_validators_membership; *)
+let handle_message buffer_size = failwith "undefined"
```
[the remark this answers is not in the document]
rosalogia: I think we can directly encode a proper ADT in Bin_prot; if we wanted to do something like e.g.
```ocaml
type message =
    | Block_level of int
    | Block_and_signature of ...
    | Signature of ...
    | ...
    | Trusted_validators_membership of ...
```
then we could just put deriving Bin_prot or whatever below it and that way we don't need to worry too much about low-level payload representation.

src/bin/deku_node.ml
rosalogia: Are you wondering _if_ these lines work or _how_ these lines work? In either case, I think looking at the implementation and compiling some questions for Eduardo or someone else knowledgeable on this part of the codebase would be productive.

src/bin/deku_node.ml
```diff
@@ -113,32 +113,74 @@ let handle_ticket_balance =
...
-         handle_receive_consensus_operation;
-         handle_withdraw_proof;
-         handle_ticket_balance;
-         handle_trusted_validators_membership;
-       ]
-  @@ Dream.not_found
+  let rec server () =
+    let open Lwt_unix in
+    (* find the good buffer size *)
+    let buffer_size = Bytes.create 100 in
```
rosalogia, line 161: I think we do not have to choose a buffer size as you've done here, there is a better way to use Bin_prot to avoid this.

src/bin/deku_node.ml
```diff
@@ -113,32 +113,74 @@ let handle_ticket_balance =
...
+    let f_descr = create_socket 3001 in
+    let socket_address = getsockname f_descr in
+    match socket_address with
+    | ADDR_UNIX somethingBis -> failwith "undefined"
+    | ADDR_INET (inet_address, file_perm) ->
```
rosalogia, line 166: I guess you used `file_perm` here because your editor told you this was the type. `file_perm` is just `int`; file permissions just happen to be represented as ints. Here, the second element of the tuple is the port, so we should call it as such I suppose :smile:.

src/bin/deku_node.ml
```diff
@@ -113,32 +113,74 @@ let handle_ticket_balance =
...
+      let _ = Printf.sprintf "Waiting on port: %d\n" file_perm in
```
rosalogia, line 167: You want `Printf.printf` here, otherwise it will just return a string that does not get used since you did `let _`. In cases where you want to do an action that returns unit `()`, you can do something like
```ocaml
Printf.printf "Waiting on port: %d\n" port;
```
that is, you can use a semi-colon instead of let-matching on nothing/unit.

src/bin/deku_node.ml
```diff
@@ -113,32 +113,74 @@ let handle_ticket_balance =
...
+      let%lwt _, client_addr = recvfrom f_descr buffer_size 0 100 [] in
+      let%lwt _ = bind f_descr socket_address in
+      let _ =
+        Printf.sprintf "Bound to address: %s\n" @@ addr_to_string socket_address
```
rosalogia, line 171: See above comment about printf, same for line 174

src/bin/deku_node.ml
```diff
@@ -113,32 +113,74 @@ let handle_ticket_balance =
...
-         handle_request_nonce;
-         handle_register_uri;
-         handle_receive_user_operation_gossip;
-         handle_receive_consensus_operation;
-         handle_withdraw_proof;
-         handle_ticket_balance;
-         handle_trusted_validators_membership;
-       ]
-  @@ Dream.not_found
+  let rec server () =
```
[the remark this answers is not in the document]
rosalogia: Seems like it is mostly correct, but I guess it's still not clear how to integrate it into what we already have here. We'll figure this out :smile:.

gsebil08:
Will be done in https://github.com/marigold-dev/pollinate as an external and agnostic library

PR #463, opened 2022-02-28
Implement Lambda VM

EduardoRFS:
## Problem

As discussed in the past, Deku needs a VM to execute Ligo smart contracts.

## Solution

As we're short time on time, unlike what have been discussed in the past, I decided to go with an untyped Lambda VM that can be used to compile both Ligo and OCaml easily. The goal is that in the future this machine could be compiled to a lower level VM.

It provides an strict lambda calculus supporting tail call optimization, combined with int64 primitives, branching using an if C-like and pairs. This allows to describe primitive recursion without stack overflow, pattern matching and complex structures such as tuples, records, natural numbers and etc...

One very important property of this machine is that all the behavior should be defined, including the failure modes, so that in the future when compiling it to assembly all contracts will exhibit the same behavior, that likely will incur in some performance loss in those early contracts when optimized, but we end up having a clear migration path.

## Related

- #163

### DIFF
--- a/src/lambda_vm/ast.mli
+++ b/src/lambda_vm/ast.mli
@@ -1,0 +1,50 @@
+type ident = string
+
+type prim =
+  | Neg
+  | Add
+  | Sub
+  | Mul
+  | Div
+  | Rem
+  | Land
+  | Lor
+  | Lxor
+  | Lsl
+  | Lsr
+  | Asr
+  | Fst
+  | Snd
+
+type expr =
+  (* calculus *)
+  | Var   of ident
+  | Lam   of ident * expr
+  | App   of {
+      funct : expr;
+      arg : expr;
+    }
+  (* prims *)
+  | Const of int64
+  | Prim  of prim
+  (* branching *)
+  | If    of {
+      (* predicate <> 0 ? consequent : alternative *)
+      predicate : expr;
+      consequent : expr;
+      alternative : expr;
+    }
+  (* memory *)
+  | Pair  of {
+      first : expr;
+      second : expr;
+    }
+
+type value =
+  | Int64 of int64
+  | Pair  of value * value
+
+type script = {
+  param : ident;
+  code : expr;
+}
--- a/src/lambda_vm/checks.ml
+++ b/src/lambda_vm/checks.ml
@@ -1,0 +1,23 @@
+exception Out_of_stack
+exception Out_of_gas
+
+(* WHY: This is used to avoid stack overflows.
+
+     In the future if OCaml stack overflow detection becomes reliable then
+   we can remove it and let it overflow, but for now, manual bound checking.
+
+     About this number, OCaml currently only uses the stack for function calls
+   which means each entry is a word and maybe on ARM64 two words,
+   in 64bits systems that is 8-16 bytes. The system that has the smallest stack
+   is macOS and iOS ARM64 where the secondary stack is about 512kb.
+
+     So by assuming the worst case scenario, that allows us up to 32768, minus
+   the stack used by Deku itself. To be conservative as we may have deep
+   primitives in the future, I decided to go with two thirds of this, and
+   rounding down, which means 20k.
+*)
+(* TODO: verify properly all this hypothesis and add references to it *)
+let max_stack_depth = 20000
+
+let check_stack ~stack = if stack <= 0 then raise Out_of_stack
+let check_gas gas = if Gas.is_empty gas then raise Out_of_gas
--- a/src/lambda_vm/compiler.ml
+++ b/src/lambda_vm/compiler.ml
@@ -1,0 +1,129 @@
+open Ast
+open Ir
+open Checks
+
+module String_map = Map.Make (String)
+
+type error = (* user program bugs *)
+  | Undefined_variable
+exception Error of error
+let raise error = raise (Error error)
+
+let compile_prim prim =
+  match prim with
+  | Neg -> P_neg
+  | Add -> P_add
+  | Sub -> P_sub
+  | Mul -> P_mul
+  | Div -> P_div
+  | Rem -> P_rem
+  | Land -> P_land
+  | Lor -> P_lor
+  | Lxor -> P_lxor
+  | Lsl -> P_lsl
+  | Lsr -> P_lsr
+  | Asr -> P_asr
+  | Fst -> P_fst
+  | Snd -> P_snd
+
+module Vars = Map_with_cardinality.Make (String)
+
+let burn_gas gas vars code =
+  (match code with
+  | Var _
+  | Lam _ ->
+    let cardinality = Vars.cardinal vars in
+    Gas.burn_log2 gas ~cardinality
+  | App _
+  | Const _
+  | Prim _
+  | If _
+  | Pair _ ->
+    Gas.burn_constant gas);
+  check_gas gas
+
+let rec compile_expr ~stack gas next_ident vars code =
+  let stack = stack - 1 in
+
+  check_stack ~stack;
+  burn_gas gas vars code;
+
+  let compile_expr vars code = compile_expr ~stack gas next_ident vars code in
+
+  match code with
+  | Var var -> (
+    match Vars.find var vars with
+    | Some ident -> E_var ident
+    | None -> raise Undefined_variable)
+  | Lam (var, body) ->
+    let ident =
+      let ident = !next_ident in
+      next_ident := Ident.next ident;
+      ident in
+    let vars = Vars.add var ident vars in
+    let body = compile_expr vars body in
+    E_lam (ident, body)
+  | App { funct; arg } ->
+    let funct = compile_expr vars funct in
+    let arg = compile_expr vars arg in
+    E_app { funct; arg }
+  (* prims *)
+  | Const value -> E_const value
+  | Prim prim ->
+    let prim = compile_prim prim in
+    E_prim prim
+  (* branching *)
+  | If { predicate; consequent; alternative } ->
+    let predicate = compile_expr vars predicate in
+    let consequent = compile_expr vars consequent in
+    let alternative = compile_expr vars alternative in
+    E_if { predicate; consequent; alternative }
+  (* memory *)
+  | Pair { first; second } ->
+    let first = compile_expr vars first in
+    let second = compile_expr vars second in
+    E_pair { first; second }
+
+let compile_expr gas next_ident vars code =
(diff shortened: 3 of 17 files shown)


src/lambda_vm/ident.ml
```diff
@@ -0,0 +1,6 @@
+(* TODO: should I care about overflowing this? *)
+type t = int
+
+let compare = Int.compare
+let initial = 0
+let next t = t
```
Zett98, line 6: Shouldn't this be succ t?

Zett98 (approved):
Otherwise lgtm, I hope I didn't miss anything critical

src/lambda_vm/ast.mli
```diff
@@ -0,0 +1,50 @@
...
+  | Or
+  | Xor
+  | Lsl
+  | Lsr
+  | Asr
+
+type expr =
+  (* calculus *)
+  | Var   of ident
+  | Lam   of ident * expr
```
anchpop, line 20: why not De Bruijn?

src/lambda_vm/interpreter.ml
```diff
@@ -0,0 +1,192 @@
...
+  let eval_call env code = eval ~stack:(stack - 1) gas env code in
+  let eval_jump env code = eval ~stack gas env code in
+
+  check_stack ~stack;
+  burn_gas gas env code;
+
+  match code with
+  | E_var var -> (
+    (* TODO: gas cost for this *)
+    (* TODO: shuold this be a hashtbl? Or an array? *)
```
anchpop, line 126:
```suggestion
    (* TODO: should this be a hashtbl? Or an array? *)
```

anchpop, line 126: to arrayify this you will need to convert to de bruijn indices, a hash table sounds bad bc functional insertions are slow

anchpop (approved):

renatoalencar (commented):
I am not really sure about it, I need to make a few checks. But I've left some trivial comments about the semantics of the AST.

I also think that it would be help to have one or two example programs written by you to understand your vision about how they would be encoded.

src/lambda_vm/checks.ml
```diff
@@ -0,0 +1,8 @@
+exception Out_of_stack
+exception Out_of_gas
+
+(* TODO: reasonable number here *)
+let max_stack_depth = 1000
```
gsebil08, line 5: What could be `reasonable number`?
Is there any way to calculate it?

src/lambda_vm/compiler.ml
```diff
@@ -0,0 +1,134 @@
...
+    let prim = compile_prim prim in
+    E_prim prim
+  (* branching *)
+  | If { condition; then_; else_ } ->
+    let condition = compile_expr vars condition in
+    let then_ = compile_expr vars then_ in
+    let else_ = compile_expr vars else_ in
+    E_if { condition; then_; else_ }
+  (* memory *)
+  | Pair { left; right } ->
```
gsebil08, line 82: In Haskell, `left` and `right` are related to the `Either` monad. I think `Either` also exist on OCaml (sadly not a Monad), aren't `left` and `right` confusing?
I feel like `Pair { left; right }` should be something like `Pair { first; second }` what do you think?

src/lambda_vm/interpreter.ml
```diff
@@ -0,0 +1,192 @@
...
+  (* user program bugs *)
+  | Value_is_not_pair
+  | Value_is_not_int64
+  | Value_is_not_function
+  | Value_is_not_zero
+
+exception Error of error
+let raise error = raise (Error error)
+
+module Pattern : sig
```
gsebil08, line 17: OCaml newbie question: Why is this part not in the `.mli` file?

gsebil08 (commented):
I am not legitimate to review the goal of this pull request 😕. Hence, I asked newbie questions 👶

renatoalencar:
I also noticed you didn't exposed the modules. Doesn't dune require you to expose a module like `Lambda_vm`, if you'd like to use it externally?

renatoalencar:
How are you currently running and testing this? utop or other?

src/lambda_vm/ast.mli
```diff
@@ -0,0 +1,50 @@
...
+    }
+  (* prims *)
+  | Const of int64
+  | Prim  of prim
+  (* branching *)
+  | If    of {
+      (* condition <> 0 ? then_ : else_ *)
+      condition : expr;
+      then_ : expr;
+      else_ : expr;
```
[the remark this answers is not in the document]
EduardoRFS: Made the change, while `predicate` is quite reasonable, I don't like `consequent` and `alternative`, but if no one else opposes to it, we will be following the SICP tradition.

src/lambda_vm/checks.ml
EduardoRFS: I updated this with the explanation, but there is still some work to be done verifying my hypothesis, because of that I was very conservative.

src/lambda_vm/interpreter.ml
EduardoRFS: Because this is not used externally, it's an internal module of this module.

src/lambda_vm/compiler.ml
EduardoRFS: Yes, those are better names, and this is correct. Especially because there is `E_fst` and `E_snd`

src/lambda_vm/ast.mli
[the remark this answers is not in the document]
renatoalencar: All hail SICP!!!

src/lambda_vm/ast.mli
```diff
@@ -0,0 +1,50 @@
...
+  | Pair  of {
+      left : expr;
+      right : expr;
+    }
+  | Fst   of expr
+  | Snd   of expr
+
+type value =
+  | Int64 of int64
+  | Pair  of value * value
```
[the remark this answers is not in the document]
renatoalencar: Yeah, that makes sense.

src/lambda_vm/interpreter.ml
```diff
@@ -0,0 +1,192 @@
...
+  storage : Ir.value;
+  operations : unit;
+}
+
+let execute gas ~arg script =
+  let { param; code } = script in
+  let env = Env.add param arg Env.empty in
+  let output = eval gas env code in
+  let storage, (operations, ()) = Pattern.(parse output script_result) in
+  { storage; operations }
```
renatoalencar, line 188: If `storage` is the output of the execution, and it is a `Ir.value` then `Ir.value` should be exposed and accessible externally so the client code can interpret the results. I also think that the results shouldn't be bound to a specific pattern on the output of the script, one could have different ways of running it (Tezos has offchain views for example).

I like the idea of validating the output against a specific pattern, but that would be better if the user could define it, but that's just my point of view.

renatoalencar:
I've prepared two cases, a increment and decrement program that fails to produce the right output. I'm attaching the patch with the examples and all the setup needed to run the tests. Is quite simple, but it gives a good idea of what we should be changing.

https://gist.github.com/renatoalencar/12c526d5c5d873fe1292e310a6cca0d2

src/lambda_vm/interpreter.ml
EduardoRFS: Then how would we know where is the operations? The result is the storage to be stored and the operations to be triggered, to do that it needs to be an specific pattern, Tezos also does that.

EduardoRFS:
@renatoalencar yes that is the expected behavior, as written at https://github.com/marigold-dev/deku/pull/463/files#diff-b67bbd45b93be23ec522ac7af93288c5c7af9693c4efeec9551fddfa0b5fec15R86

You applied a single parameter, so nothing will happen.

src/lambda_vm/dune
```diff
@@ -0,0 +1,3 @@
+(library
+ (name lambda_vm)
+ (modules_without_implementation ir ast))
```
anchpop, line 3: why not make them `.ml` files instead of putting everything in the mli file?

anchpop: edu says it's just principle of least power

src/lambda_vm/compiler.ml
```diff
@@ -0,0 +1,134 @@
...
+
+  match code with
+  | Var var -> (
+    match Vars.find var vars with
+    | Some ident -> E_var ident
+    | None -> raise Undefined_variable)
+  | Lam (var, body) ->
+    let ident =
+      let ident = !next_ident in
+      next_ident := Ident.next ident;
```
lykimq, line 61: Why not just write: 
`let next_ident := Ident.next !next_ident`?

src/lambda_vm/compiler.ml
```diff
@@ -0,0 +1,134 @@
...
+      ident in
+    let vars = Vars.add var ident vars in
+    let body = compile_expr vars body in
+    E_lam (ident, body)
```
lykimq, line 65: is `E_lam (vars, body)`?

lykimq, line 65: Why return ident instead of  `E_lam (vars, body)`?

src/lambda_vm/gas.ml
```diff
@@ -0,0 +1,19 @@
+(* TODO: satured arithmetic *)
+type t = int ref
+
+let constant_burn = 100
```
lykimq, line 4: Do we have a file to declare constant parameters?

src/lambda_vm/gas.mli
```diff
@@ -0,0 +1,8 @@
+type t
+
+val make : initial_gas:int -> t
+
+val is_empty : t -> bool
+
+val burn_constant : t -> unit
```
lykimq, line 7: Can you add doc-strings for instance, in which case we use burn_constant, and burn_log2?

src/lambda_vm/compiler.ml
EduardoRFS: I don't think that's valid OCaml

src/lambda_vm/compiler.ml
EduardoRFS: Because a Lambda is a param + body.

src/lambda_vm/ast.mli
EduardoRFS: As we discussed in private, maybe in the future we should do De Bruijn for the IR, but it makes easier for developers to reason on top of the version that contains the names.

PR #464, opened 2022-03-01
sandbox(fixDockerCompose): add a dedicated name for each container

gsebil08:
## Problem

`docker-compose` default naming convention isn't use in `sandbox.sh`
```shell script
$ docker-compose up -d
Creating deku_flextesa_1 ... done
Creating deku_postgres_1 ... done
Creating deku_elastic_1 ... done
Creating deku_api_1      ... done
Creating deku_metrics_1  ... done
Creating sandbox-gui   ... done
Creating deku_indexer_1  ... done
```
See the `_1` suffix for every container

While in `sandbox.sh`:
```shell script
docker exec -it deku-flextesa-1 tezos-client "$@"
```
Hence, this leads to:
```shell script
$ ./sandbox.sh
=========== Configuring Tezos client ===========
Error: No such container: deku-flextesa-1
```

## Solution

Set a `container_name` for every container, with the same naming convention.

## Related

- #422 #428

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ yarn global add esy
 Alternatively, you can use the `nix` package manager with
 [flakes enabled](https://nixos.wiki/wiki/Flakes#Installing_flakes) to enter an
 environment preloaded with the correct dependencies, including `esy`:
+
 ```
 nix --experimental-features "nix-command flakes" run
 ```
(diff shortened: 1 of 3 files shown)


EduardoRFS (approved):
konata saying good job

PR #465, opened 2022-03-01
chore: move .ocamlformat to root

EduardoRFS:
## Problem

We have multiple `.ocamlformat` one at `src/` and another one at `tests/`, it would be better to have a single one.

## Solution

Here I move the `.ocamlformat` file to the root of the project and also add an `.ocamlformat-ignore` to prevent formatting files inside of the `esy.lock`.


.ocamlformat-ignore
```diff
@@ -0,0 +1,5 @@
+esy.lock/*
+esy.lock/*/*
+esy.lock/*/*/*
+esy.lock/*/*/*/*
+esy.lock/*/*/*/*/*
```
EduardoRFS, line 5: yes, .ocamlformat doesn't support **

Zett98 (approved):

.ocamlformat-ignore
Zett98: Lmao

gsebil08: F**k I was too late 🤣 
Thanks 😉

gsebil08 (approved):

d4hines (approved):

PR #466, opened 2022-03-01
chore: bump OCaml 5 to trunk

EduardoRFS:
## Problem

To add Nix support we need to be using compatible versions of OCaml, but currently our OCaml 5 version is quite outdated and OCaml did a couple of breaking changes since this version.

## Solution

Here I bumped the compiler and fixed the dependencies that presented problems, the resolution tower is growing but that is only temporary, as JaneStreet packages `v0.15.0` and OCaml 4.14 is released we can start to reduce it.


src/tezos_rpc/http.ml
```diff
@@ -58,9 +58,9 @@ let http_post_data_encoding ~node_uri ~path ~of_yojson ~data =
    resumed when `feed` is called, `feed` is then driven by an string Lwt_stream.
 *)
 let make_lazy_lexbuf read =
-  let open EffectHandlers in
+  let open Effect in
   let module M = struct
-    type _ eff += Poll_data : unit eff
+    type _ Effect.t += Poll_data : unit Effect.t
```
anchpop, line 63: btw where can I find documentation for the Effect stuff?

anchpop (approved):

src/tezos_rpc/http.ml
EduardoRFS: Not sure, I read about it here https://github.com/ocaml/ocaml/blob/trunk/stdlib/effect.mli

PR #467, opened 2022-03-02
chore: fix utop

EduardoRFS:
## Problem

On #446 utop was broken because of the OCaml compiler changing.

## Solution

`lwt` and `ocplib-endian` resolutions are needed to properly use utop.

PR #468, opened 2022-03-03
chore: dune 3

### LINKED ISSUE #451: Using Dune 3
At #449 we locked Dune to 2.9.3 as Dune 3 is a breaking change and the ecossystem is still not adapted, this is a temporary solution and should be fixed in the future.

EduardoRFS:
## Problem

Currently we're not on dune 3 due to the breaking changes that happened, this is non ideal.

## Solution

After #467 there is only a couple dependencies missing, this PR address it.

## Related

- Closes #451


d4hines (approved):
Tested. LGTM

PR #469, opened 2022-03-03
Rename package.json to esy.json

ulrikstrid:
## Problem

The nix setup (#438) doesn't use esy for anything but we have to use the node packages until the OCaml solution is implemented.
Therefore we need to claim the `package.json` name for node dependencies.

## Solution

To make the nix PR smoother I propose we just change the name of `package.json`to the equally valid `esy.json` and then we can introduce a new `package.json` in the nix PR.

## Related

- #438


d4hines (approved):
LGTM

gsebil08 (approved):
LGTM! 👍

PR #470, opened 2022-03-03
Simple end-to-end smoke test

d4hines:
## Depends

- [x] #486 

## Problem

Deku's tezos integration is easy to break and not notice.

## Solution

This PR adds a simple e2e smoke test. It does the following:
- runs a deku cluster of 3 nodes for 30 seconds, then kills them.
- checks that a state root hash was published with a block height greater than 20
Control is then passed to a small OCaml program that uses the protocol library to unmarshal each node's `state.bin` and check the following:
- each node's block height is exactly 30 (yes, this is abuses internal knowledge about the block production rate)
- each node's current state root hash is the one posted to Tezos.

These checks are trivial and won't catch any subtle errors in logic, but they're a decent smoke test.

 Currently the test needs to be run manually with `./sandbox.sh smoke-test`. I'm talking to infra now about how to have this as a github action.

Code is duplicated between start.sh and e2e_smoke_test.sh. Not sure the best way to handle that - we could refactor the scripts to share code, but I get sad when I need to start creating bash libraries for my project.

## Related
- #207

### DIFF
--- a/sandbox.sh
+++ b/sandbox.sh
@@ -289,6 +289,10 @@ start)
   wait_for_servers
   ;;
 smoke-test)
+  if ! command -v jq; then
+    echo "Command jq not found, but it is required to verify the result of the test."
+    exit 1
+  fi
   start_deku_cluster
   seconds=35
   sleep $seconds
--- a/sandbox.sh
+++ b/sandbox.sh
@@ -235,6 +235,25 @@ wait_for_servers() {
   done
 }
 
+assert_deku_state() {
+  contract=$(< "$data_directory/0/tezos.json" jq '.consensus_contract' | xargs)
+  storage=$(curl "$RPC_NODE/chains/main/blocks/head/context/contracts/$contract/storage")
+  current_state_hash=$(echo "$storage" | jq '.args[0].args[0].args[2].bytes' | xargs)
+  current_block_height=$(echo "$storage" | jq '.args[0].args[0].args[0].args[1].int' | xargs)
+
+  echo "The current block height is" "$current_block_height"
+
+  # Check that a state root hash was published recently
+  if [ "$current_block_height" -lt 20 ]; then
+    echo "Error: no recent state root hash update found. Exiting."
+    exit 1
+  fi
+
+  for i in "${VALIDATORS[@]}"; do
+    asserter "$data_directory/$i" "$current_state_hash" "$@"
+  done
+}
+
 help() {
   # FIXME: fix these docs
   echo "$0 automates deployment of a Tezos testnet node and setup of a Deku cluster."
@@ -250,6 +269,8 @@ help() {
   echo "  Starts a Deku cluster configured with this script."
   echo "tear-down"
   echo "  Stops the Tezos node and destroys the Deku state"
+  echo "smoke-test"
+  echo "  Starts a Deku cluster and performs some simple checks that its working."
 }
 
 message "Running in $mode mode"
@@ -266,6 +287,13 @@ setup)
 start)
   start_deku_cluster
   wait_for_servers
+  ;;
+smoke-test)
+  start_deku_cluster
+  seconds=35
+  sleep $seconds
+  killall deku-node
+  assert_deku_state $seconds
   ;;
 tear-down)
   tear-down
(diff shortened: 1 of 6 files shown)


tests/e2e/asserter.ml
```diff
@@ -0,0 +1,44 @@
...
+  assert (protocol_state.block_height >= expected_block_height);
+  let actual_state_root_hash =
+    protocol_state.state_root_hash |> BLAKE2B.to_string in
+  Format.printf "Expected state root hash: %s. Actual state root hash: %s\n%!"
+    expected_state_root_hash actual_state_root_hash;
+  assert (actual_state_root_hash = expected_state_root_hash);
+  print_endline "State looks good 👍"
+(* TODO: add more assertions *)
+
+let args =
```
lykimq, line 21: Can we move the 3 functions `folder_node, expected_state_root_hash, expected_block_height` outside, and before that open `Arg`, like this:
```
open Arg

let folder_node = ...
let expected_state_root_hash = ...
let expected_block_height = ...

let args =
let open Term in
const assert_state_correct ..
```

anchpop:
I wonder if the test should clean up ./data afterwards - currently, when running it twice it fails the second time because it starts on block 35

anchpop (approved):

sandbox.sh
```diff
@@ -189,11 +190,33 @@ start_deku_cluster() {
...
+}
+
+wait_for_servers() {
   for PID in ${SERVERS[@]}; do
     wait $PID
   done
 }
 
+assert_deku_state() {
+  contract=$(cat "$data_directory/0/tezos.json" | jq '.consensus_contract' | xargs)
```
aguillon, line 202: We don't list `jq` as a dev dependency, so I'd either add it to the README or add something as:
```
if ! command -v jq
then
    echo Command jq not found, but it is required to verify the result of the test.
    exit 1
fi
```

sandbox.sh
```diff
@@ -189,11 +190,33 @@ start_deku_cluster() {
...
+  storage=$(curl "$RPC_NODE/chains/main/blocks/head/context/contracts/$contract/storage")
```
aguillon, line 203: Shouldn't we prefer `curl --silent`?

aguillon:
> > I wonder if the test should clean up ./data afterwards - currently, when running it twice it fails the second time because it starts on block 35
> 
> I've opted to leave the operations atomic and simpler, at the cost of requiring users to understand the order they have to be run in. It's the same with `./sandbox.sh start` - you can't run it before running `./sandbox.sh setup`. Likewise, you can't run `./sandbox.sh smoke-test` without first running `./sandbox.sh setup`, whether it's first or the second time.
> 
> Does this seem sensible? I feel like the alternative (Make-style dependency graphs) could be arbitrarily complex to implement.

I'm not sure what @anchpop was referring to because I did not observe this behaviour (I only tested last version) so I'm not sure this is related but: wouldn't another direction be to make this test more robust to a preexisting state? Say you already ran deku for a few blocks and you run the smoke test afterwards, then testing that it reached `35` does not make sense anymore (as it could have started here anyway), does it? Draft here: https://github.com/marigold-dev/deku/tree/e2e-test-relative-height

sandbox.sh
d4hines: addressed in latest commit

sandbox.sh
d4hines: Should we? I think you're the superier bash scripter here. I've addressed it in the latest commit.

sandbox.sh
aguillon: For nix only, no? By the way, how did it work before with Nix? I believe you have to list all the dependencies with nix files, don't you?

d4hines: So, when using `nix develop`, the default is for your shell environment to "come with you" - you can use any tools you already had installed, in addition to the tools nix adds (this can be overridden with the `--ignore-environment` flag).

Being used in the smoke test, `jq` is a dev dependency, not needed for consumers of the Deku who want to merely build and run it. Hence I don't think it needs a spot in the README. For Deku devs, I think the error message you added is sufficient.

anchpop:
I agree with @aguillon, it shouldn't block this PR but it would be nice to tear-down the node state before running the smoke test, or run from a known-good state.

PR #474, opened 2022-03-07
exclude docker folder from docker

d4hines:
## Problem

On every change to the dockerfile, the change is sent to the buid context, triggering deku to be built from scratch every time.

## Solution

Ignore ./docker in .dockerignore


gsebil08 (approved):
LGTM

PR #475, opened 2022-03-07
chore: fix dune 3

EduardoRFS:
## Problem

As documented at ocaml/dune#5507, dune.3.0.3 has a bug with esy when removing directories. Leading to all sorts of pain when developing.

## Solution

This PR is a temporary solution, by using the dune commit directly from ocaml/dune#5507


anchpop (approved):
I'm not really confident about what problem this solves or how it solves it, but it fixed the build for me (I ran `esy` and `esy dune runtest` and both worked fine)

esy.json
```diff
@@ -125,23 +123,23 @@
...
+      "override": {
         "dependencies": {
           "@opam/ppx_sexp_conv": "*"
         }
       }
     },
     "@opam/ppx_base": {
       "version": "opam:v0.14.0",
-      "override" : {
+      "override": {
```
gsebil08, line 142: I guess this is formatter-related, so the version without space is the correct one?
Do we have a JSON formatter definer somewhere? To avoid this kind of noise in future PRs?

d4hines (approved):
LGTM

esy.json
```diff
@@ -43,7 +40,8 @@
     "@opam/ppx_blob": "*",
     "@opam/secp256k1-internal": "*",
     "@opam/domainslib": "0.4.1",
-    "@opam/data-encoding": "*"
+    "@opam/data-encoding": "*",
+    "@opam/graphql": "*"
```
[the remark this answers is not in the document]
EduardoRFS: Nope, that was a mistake

esy.json
EduardoRFS: I'm just using prettier by default

esy.json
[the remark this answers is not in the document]
gsebil08: So do I, but looks like we are not all using it.

esy.json
gsebil08: So do I, but looks like we are not all using it.

gsebil08 (approved):

PR #476, opened 2022-03-07
Migrate tests to OCaml

anchpop:
## Depends

- [x] #475

## Problem

Most of the tests are currently written in Reason, but the rest of the codebase is OCaml

## Solution

Convert the tests to OCaml syntax


d4hines (approved):
LGTM :+1:

PR #477, opened 2022-03-07
Don't use double semicolon

anchpop:
## Depends

- [x] #476

## Problem

`reason_to_ocaml.sh` generates `;;` everywhere. But I think that's kind of unidiomatic - it would be better to use `let () = ` in most cases. 

## Solution

Convert every usage of `;;` to `let () = `

### DIFF
--- a/tests/test_ledger.ml
+++ b/tests/test_ledger.ml
@@ -2,182 +2,185 @@ open Setup
 open Helpers
 open Crypto
 open Core
-open Ledger;;
-describe "ledger" (fun { test; _ } ->
-    let make_ticket ?ticketer ?data () =
-      let open Tezos in
-      let ticketer =
-        match ticketer with
-        | Some ticketer -> ticketer
-        | None ->
-          let random_hash =
-            Random.generate 20
-            |> Cstruct.to_string
-            |> BLAKE2B_20.of_raw_string
-            |> Option.get in
-          Address.Originated { contract = random_hash; entrypoint = None } in
-      let data =
-        match data with
-        | Some data -> data
-        | None -> Random.generate 256 |> Cstruct.to_bytes in
-      let open Ticket_id in
-      { ticketer; data } in
-    let make_address () =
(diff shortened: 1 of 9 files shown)


tests/test_ledger.ml
```diff
@@ -0,0 +1,186 @@
+open Setup
+open Helpers
+open Crypto
+open Core
+open Ledger
+
+let () =
```
lykimq, line 7: It is a very big function. 
Could it be possible to split them into smaller functions and then each test dedicated for the kind of test that you want to test, in the end, call those test functions?

tests/test_tezos_interop.ml
```diff
@@ -0,0 +1,605 @@
...
+            ()))
+let () =
+  describe "key_hash" (fun { test; _ } ->
+      let open Key_hash in
+      let tz1 = of_key genesis_wallet in
+      let tz2 = of_key TZ2_ex.pk in
+      let tz3 = of_key TZ3_ex.pk in
+      test "to_string" (fun { expect; _ } ->
+          (expect.string (to_string tz1)).toEqual
+            "tz1LzCSmZHG3jDvqxA8SG8WqbrJ9wz5eUCLC";
```
lykimq, line 103: Do you think defining the address name for this is better? for instance (you can have a better name):
`let add_1 = "tz1LzCSmZHG3jDvqxA8SG8WqbrJ9wz5eUCLC"` 
Then when we write the test below it is easier to read the code and know what we are comparing.

lykimq:
General review note: The test functions are very big, I am in favor of splitting them into smaller functions, it is better while reading the code and debugging in case of failure.

d4hines (approved):
LGTM :+1:

Zett98 (approved):

tests/test_tezos_interop.ml
EduardoRFS: this is not related to this PR

Issue #455, opened 2022-02-24
Chore: migrate to alcotest


d4hines:
Rely depends on Reason, which isn't ready for OCaml 5 yet. We should migrate to alcotest in the meantime.

I'm open to other suggestions, but Alcotest is what I've used with Tezos, and it's fine by me.

d4hines:
Done.

PR #479, opened 2022-03-07
Testing for Lambda VM

renatoalencar:
## Depends

- [x] #463 
- [ ] #505
- [ ] #504 

## Problem

Lambda VM needs proper testing and practical examples of use.

## Solution

I tested all the base structures of the interpreter and AST, errors and primitive operations. For the basic interpreter features I've just built simple increment/decrement programs. For primitive operations I've built qcheck tests generating int64 for all the operations.

All gas value were adjusted to the bare minimum, so they won't get larger without being noticed.

### Simple expressions

I've built a few simple programs to test the basic functionality the AST provides, like applications, primitives for arithmetic operations, if expressions, lambdas and manipulating pairs.

### Compilation and execution errors

Checked for different situations that would happen in real programs, making sure the runtime checks works.

### Primitive operations

Used QCheck to make property based testing for all available primitive operations, generating random 64-bit integers and running operations on them.

### Recursion

Making sure that is possible and viable to run complex programs with loops and recursion, and that gas and limits work in order to prevent exhaustion of resources by third party executed code. Examples include factorial, Fibonacci, a counter and infinite non tail recursion.

### Gas model

Also added a few tests in order to validate an constraint gas model, the way this works is by running this examples in the limit both above and below the necessary to run out of gas and so making sure that's constant and in case it changes it is detectable by the test suite.

I believe a better way to do this would be by benchmarking the VM and comparing with the corresponding gas consumption to check if they're correlated in a way that enough gas is charged in order to run a corresponding expensive computation.

### DIFF
--- a/src/lambda_vm/ir.ml
+++ b/src/lambda_vm/ir.ml
@@ -1,5 +1,7 @@
 type ident = Ident.t
 
+let pp_ident = Ident.pp
+
 type prim =
   | P_neg
   | P_add
@@ -15,6 +17,13 @@ type prim =
   | P_asr
   | P_fst
   | P_snd
+[@@deriving show]
+
+module Env = struct
+  include Map_with_cardinality.Make (Ident)
+
+  let pp _ fmt _ = Format.fprintf fmt "<env>"
+end
 
 type expr =
   (* calculus *)
@@ -39,6 +48,7 @@ type expr =
       first : expr;
       second : expr;
     }
+[@@deriving show]
 
 type value =
   | V_int64     of int64
@@ -47,7 +57,7 @@ type value =
       second : value;
     }
   | V_closure   of {
-      env : env;
+      env : value Env.t;
       param : Ident.t;
       body : expr;
     }
@@ -55,7 +65,7 @@ type value =
       args : value list;
       prim : prim;
     }
-and env = value Map_with_cardinality.Make(Ident).t
+[@@deriving show]
 
 type script = {
   param : Ident.t;
--- a/tests/vm/dune
+++ b/tests/vm/dune
@@ -1,0 +1,5 @@
+(test
+ (name test_vm)
+ (libraries lambda_vm alcotest qcheck-alcotest)
+ (preprocess
+  (pps ppx_lambda_vm)))
--- a/tests/vm/test_errors.ml
+++ b/tests/vm/test_errors.ml
@@ -1,0 +1,117 @@
+open Lambda_vm
+
+module Testable = Vm_test.Testable
+
+let check_execution_error result expected_error =
+  let open Vm_test in
+  match result with
+  | Error (Execution_error error) ->
+    Alcotest.(check Testable.execution_error) "Same error" expected_error error
+  | Ok _ -> Alcotest.fail "Ast shouldn't execute"
+  | Error (Compilation_error error) ->
+    Alcotest.failf "%a" pp_compile_error error
+
+let check_compilation_error result expected_error =
+  let open Vm_test in
+  match result with
+  | Error (Compilation_error error) ->
+    Alcotest.(check Testable.compilation_error)
+      "Same error" expected_error error
+  | Ok _ -> Alcotest.fail "Ast shouldn't execute"
+  | Error (Execution_error error) ->
+    Alcotest.failf "%a" pp_execution_error error
+
+let test_compilation_undefined_variable () =
+  let script = [%lambda_vm.script fun _ -> x] in
+  check_compilation_error
+    (Vm_test.execute_ast 2000 (Int64 0L) script)
+    Undefined_variable
+
+let test_fst_value_is_not_pair () =
+  let script = [%lambda_vm.script fun _ -> (fst 1L, (0L, 0L))] in
+  check_execution_error
+    (Vm_test.execute_ast 1201 (Int64 0L) script)
+    Value_is_not_pair
+
+let test_snd_value_is_not_pair () =
+  let script = [%lambda_vm.script fun _ -> (snd 1L, (0L, 0L))] in
+  check_execution_error
+    (Vm_test.execute_ast 1201 (Int64 0L) script)
+    Value_is_not_pair
+
+let test_neg_value_is_not_int64 () =
+  let script = [%lambda_vm.script fun _ -> (not (0L, 0L), (0L, 0L))] in
+  check_execution_error
+    (Vm_test.execute_ast 1601 (Int64 0L) script)
+    Value_is_not_int64
+
+let test_op2_value_is_not_int64 () =
+  let script = [%lambda_vm.script fun _ -> ((0L, 0L) + 0L, (0L, 0L))] in
+  check_execution_error
+    (Vm_test.execute_ast 2001 (Int64 0L) script)
+    Value_is_not_int64
+
+let test_if_value_is_not_int64 () =
+  let script =
+    [%lambda_vm.script fun _ -> if (0L, 0L) then 1L else (1L, (0L, 0L))] in
+  check_execution_error
+    (Vm_test.execute_ast 1601 (Int64 0L) script)
+    Value_is_not_int64
+
+let test_value_is_not_function () =
+  let script = [%lambda_vm.script fun _ -> (0L 0L, (0L, 0L))] in
+  check_execution_error
+    (Vm_test.execute_ast 1201 (Int64 0L) script)
+    Value_is_not_function
+
+let test_pattern1_value_is_not_pair () =
+  let script = [%lambda_vm.script fun _ -> (0L, 0L)] in
+  check_execution_error
+    (Vm_test.execute_ast 701 (Int64 0L) script)
+    Value_is_not_pair
+
+let test_pattern2_value_is_not_pair () =
+  let script = [%lambda_vm.script fun _ -> 0L] in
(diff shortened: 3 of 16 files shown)


anchpop:
I know it's a draft, but here are my preliminary thoughts: 

1. Not that I'm opposed to it, but I don't see the purpose of using alcotest when we're already using rely.

2. You don't need to write functions like `prim_to_string` by hand, you can use a ppx. In case it's unclear how, I made a demo for you in my [@anchpop/show-lambda-vm](https://github.com/marigold-dev/deku/commit/33155d65d448cdf65c04ec280b35ea486ef90239) branch, just look at the top commit.

Zett98:
> 1. Not that I'm opposed to it, but I don't see the purpose of using alcotest when we're already using rely.

from #455
> Rely depends on Reason, which isn't ready for OCaml 5 yet. We should migrate to alcotest in the meantime.

renatoalencar:
> > 1. Not that I'm opposed to it, but I don't see the purpose of using alcotest when we're already using rely.
> 
> from #455
> 
> > Rely depends on Reason, which isn't ready for OCaml 5 yet. We should migrate to alcotest in the meantime.

That's why I am using Alcotest @anchpop

anchpop:
> > > 1. Not that I'm opposed to it, but I don't see the purpose of using alcotest when we're already using rely.
> 
> > 
> 
> > from #455
> 
> > 
> 
> > > Rely depends on Reason, which isn't ready for OCaml 5 yet. We should migrate to alcotest in the meantime.
> 
> 
> 
> That's why I am using Alcotest @anchpop 

I see, that makes sense

renatoalencar:
> You don't need to write functions like prim_to_string by hand, you can use a ppx. In case it's unclear how, I made a demo for you in my [@anchpop/show-lambda-vm](https://github.com/marigold-dev/deku/commit/33155d65d448cdf65c04ec280b35ea486ef90239) branch, just look at the top commit.

That's great, thank you!

esy.json
```diff
@@ -16,31 +16,35 @@
...
-    "@opam/mirage-crypto-ec": "*",
-    "@opam/dream": "*",
-    "@opam/piaf": "*",
     "@opam/mrmime": "0.3.2",
-    "@opam/hex": "*",
+    "@opam/piaf": "*",
+    "@opam/ppx_blob": "*",
+    "@opam/ppx_deriving": "*",
+    "@opam/ppx_deriving_yojson": "*",
+    "@opam/ppx_expect": "github:renatoalencar/ppx_expect:ppx_expect.opam#94d2862f6df315955de9951d8c77f47085a9a876",
```
renatoalencar, line 37: I don't know why esy is messing up the `esy.json` file, but in case anyone tries to fix it I need this dependency to have this resolution.

src/lambda_vm/gas.ml
```diff
@@ -9,11 +9,24 @@ let log2_burn ~cardinality =
...
+
+let burn t amount =
+  match t with
+  | Burn t -> t := !t - amount
+  | Measure t -> t := !t + amount
+
+let current t =
+  match t with
+  | Measure t
+  | Burn t -> !t
```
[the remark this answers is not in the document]
EduardoRFS: Nope, if you want to measure it, just put a high gas cost on the reference and look at the diff since the initial after executing

src/lambda_vm/ir.ml
```diff
@@ -15,6 +17,13 @@ type prim =
   | P_asr
   | P_fst
   | P_snd
+[@@deriving show]
+
+module Env = struct
+  include Map_with_cardinality.Make(Ident)
+
+  let pp _fmt _ _ = ()
```
EduardoRFS, line 25:
```suggestion
  let pp fmt _ _ = Format.fprintf fmt "<env>"
```

tests/vm/dune
```diff
@@ -0,0 +1,34 @@
...
+ (libraries vm_test lambda_vm alcotest)
+ (preprocess
+  (pps ppx_lambda_vm)))
+
+(test
+ (name test_prim)
+ (modules test_prim)
+ (libraries vm_test lambda_vm alcotest qcheck-alcotest)
+ (preprocess
+  (pps ppx_lambda_vm)))
```
EduardoRFS, line 33: There is no need to put each module in it's own library, put everything on vm_test, it's good enough

tests/vm/test_errors.ml
```diff
@@ -0,0 +1,128 @@
...
+    let script = [%lambda_vm.script fun _ -> (snd 1L, (0L, 0L))] in
+    check_execution_error
+      (Vm_test.execute_ast 1201 (Int64 0L) script)
+      Value_is_not_pair
+
+  let test_op1_value_is_not_int64 () =
+    let script = [%lambda_vm.script fun _ -> (not (0L, 0L), (0L, 0L))] in
+    check_execution_error
+      (Vm_test.execute_ast 1601 (Int64 0L) script)
+      Value_is_not_int64
```
EduardoRFS, line 48: Some op1 will work on pairs, like `fst` and `snd` so this test name is not very good

tests/vm/test_gas.ml
```diff
@@ -0,0 +1,45 @@
...
+              0L),
+        (0L, 0L) )]
+
+let%expect_test _ =
+  let gas = Gas.measure () in
+  let _ =
+    Vm_test.compile_value_exn gas (Int64 0L) in
+  ();
+  print_int (Gas.current gas);
+  [%expect{| 100 |}]
```
EduardoRFS, line 19: In general the way that the gas model is being tested here is not very good, `(Gas.is_empty gas)` should be true, which means what is parameterized is the initial_gas not the final one.

EduardoRFS (dismissed):

src/lambda_vm/gas.ml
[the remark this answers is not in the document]
renatoalencar: Ok, that's a lot better. What do you recommend? A `current` function in order to get the value?

tests/vm/dune
renatoalencar: Oh ok, `test_gas` does need its own stanza though.

tests/vm/test_gas.ml
renatoalencar: Oh, but then I think that would be changed to a regular Alcotest testsuite instead of using ppx_expect.

tests/vm/test_recursion.ml
```diff
@@ -0,0 +1,101 @@
...
+let test_stack_limit () =
+  Alcotest.check_raises "Stack has a limit" Out_of_stack (fun () ->
+      let _ =
+        Vm_test.execute_ast_exn 71_990_801
+          (Int64 19996L) (* Bare minimum close to the limit of 20k *)
+          counter in
+      ())
+
+let infinite_recursion =
+  [%lambda_vm.script fun _ -> (fun f -> f f) (fun f -> f f + 0L)]
```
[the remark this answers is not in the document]
renatoalencar: Yeah, for this specific case I would like the stack limit to be triggered. I'll definitely add a test case for the Z combinator though.

tests/vm/dune
renatoalencar: Changed the way gas is tested base on your advice. Now it checks if gas is empty, also changed the way the check and burn gas works in order to have it empty at the and not needing an extra gas to work.

tests/vm/test_recursion.ml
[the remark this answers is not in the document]
renatoalencar: Added the z combinator and tested against the `Out_of_gas` exception

tests/vm/test_gas.ml
renatoalencar: Changed to Alcotest

renatoalencar:
@EduardoRFS I've rebase `main` into my branch. You`d need to update yours in order to remove the conflicts.

tests/vm/test_gas.ml
```diff
@@ -0,0 +1,48 @@
...
+      ( (fun f -> f f x) (fun f n ->
+            if n then
+              1L + f f (n - 1L)
+            else
+              0L),
+        (0L, 0L) )]
+
+let test_compile_value () =
+  let gas = Gas.make ~initial_gas:100 in
+  let _ = Vm_test.compile_value_exn gas (Int64 0L) in
```
Zett98, line 15: can be just 
```ocaml
 let _ = expr in 
 ```
 instead of 
```ocaml
let _ = expr in
();
 ```

File: tests/vm/test_gas.ml [hunk 9]
```diff
@@ -0,0 +1,48 @@
...
+  let x = 4096L in
+  let arg =
+    let gas = Gas.make ~initial_gas:101 in
+    Vm_test.compile_value_exn gas (Int64 x) in
+  let ir =
+    let gas = Gas.make ~initial_gas:5000 in
+    Vm_test.compile_exn gas counter in
+
+  let gas = Gas.make ~initial_gas:14747900 in
+  let _ = Vm_test.execute_exn gas arg ir in
```
Zett98, line 37: same

Zett98 (changes requested):
all good in general, but you need to split it up in separate prs: 
- addition of alcotest and qcheck 
- addition of ppx_lambda_vm
- the changes to lambda_vm and tests

Zett98 (approved):

PR #480, opened 2022-03-08
Add test to Lambda_vm

anchpop:
## Depends

- [x] #476
- [x] #477

## Problem

`Lambda_vm` has no tests

## Solution

Add a test. The test is very simple (it only tests lambda application) so more tests are definitely needed

## Related

I forgot @renatoalencar was working on this with #479, maybe this PR will be useful to him?


lykimq:
is it the same work: with https://github.com/marigold-dev/deku/pull/479/?

PR #482, opened 2022-03-08
tezos_interop: increase polling rate for Taquito

EduardoRFS:
## Problem

Currently Deku is loosing Tezos operations in some nodes, leading to nodes not agreeing on what is a valid Tezos operation and stalling the chain.

## Solution

The issue is only happening on Flextesa, due to the reduced block time and the internal polling rate of Taquito, this PR workaround it by increase the polling rate, in the future we should replace Taquito by Tezos_rpc.

### DIFF
--- a/src/tezos_interop/fetch_storage.js
+++ b/src/tezos_interop/fetch_storage.js
@@ -6,6 +6,7 @@ const { RpcClient } = require("@taquito/rpc");
 const { InMemorySigner } = require("@taquito/signer");
 
 const { TezosToolkit } = taquito;
+
 /**
  * @typedef Input
  * @type {object}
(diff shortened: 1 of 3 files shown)


src/tezos_interop/fetch_storage.js
```diff
@@ -40,6 +41,14 @@ const error = (error) =>
   const { rpc_node, contract_address, confirmation } = input();
   const client = new RpcClient(rpc_node);
   const Tezos = new TezosToolkit(rpc_node);
+  Tezos.setProvider({
+    config: {
+      shouldObservableSubscriptionRetry: true,
```
gsebil08, line 46: Since we use three times the same values, I would have extract them to commons const define somewhere relevant, in order to do something like
```javascript
# relevant place to define these const
const shouldObservableSubscriptionRetry = true;
const streamerPollingIntervalMilliseconds = 1000;

# Then within the other places:
  Tezos.setProvider({
    config: {
      shouldObservableSubscriptionRetry: shouldObservableSubscriptionRetry,
      streamerPollingIntervalMilliseconds: streamerPollingIntervalMilliseconds,
[...]
    },
  });
```
What do you think?

EduardoRFS: We have quite a lot of shared code in those files, and they're being removed in the near future, I would like to avoid adding more infrastructure for them.

gsebil08: Got it, maybe a simple comment like
> this is also defined in listen_transactions.js

to avoid any miss when we change/remove them?

EduardoRFS: done boss

Issue #481, opened 2022-03-08
Partially signed block


EduardoRFS:
A block may contain invalid operations, but not necessarily this block needs to be rejected as that adds latency to the network, another option is to allow signatures over part of a block, and only the part that has enough signatures will be included.

This will happens both in the case of unknown consensus operations or unknown Tezos operations,

PR #483, opened 2022-03-08
node: delay before including Tezos operations

EduardoRFS:
## Depends

- [x] #482

## Problem

Deku or Tezos nodes will not have clocks perfectly in sync, this means that some nodes will receive Tezos operations faster than others, if that happens to be the block producer it will include a Tezos operation not known to the others making this block to be skipped.

## Solution

The ideal solution would be #481 but that is way more complex to implement, so this PR implements an workaround,. by adding a small delay between receiving a Tezos operation and actually including it on the block.

### DIFF
--- a/src/node/building_blocks.ml
+++ b/src/node/building_blocks.ml
@@ -33,6 +33,12 @@ let is_current_producer state ~key_hash =
   Some (current_producer.address = key_hash)
 let minimum_signable_time_between_epochs = 10.0
 let maximum_signable_time_between_epochs = 20.0
+
+(** Used to add a delay between a tezos operation being confirmed,
+  needs to be bigger than the polling interval for operations *)
+let minimum_waiting_period_for_tezos_operation = 5.0
+(* TODO: this is an workaround solution, replace it by Tezos_rpc *)
+
 let block_matches_current_state_root_hash state block =
   BLAKE2B.equal block.Block.state_root_hash state.Node.protocol.state_root_hash
 let block_matches_next_state_root_hash state block =
@@ -62,7 +68,7 @@ let is_signable state block =
     match operation with
     | Protocol.Operation.Core_tezos _ ->
       List.exists
-        (fun op -> Protocol.Operation.equal op operation)
+        (fun op -> Protocol.Operation.equal op.Node.operation operation)
         state.pending_operations
     | Core_user _ -> true
     | Consensus consensus_operation -> (
@@ -102,18 +108,38 @@ let should_start_new_epoch last_state_root_update current_time =
     \    The block producer uses this function to determine when to send a\n\
     \    block with an updated state root hash.\n"]
 
+(** Can only included a tezos operation if enough time has already elapsed *)
+let can_include_tezos_operation ~current_time pending_operation =
+  current_time -. pending_operation.Node.requested_at
+  > minimum_waiting_period_for_tezos_operation
+
 let produce_block state =
+  let current_time = Unix.time () in
   let start_new_epoch =
     should_start_new_epoch state.Node.protocol.last_state_root_update
-      (Unix.time ()) in
+      current_time in
   let next_state_root_hash =
     if start_new_epoch then
       let%some snapshot = Snapshots.get_next_snapshot state.snapshots in
       Some snapshot.hash
     else
       None in
+  let operations =
+    List.filter_map
+      (fun pending_operation ->
+        let operation = pending_operation.Node.operation in
+        match operation with
+        | Operation.Core_tezos _ ->
+          if can_include_tezos_operation ~current_time pending_operation then
+            Some operation
+          else
+            None
+        | Core_user _
+        | Consensus _ ->
+          Some operation)
+      state.pending_operations in
   Block.produce ~state:state.Node.protocol ~author:state.identity.t
-    ~next_state_root_hash ~operations:state.pending_operations
+    ~next_state_root_hash ~operations
 let is_valid_block_height state block_height =
   block_height >= 1L && block_height <= state.Node.protocol.block_height
 let signatures_required state =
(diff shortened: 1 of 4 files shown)


src/node/building_blocks.ml
```diff
@@ -102,18 +108,38 @@ let should_start_new_epoch last_state_root_update current_time =
...
+        let operation = pending_operation.Node.operation in
+        match operation with
+        | Operation.Core_tezos _ ->
+          if can_included_tezos_operation ~current_time pending_operation then
+            Some operation
+          else
+            None
+        | Core_user _
+        | Consensus _ ->
+          Some operation)
```
renatoalencar, line 139: I mean, not exactly a proposal, but a question. Why not like this?

```ocaml
match operation with
| Consensus _
| Operation.Core_tezos _ when can_included_tezos_operation ~current_time pending_operation ->
   Some operation
| _ -> None
 ```

File: src/node/building_blocks.ml [hunk 2]
```diff
@@ -62,7 +68,7 @@ let is_signable state block =
     match operation with
     | Protocol.Operation.Core_tezos _ ->
       List.exists
-        (fun op -> Protocol.Operation.equal op operation)
+        (fun op -> Protocol.Operation.equal op.Node.operation operation)
```
rosalogia, line 71: Why is this change necessary?

src/node/building_blocks.ml
```diff
@@ -102,18 +108,38 @@ let should_start_new_epoch last_state_root_update current_time =
     \    The block producer uses this function to determine when to send a\n\
     \    block with an updated state root hash.\n"]
 
+(** Can only included a tezos operation if enough time has already elapsed *)
+let can_included_tezos_operation ~current_time pending_operation =
```
rosalogia, line 112: Should be named `can_include_tezos_operation`

rosalogia (approved):

src/node/building_blocks.ml
```diff
@@ -33,6 +33,12 @@ let is_current_producer state ~key_hash =
   Some (current_producer.address = key_hash)
 let minimum_signable_time_between_epochs = 10.0
 let maximum_signable_time_between_epochs = 20.0
+
+(** Used to add a delay between a tezos operation being confirmed,
+  needs to be bigger than the polling interval for operations *)
+let minimum_waiting_period_for_tezos_operation = 5.0
```
gsebil08, line 39: Is there also a maximum? Like before considering this operation is expired?

lykimq:
General question not related to this MR: why the file `building_blocks.ml` does not have the mli file?

src/node/building_blocks.ml
```diff
@@ -102,18 +108,38 @@ let should_start_new_epoch last_state_root_update current_time =
...
       Some snapshot.hash
     else
       None in
+  let operations =
+    List.filter_map
+      (fun pending_operation ->
+        let operation = pending_operation.Node.operation in
+        match operation with
+        | Operation.Core_tezos _ ->
+          if can_included_tezos_operation ~current_time pending_operation then
```
gsebil08, line 133: What about changing `can_include_tezos_operation` to return `Some operation` or `None`, and here get rid of the `if` statement?

src/node/building_blocks.ml
EduardoRFS: The difference is that in one you can read the match separated from the conditional.

In general I avoid this pattern when I'm talking about a single case, the way that I read the current code is, `if it is a Core_tezos then do this thing` while the one proposed by you is, `if it's a Core_tezos and can be included then do this thing`. 

Also in general avoid catch all `_`.

src/node/building_blocks.ml
gsebil08: Hum, almost agreed. I still feel weird to see:
```ocaml
if can_included_tezos_operation ~current_time pending_operation then
    Some operation
  else
    None
```
within this function. Any argument to help me appreciate this?

src/node/building_blocks.ml
d4hines: Answer: Eduardo changed the type by wrapping it in a record that additionally includes the timestamp marking when it was requested.

src/node/building_blocks.ml
d4hines: In one sense Tezos operations never "expire" - so long as they are included in the canonical chain. However, this raises an interesting point: what if two nodes are following different forks of Tezos temporarily?

If only a minority is following a particular fork, their blocks won't be signed but the chain will continue, no problem.

If the majority is following a particular fork that turns out to be wrong (i.e, there is a re-org), then (@EduardoRFS correct me if I'm wrong) I think what will happen is when the next block is posted to Tezos (i.e, one based on the wrong fork), Deku will also fork in a sense. There will be two histories: the old history based on the wrong fork, and a new history based on the new fork. Eventually a block producer will produce a block following the new, correct history. Assuming everyone is following the correct history (and if they're using a Tezos node that's functioning properly, they will), they will eventually all agree. When a block from this new history gets posted to Tezos, it will represent the validators collectively overwriting the history. There will be no sequence of blocks to get from the block referencing the bad fork to the block referencing the good fork, but it's OK because our proof-of-authority consensus currently allows them to rewrite history.

d4hines:
LGTM :+1: 
when you come up with a quick workaround for a deep systemic issue but its great

d4hines (approved):

src/node/building_blocks.ml
gsebil08: Thanks for the detailed answer! I appreciate, it helps me to better understand the blockchain 👶

Zett98 (approved):

gsebil08 (approved):

renatoalencar (approved):
LGTM

aguillon (approved):

PR #485, opened 2022-03-10
Add error logging to taquito scripts

d4hines:
## Problem

I've found that any errors in Taquito aren't properly
logged, making it hard to debug them.

## Solution

Just add `console.error`.

### DIFF
--- a/src/tezos_interop/fetch_storage.js
+++ b/src/tezos_interop/fetch_storage.js
@@ -1,6 +1,7 @@
 "use strict";
 
 const fs = require("fs");
+const { inspect } = require("util");
 const taquito = require("@taquito/taquito");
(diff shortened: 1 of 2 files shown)


gsebil08 (approved):
LGTM 👍

EduardoRFS:
Why are they not properly logged? This is not addressing it and will duplicate a bunch of errors.

d4hines:
As a compromise, I kept the `console.error` in NodeJS, but only send  a `{status: "error"}` obj to OCaml, omitting the error message. In  this way, the formatting and stack trace of the error message is preserved, but the information is only logged once instead of twice.

What do you think @EduardoRFS ?

d4hines:
This PR is superceded by #497 and company

PR #486, opened 2022-03-10
fix: use default confirmation polling timeout

d4hines:
## Problem

The e2e test added in #470 is failing because the confirmation polling is timing out. You can see an example here: https://github.com/marigold-dev/deku/runs/5500148802?check_suite_focus=true#step:8:61

This doesn't happen locally, at least not for me. I think it has to do with the speedo f the machine involved.

## Solution

Raise the confirmation timeout. I'm not sure if this has any bad effects. If needed, we could make it configurable through an environment variable
that we set in the CI, but I doubt an increase of 6 seconds is going to break us.

## Related

- #470

### DIFF
--- a/src/tezos_interop/fetch_storage.js
+++ b/src/tezos_interop/fetch_storage.js
@@ -47,7 +47,6 @@ const error = (error) =>
       shouldObservableSubscriptionRetry: true,
       streamerPollingIntervalMilliseconds: 1000,
       confirmationPollingIntervalSecond: 1,
-      confirmationPollingTimeoutSecond: 4,
     },
   });
   const block = await client.getBlock(); // fetches the head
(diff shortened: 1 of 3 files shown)


d4hines:
Update: I encountered the error I linked to locally as well. I'm not sure 10 seconds is enough time. Why not make it much bigger, like 50 seconds or something? Is there a drawback?

I realized this timeout was added very recently in #482 and I'm wondering if it introduced a regression.

src/tezos_interop/fetch_storage.js
```diff
@@ -47,7 +47,7 @@ const error = (error) =>
       shouldObservableSubscriptionRetry: true,
       streamerPollingIntervalMilliseconds: 1000,
       confirmationPollingIntervalSecond: 1,
-      confirmationPollingTimeoutSecond: 4,
+      confirmationPollingTimeoutSecond: 10,
```
gsebil08, line 50: It is also set to `4` within listen_transaction.js
Regarding your issue, I think you also should go for 10 in the previous file, am I right?

EduardoRFS:
This is not a fix, it is a workaround, why would it take more than 4s for a single confirmation?

gsebil08 (approved):
LGTM

EduardoRFS (approved):
My bad, this was my mistake.

konata saying good job

PR #488, opened 2022-03-10
benchmarking: added core_bench to esy.json

InfiniteSwerve:
## Problem

Need access to core_bench to proceed with benchmarking the VM. 

## Solution

Built core_bench separately and referenced in esy.json


gsebil08 (approved):
LGTM 👍

PR #493, opened 2022-03-17
chore: OCaml 5 bug changed the process name

EduardoRFS:
## Problem

Due to https://github.com/ocaml/ocaml/issues/11090 deku processes are currently named Domain0, breaking commands like `killall`.

## Solution

This is a workaround until it is fixed on upstream, manually setting the domain name fixes the problem.


d4hines (approved):
Tested. LGTM :+1:

PR #495, opened 2022-03-18
tezos_interop: move scripts blob to it's own file

EduardoRFS:
## Problem

Due to a weird interaction between esy and `ppx_blob`, merlin cannot find the needed asset leading to the entire `Tezos_interop` not showing properly on the LSP.

## Solution

This is more like an workaround, by moving all the blobs to a different module(`Scripts`), now LSP on `Tezos_interop` is working properly.


gsebil08 (approved):

Zett98 (approved):

d4hines (approved):
LGTM

PR #496, opened 2022-03-18
Update openssl resolution from 1.1.1g to 1.1.1l

fraidev:
## Problem

macOS arm64 users need OpenSSL 1.1.1.l.

## Solution

Use the new resolution from [esy-opam-override](https://github.com/esy-ocaml/esy-opam-override/pull/123)


EduardoRFS (approved):
anime girl approves Deku on M1

PR #497, opened 2022-03-19
tezos_interop: make run_entrypoint long lived

EduardoRFS:
## Depends

- [x] #495

## Problem

Due to Multicore not working with `Unix.fork` we had some problems on #414, this happens because spawning processes doesn't work after multiple domains are running.

## Solution

To work around this problem, this PR makes `run_entrypoint.js` long lived and critical, so if it fails the node also dies with it.

## **Warning**

I didn't review this PR, so please be careful. But it seems to work locally

### DIFF
--- a/src/tezos_interop/run_entrypoint.js
+++ b/src/tezos_interop/run_entrypoint.js
@@ -73,23 +73,20 @@ const read = (callback) => {
       }
       buf = buf + chunk;
 
-      const newline = buf.indexOf("\n");
-      if (newline === -1) {
-        done();
-        return;
-      }
+      const parts = buf.split("\n");
+      const messages = parts.slice(0, -1); // everything except last
 
-      const message = buf.slice(0, newline);
-      // skips the newline
-      buf = buf.slice(newline + 1);
+      buf = parts.slice(-1)[0]; // last
 
-      try {
-        const json = JSON.parse(message);
-        this.push(json);
-        done();
-      } catch (err) {
-        done(err);
+      for (const message of messages) {
+        try {
+          const json = JSON.parse(message);
+          this.push(json);
+        } catch (err) {
+          return done(err);
+        }
       }
+      return done();
     },
   });
   const errorHandler = (err = "close") => failure(err);
--- a/src/tezos_interop/run_entrypoint.js
+++ b/src/tezos_interop/run_entrypoint.js
@@ -1,11 +1,11 @@
 "use strict";
 
-const fs = require("fs");
 const { TezosToolkit } = require("@taquito/taquito");
 const { InMemorySigner } = require("@taquito/signer");
+const { pipeline, Transform } = require("stream");
 
 /**
- * @typedef Input
+ * @typedef TransactionRequest
  * @type {object}
  * @property {string} rpc_node
  * @property {string} secret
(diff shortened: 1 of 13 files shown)


src/tezos_interop/tezos_interop.ml
```diff
@@ -179,18 +143,18 @@ module Listen_transactions = struct
...
     Lwt.async start
 end
 module Consensus = struct
   open Michelson.Michelson_v1_primitives
   open Tezos_micheline
-  let commit_state_hash ~context ~block_height ~block_payload_hash ~state_hash
+  let commit_state_hash t ~block_height ~block_payload_hash ~state_hash
       ~withdrawal_handles_hash ~validators ~signatures =
     let module Payload = struct
       type t = {
```
gsebil08, line 160: I am confused about this type being defined within `commit_state_hash` and not "globally", would you mind explain me why this is the good thing to do here instead of splitting `Payload` to a dedicated module?

d4hines: For what purpose? The structure of the payload is only needed at the interface between the smart contract and Deku, so I think it makes sense to keep it relatively private.

src/tezos_interop/run_entrypoint.ml
```diff
@@ -0,0 +1,218 @@
...
+    let module T = struct
+      type header = { id : int } [@@deriving of_yojson { strict = false }]
+    end in
+    let%ok { id } = T.header_of_yojson json in
+    let%ok content = transaction_of_yojson json in
+    Ok { id; content }
+end
+
+module Process : sig
+  type t
```
d4hines, line 88: What is the purpose of defining the sig both here and in the mli?

src/tezos_interop/tezos_interop.ml
EduardoRFS: That's exactly it, it would be reasonable to move it to `Consensus.Payload` but by default just put stuff where you're gonna use it.

src/tezos_interop/run_entrypoint.js
```diff
@@ -1,63 +1,144 @@
...
+};
+
+/** @param {Response} message */
+const write = (message) => {
+  const callback = (err) => {
+    if (err) {
+      failure(err);
+    }
+  };
+  const messageString = JSON.stringify(message);
```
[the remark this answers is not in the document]
d4hines: What do you think about using a global error handler:
```
try {...}
catch (err) {
    console.error(err);
    process.exit(1);
}
```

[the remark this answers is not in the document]
d4hines: Oh, I didn't read the whole file yet. We're doing that at the bottom with `.catch(failure)`. Nvm

d4hines (commented):
This looks good. I did some testing by manually running a cluster and randomly killing the block producer. I've left just a few comments. Address them and I'll approve.

src/tezos_interop/run_entrypoint.ml
```diff
@@ -2,63 +2,217 @@ open Helpers
...
+    let command = (node, [|node; file|]) in
+
+    let process = Lwt_process.open_process command in
+
+    (* handle process failure *)
+    Lwt.async (fun () ->
+        Lwt.catch
+          (fun () ->
+            (* this should never return *)
+            let%await status = process#status in
```
anchpop, line 127: noob question, what's the `#` here? I've only seen it when matching on polyvars

src/tezos_interop/run_entrypoint.ml
```diff
@@ -0,0 +1,218 @@
...
+    Lwt.async (fun () ->
+        Lwt.catch
+          (fun () -> Lwt_io.write_lines process#stdin input_stream)
+          on_fail);
+
+    let output_stream = Lwt_io.read_lines process#stdout in
+    let output_stream =
+      Lwt_stream.map
+        (fun line ->
+          Format.eprintf "Run_entrypoint: %s\n%!" line;
```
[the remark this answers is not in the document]
anchpop: As before I don't understand why we're using eprintf and not fprintf

src/tezos_interop/run_entrypoint.ml
```diff
@@ -2,63 +2,217 @@ open Helpers
...
+
+  let on_fail exn =
+    Format.eprintf "tezos_interop failure: %s\n%!" (Printexc.to_string exn);
+    exit 1
+
+  let spawn_node_process ~file ~to_yojson ~of_yojson input_stream =
+    let node = "node" in
+    let command = (node, [|node; file|]) in
+
+    let process = Lwt_process.open_process command in
```
anchpop, line 120: LWT is still black magic to me. I can't find [any real documentation for this function](https://dsheets.github.io/codoc/lwt.2.4.6/_build/src/unix/lwt_process/index.html#/val:open_process) so I'm not sure I can review this section of the code properly

anchpop (approved):

src/tezos_interop/run_entrypoint.ml
EduardoRFS: Why not? It's an abstraction layer ensuring that inside of run_entrypoint.ml process is not used in the wrong way. And if it's here it must be in the mli.

src/tezos_interop/run_entrypoint.ml
```diff
@@ -0,0 +1,218 @@
...
+    (* handle process failure *)
+    Lwt.async (fun () ->
+        Lwt.catch
+          (fun () ->
+            (* this should never return *)
+            let%await status = process#status in
+            raise (Process_closed status))
+          on_fail);
+
+    (* TODO: test that exception on input and output kill the node *)
```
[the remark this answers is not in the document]
EduardoRFS: It's not as simple as that to test it, otherwise I would had done it(and in fact I did), there is many streams, and many places where errors can happen and I'm not sure how exceptions works in the presence of Lwt_stream.map.

src/tezos_interop/run_entrypoint.ml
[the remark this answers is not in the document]
EduardoRFS: @anchpop stderr

src/tezos_interop/run_entrypoint.js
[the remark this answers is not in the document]
EduardoRFS: Also, JSON.stringify will only fail if someone does a very cursed thing, like overloading .toJSON() with an exception inside. I hope that JS people are sane enough.

src/tezos_interop/run_entrypoint.ml
EduardoRFS: object access `obj#field`

src/tezos_interop/run_entrypoint.ml
EduardoRFS: Even if there was documentation it would not document the behavior properly, this requires a LOT of Unix knowledge to understand in details, which I don't have. It's one of the cases where even if you look at the implementation is not enough and I did.

d4hines (approved):
LGTM

PR #500, opened 2022-03-20
tezos_interop: integrates fetch_storage

EduardoRFS:
## Depends

- [x] #499

## Problem

After #499 Deku should have an unified tezos bridge, but currently `fetch_storage.js` is not integrated on it.

## Solution

This PR moves `fetch_storage.js` entirely to `tezos_js_bridge.js` and properly integrates it on Tezos_interop.


d4hines (approved):
LGTM. Appreciate the better abstractions we're forming around the JS layer.

hooray for better abstractions

anchpop:
@EduardoRFS should I rebase #450 onto this branch? I add a js script for grabbing the validators from the discovery.mligo big map, and I think that should be merged into here?

EduardoRFS:
@anchpop I would say yes

src/tezos_interop/tezos_js_bridge.js
```diff
@@ -124,13 +139,56 @@ const onRequest = async (request) => {
...
+  const Tezos = new TezosToolkit(rpc_node);
+  Tezos.setProvider({ config });
+
+  const block = await client.getBlock(); // fetches the head
+  const contract = await client.getContract(destination, {
+    block: block.hash,
+  });
+  const storage = contract.script.storage;
+  /* To make sure the storage state is finalized, we query the last
+     block and any one of it's operation, and wait till it receives
```
gsebil08, line 155:
```suggestion
     block and any one of its operation, and wait till it receives
```

src/tezos_interop/tezos_js_bridge.js
```diff
@@ -124,13 +139,56 @@ const onRequest = async (request) => {
...
+  const { rpc_node, confirmation, destination } = request;
+  const client = new RpcClient(rpc_node);
+  const Tezos = new TezosToolkit(rpc_node);
+  Tezos.setProvider({ config });
+
+  const block = await client.getBlock(); // fetches the head
+  const contract = await client.getContract(destination, {
+    block: block.hash,
+  });
+  const storage = contract.script.storage;
```
gsebil08, line 153: Why extracting it so long before needing it? I expected to find some transformation on it. I think this line should be right before the `return` one.

src/tezos_interop/tezos_js_bridge.js
```diff
@@ -104,7 +119,7 @@ const config = {
 };
 
 /** @param {TransactionRequest} request */
-const onRequest = async (request) => {
+const onTransactionRequest = async (request) => {
```
gsebil08, line 122: Clever renaming, thanks.

aguillon (approved):

src/tezos_interop/tezos_bridge.ml
```diff
@@ -50,6 +53,42 @@ module Transaction = struct
     | _ -> Error "invalid status"
 end
 
+module Storage = struct
+  type kind = Storage
+  let kind_to_yojson Storage = `String "storage"
```
gsebil08, line 58: Noob question: would it be the same to do:
```ocaml
type kind = Storage [@name storage]
  [@@deriving to_yojson]
```

gsebil08 (approved):
Newbie questions, but LGTM.

src/tezos_interop/tezos_bridge.ml
EduardoRFS: @gsebil08 nope, also tried that, TLDR variants are encoded as `[header, ...values]` so it endup as `["storage"]`

src/tezos_interop/tezos_js_bridge.js
EduardoRFS: I just kept the code in the same order, this is quite fragile and I didn't want to break it.

anchpop (approved):

aguillon (approved):

PR #501, opened 2022-03-20
tezos_interop: integrates listen_transactions

EduardoRFS:
## Depends

- [x] #500

## Problem

Similar to #499 `listen_transactions.js` is not integrated on JS the bridge, using different infrastructure.

## Solution

This PR implements a mechanism where a bridge request can have multiple responses to model the behavior of `listen_transactions.js` and then integrates `listen_transactions.js` on the bridge.


src/tezos_interop/tezos_bridge.mli
```diff
@@ -1,11 +1,22 @@
 open Crypto
 open Tezos
 
-type t
-val spawn : unit -> t
+(* push *)
```
d4hines, line 4: This single word in context doesn't signify anything to me. What do you mean here?

src/tezos_interop/tezos_bridge.ml
```diff
@@ -93,10 +116,22 @@ type t = Long_lived_js_process.t
 let spawn () =
   let file = Scripts.file_tezos_js_bridge in
   Long_lived_js_process.spawn ~file
-let transaction t ~rpc_node ~secret ~required_confirmations ~destination
+let listen_transaction t ~rpc_node ~required_confirmations ~destination =
```
d4hines, line 119:
```suggestion

let listen_transaction t ~rpc_node ~required_confirmations ~destination =
```
To keep consistent with the rest of the file we should have a newline separating the functions. We need a linting rule to standardize this - I don't care one way or the other, but it's confusing when it switches.

d4hines (approved):
LGTM. I did a quick smoke test as well. Again, will test more thoroughly by rebasing #414 and testing that.

src/tezos_interop/tezos_bridge.mli
EduardoRFS: https://en.wikipedia.org/wiki/HTTP/2_Server_Push

src/tezos_interop/tezos_bridge.ml
EduardoRFS: There is no standard, the switch in many places are intentional, this was one missing

src/tezos_interop/tezos_bridge.ml
```diff
@@ -2,6 +2,36 @@ open Helpers
...
+module Michelson = struct
+  include Michelson
+  type t = Michelson.t
+  let of_yojson json =
+    let%ok json = Yojson.Safe.to_string json |> Data_encoding.Json.from_string in
+    try Ok (Data_encoding.Json.destruct Michelson.expr_encoding json) with
+    | _ -> Error "invalid json"
+end
+module Listen_transaction = struct
+  type kind = Listen
```
gsebil08, line 14: I am not sure about this, but I feel like this `kind = listen` is kinda the same that the one in `Pending`.
Do they have the same functional meaning?

src/tezos_interop/tezos_bridge.mli
gsebil08: Agreed with @d4hines, the single word does not help. I think a full sentence briefly explaining **why** instead of **what** would be better.
What do you think?

src/tezos_interop/tezos_interop.ml
```diff
@@ -172,7 +111,9 @@ module Consensus = struct
     transactions : transaction list;
   }
   let parse_transaction transaction =
-    match (transaction.Listen_transactions.entrypoint, transaction.value) with
+    let Tezos_bridge.Listen_transaction.{ entrypoint; value } = transaction in
```
gsebil08, line 114: What about record pattern here:
```ocaml
  let parse_transaction Tezos_bridge.Listen_transaction.{ entrypoint; value } =
    let value = Micheline.root value in
```
since we don't need the `transaction` but only each sub-field.

src/tezos_interop/tezos_interop.ml
```diff
@@ -234,15 +175,18 @@ module Consensus = struct
       Some (Deposit { ticket; destination; amount })
     | _ -> None
   let parse_operation output =
-    let%some hash = Operation_hash.of_string output.Listen_transactions.hash in
-    let transactions = List.filter_map parse_transaction output.transactions in
+    let Tezos_bridge.Listen_transaction.{ hash; transactions } = output in
```
gsebil08, line 178: Same question here, record pattern?

src/tezos_interop/tezos_bridge.mli
aguillon: @EduardoRFS yes that's more or less how @d4hines and I understood it. I'm in favor of adding more documentation than a single word but I won't block the PR over it.

gsebil08 (approved):
Ok with the content, questions about the form. Nothing to block the merge.

src/tezos_interop/tezos_bridge.ml
aguillon: @gsebil08 yes and they're the same as the ones expected by the JavaScript part (see tezos_js_bridge.js).

@EduardoRFS: Following @gsebil08 question, is there a reason you went for a polymorphic `request` function with explicit parameters to handle conversion instead of using modules/objects?

aguillon (approved):

src/tezos_interop/tezos_bridge.ml
gsebil08: @aguillon 
>yes and they're the same as the ones expected by the JavaScript part (see tezos_js_bridge.js).

They **have the same functional meaning**, but they **are not** the same :)
That was the hidden part of my question, why don't we define it only once?

src/tezos_interop/tezos_bridge.mli
EduardoRFS: It's not a `why` or `what`, nor `documentation` is just a tag, it means a push. If you don't know what is a push this is not the place to explain.

More references on this. 

https://en.wikipedia.org/wiki/Push_technology

src/tezos_interop/tezos_bridge.ml
EduardoRFS: Sorry, where is kind defined twice? Pending.kind is not the same as Listen_transaction.kind, they don't have the same functional meaning either.

If your question is `why not define as Listen | Pending` because that would allow to describe more bad states.

src/tezos_interop/tezos_interop.ml
EduardoRFS: I personally don't like it, the function header now has so much more noise, before I could read as `parse_transaction transaction` which is the function parse_transaction that accepts a transaction as a parameter, on your code I can see that it accept a record with `entrypoint` and `value`, but I don't know without looking at the type on what is the name that you give to this record.

Looking at the type is okay, but essentially my point is you removed a very useful name.

src/tezos_interop/tezos_interop.ml
EduardoRFS: Same answer as above

anchpop (approved):

src/tezos_interop/long_lived_js_process.ml
```diff
@@ -61,30 +61,40 @@ module Pending : sig
...
+  let push t id json =
     match find t id with
-    | Some resolver ->
-      Lwt.wakeup_later resolver json;
+    | Some (Request, resolver) ->
+      t := Id.Map.remove id !t;
+      resolver json;
+      Ok ()
+    | Some (Listen, resolver) ->
+      resolver json;
```
anchpop, line 97: This normally removes something from the pending list, so I think it would be more appropriate to call it `Pending.pop` or something like that. I think `Pending.push` implies that it's adding something to the list of pending operations. (I didn't mind `Pending.resolve` either)

EduardoRFS: Pending.resolve is also not correct as a Listen is not resolved, will be leaving the current name to not block on this.

PR #503, opened 2022-03-22
bumped core reference

InfiniteSwerve:
## Problem

We want to run build and benching commands with esy to conform to the deku standard. This was not possible due to the missing implementation of memprof in ocaml5. 

## Solution

This PR adds a reference to a more useful version of core that noops memprof, as it is not needed currently. 

This should allow building this branch with `esy b dune build` and execution with `esy b dune exec ./benchmarks/benchmarks <bench>`

## Related

The referenced version of core: https://github.com/InfiniteSwerve/core


lykimq:
Lgtm.

lykimq:
Lgtm.

PR #510, opened 2022-03-28
Add Dockerfile using esy at runtime

ulrikstrid:
## Problem

The current docker setup is broken for some reason and we want 

## Solution

Since there's a demo coming up we just want some docker setup to work for the testnet, this is a workaround that just uses `esy x ...` at runtime to get things going. Should be good enough until we can fix the static image.

### DIFF
--- a/docker/esy.Dockerfile
+++ b/docker/esy.Dockerfile
@@ -1,0 +1,33 @@
+FROM alpine:latest as certs
+RUN apk --update add ca-certificates
+
+FROM esydev/esy:nightly-alpine-latest as builder
+
+RUN apk add libexecinfo-dev
+
+WORKDIR /app
+
+# Add things that doesn't change much, or should bust cache when it does
+COPY ./esy.lock esy.json ./
+
+RUN esy install
+RUN esy build-dependencies --release
+
+# Copy the rest of the files
+COPY . .
+
+# TODO: investigate why esy complains that it's not installed if we don't install again
+RUN esy install
+RUN esy build
+
+# Copy the static binaries to a known location
+RUN esy cp "#{self.target_dir / 'default' / 'src' / 'bin' / 'deku_cli.exe'}" deku_cli.exe && \
+    esy cp "#{self.target_dir / 'default' / 'src' / 'bin' / 'deku_node.exe'}" deku_node.exe
+RUN strip ./deku_node.exe && strip ./deku_cli.exe
+
+ENV SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt
+ENV SSL_CERT_DIR=/etc/ssl/certs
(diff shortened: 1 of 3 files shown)


docker/esy.Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN esy build
+
+# Copy the static binaries to a known location
+RUN esy cp "#{self.target_dir / 'default' / 'src' / 'bin' / 'deku_cli.exe'}" deku_cli.exe && \
+    esy cp "#{self.target_dir / 'default' / 'src' / 'bin' / 'deku_node.exe'}" deku_node.exe
+RUN strip ./deku_node.exe && strip ./deku_cli.exe
+
+ENV SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt
+ENV SSL_CERT_DIR=/etc/ssl/certs
+COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
```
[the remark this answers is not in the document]
ulrikstrid: I just wanted to keep it as close as possible but you are correct 👌

docker/esy.Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+# Copy the rest of the files
+COPY . .
+
+# TODO: investigate why esy complains that it's not installed if we don't install again
+RUN esy install
+RUN esy build
+
+# Copy the static binaries to a known location
+RUN esy cp "#{self.target_dir / 'default' / 'src' / 'bin' / 'deku_cli.exe'}" deku_cli.exe && \
```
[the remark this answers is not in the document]
ulrikstrid: Yes, didn't think about it

d4hines (approved):
I tested this by building locally and running while linking `/app/data` to the correct files. The node came up successfully (I didn't test a cluster but I think this test was sufficient).

anchpop (approved):

docker/esy.Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM esydev/esy:nightly-alpine-latest as builder
+
+RUN apk add libexecinfo-dev && apk --update add ca-certificates
+
+WORKDIR /app
+
+# Add things that doesn't change much, or should bust cache when it does
+COPY ./esy.lock esy.json ./
```
callistonianembrace, line 8: This doesn't copy the `esy.lock` directory properly (causing esy to re-solve). I used

```
COPY ./esy.lock ./esy.lock
COPY ./esy.json ./esy.json
```

ulrikstrid: Can you push this change? This is probably why we need the second `esy install`

ulrikstrid: Actually, I can make the change

ulrikstrid: I fixed this, removed the extra `esy install` and changed to `esy build --release` things are much better now :)

PR #514, opened 2022-03-28
Temporary Hack: Turn off Nix CI

### LINKED ISSUE #487: Update Cmdliner usage
The next version of Cmdliner has breaking changes. We should upate deku-node and deku-cli accordingly.

d4hines:
## Problem

Nix CI is currently broken. 

## Solution

Turn it off temporarily. I'm working on a PR to fix it but it's very low priority and I don't want people to have to wait with red CI. I also think it will be cleaner if we fix #487 first.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,9 +1,5 @@
 name: "Test in nix"
-on:
-  pull_request:
-  push:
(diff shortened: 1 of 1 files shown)


d4hines:
Superseded by #518

PR #515, opened 2022-03-28
Smart contracts: Origination of smart-contracts

Zett98:
## Depends
- [x] #506

## Problem
Currently we can't originate the contracts

## Solution
add simple code that allows originating the contract

PR #516, opened 2022-03-28
Smart contracts: Transfers to and between originated addresses

Zett98:
## Depends
- [ ] #506
- [ ] #515

## Problem
We need transfers to set change balance of the smart contract on origination/invocation
## Solution
add the ability to transfer amount from implicit address to originated one.(not so useful for now, but will be useful when we will have ctez)


EduardoRFS (changes requested):
As discussed in private, contracts will not be in this ledger, that's not how tickets will behave in general, look at the Ticket table PR for more ideas.

PR #518, opened 2022-03-29
fix: add alcotest to nix setup

d4hines:
## Problem

The nix build is broken because it doesn't include alcotest and qcheck-alcotest.

## Solution

Add those packages. Also includes the overrides necessary for them to build (I used .esy.lock/index.json as the reference).


aguillon (commented):
Thank you!

aguillon (approved):

PR #519, opened 2022-03-29
Fix broken Githun action ID

d4hines:
## Problem

Observe that our Docker CI is failing because we have multiple builds with id `docker_build`: https://github.com/marigold-dev/deku/actions/runs/2059345040

## Solution

- change the id's to be unique
- ~disable the docker static build, since it's broken anyway right now.~

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -23,24 +23,25 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Build and push
-        # Only push if on main branch
-        id: docker_build
-        uses: docker/build-push-action@v2
-        with:
-          file: ./docker/static.Dockerfile
-          context: .
-          platforms: linux/amd64
-          cache-from: type=gha
-          cache-to: type=gha,mode=max
-          push: ${{ github.ref == 'refs/heads/main' && github.event_name == 'push' }}
-          tags: |
-            ghcr.io/marigold-dev/deku:latest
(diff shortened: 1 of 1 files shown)


EduardoRFS:
Please, just do the proper thing. https://github.com/marigold-dev/deku/pull/522

.github/workflows/docker.yml
```diff
@@ -23,24 +23,25 @@ jobs:
...
+      #   with:
+      #     file: ./docker/static.Dockerfile
+      #     context: .
+      #     platforms: linux/amd64
+      #     cache-from: type=gha
+      #     cache-to: type=gha,mode=max
+      #     push: ${{ github.ref == 'refs/heads/main' && github.event_name == 'push' }}
+      #     tags: |
+      #       ghcr.io/marigold-dev/deku:latest
+      #       ghcr.io/marigold-dev/deku:${{ github.sha }}
```
ulrikstrid, line 40: If we uncomment this we can merge it

ulrikstrid (changes requested):
Since #522 fixes the static build we can un-comment but keep the changed id

.github/workflows/docker.yml
ulrikstrid: I took the liberty to do this

ulrikstrid (approved):

ulrikstrid:
Since this PR is strictly better than what we have in `main` I think we should merge it so that we can keep moving on the testnet before merging the new static build

d4hines:
superseded by #522

PR #508, opened 2022-03-25
lambda_vm: support simple let expressions in lambda vm

renatoalencar:
## Problem

ppx_lambda_vm doesn't support expressions like `let a = b in c`, which makes some kind of abstractions difficult.

## Solution

Add support for simple let expressions encoding them as immediate calling lambdas. 

Expressions like `let a = b in c` are encoded like `(fun a -> c) b`.

PR #520, opened 2022-03-29
Fix the land primitive on ppx_lambda_vm

renatoalencar:
## Depends

- [x] #508 

## Problem

Using the `land` primitive on ppx_lambda_vm translates to `And` and not `Land` which gives the following error:

```
1 | let _a = [%lambda_vm 0L land 1L]
                            ^^^^
Error: This variant expression is expected to have type prim
       There is no constructor And within type prim
```

## Solution

Use the proper type constructor

### DIFF
--- a/ppx_lambda_vm/ppx_lambda_vm.ml
+++ b/ppx_lambda_vm/ppx_lambda_vm.ml
@@ -79,7 +79,7 @@ let rec expr_of_ocaml_expr expr =
     | "*" -> [%expr Prim Mul]
     | "/" -> [%expr Prim Div]
     | "mod" -> [%expr Prim Rem]
-    | "land" -> [%expr Prim And]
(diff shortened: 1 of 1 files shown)


ppx_lambda_vm/ppx_lambda_vm.ml
```diff
@@ -80,7 +79,7 @@ let rec expr_of_ocaml_expr expr =
     | "*" -> [%expr Prim Mul]
     | "/" -> [%expr Prim Div]
     | "mod" -> [%expr Prim Rem]
-    | "land" -> [%expr Prim And]
+    | "land" -> [%expr Prim Land]
```
EduardoRFS, line 82: mb

gsebil08 (approved):
LGTM

Zett98 (approved):

PR #521, opened 2022-03-29
Voting contract test for Lambda VM

renatoalencar:
## Depends

- [x] #508 
- [x] #520 

## Problem

We need to make sure that Lambda VM works for more complex contracts, for this we're proposing a voting contract with voting, pausing and resetting capabilities.

## Solution

Encode the contract with ppx lambda vm and check if the contract produces correct outputs.

The contract also encodes lists and strings as pairs of pairs, using the first element of the pair as the size of the list
in order to have a point of stop when iterating through the list and the second element being the pair of pairs as the contents of the list. The list ends with a `(0L, 0L)` so it doesn't crash from trying to unpair an integer.

### Strings

Strings are encoded as follows:

```
(<size>, (<first char code>, (<second char code>, (third char code, .... (<last char code, 0L)...)
```

Example: `"Tezos"`

```ocaml
(5L, (84L, (101L, (122L, (111L, (115L, 0L))))))
```

### Lists

List (in this case associative lists) are encoded a similar way.

```
(<size>, (<first element>, (<second element>, ...., (0L, 0L))...)
```

### DIFF
--- a/tests/vm/test_vm.ml
+++ b/tests/vm/test_vm.ml
@@ -9,4 +9,5 @@ let () =
       Test_simple_expr.test;
       Test_gas.test;
       Test_io.test;
+      Test_voting_contract.test;
     ]
--- a/tests/vm/test_voting_contract.ml
+++ b/tests/vm/test_voting_contract.ml
@@ -1,0 +1,401 @@
+open Lambda_vm
+
+let value_of_string gas string =
+  Lambda_vm.Ast.value_of_string (Gas.make ~initial_gas:gas) string
+
+let compare =
+  [%lambda_vm
+    let delta f = f f in
+    let aux aux size a b =
+      if size then
+        if fst a - fst b then
+          0L
+        else
+          aux aux (size - 1L) (snd a) (snd b)
+      else
+        1L in
+    let aux = delta aux in
+    fun a b ->
+      let size_a = fst a in
+      let size_b = fst b in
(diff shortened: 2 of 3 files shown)


EduardoRFS:
This should not be done using ppx_lambda_vm, this is a real contract and should be implemented in something like Ligo.

renatoalencar:
> This should not be done using ppx_lambda_vm, this is a real contract and should be implemented in something like Ligo.

@EduardoRFS I was asked to do that next and unfortunately we still don't have a compiler from Ligo to lambda vm. I believe that's for validating that writing more complex contracts are possible indeed. I guess @Zett98 could have a better explanation.

Zett98:
Yup, as we don't have ligo to lambda yet, this is a good solution to have multiple contracts running

EduardoRFS:
This is not a good solution, if you want to just have a contract calling another one this is too overkill. And if you want to test lambda vm with real contract it should be written in Ligo.

renatoalencar:
> This is not a good solution, if you want to just have a contract calling another one this is too overkill. And if you want to test lambda vm with real contract it should be written in Ligo.
>
> Will not block, but the PR doesn't make a lot of sense to me.

I partially agree with you, we can notes as comments to migrate this to Ligo as soon as we have a Ligo to Lambda VM compiler. What do you think?

Zett98 (approved):

PR #522, opened 2022-03-30
docker: fix static image

EduardoRFS:
## Problem

Currently the static docker image is not working.

## Solution

Ensuring that only `deku_{node,cli}.exe` builds and also removing the unused / broken flags fix the static binary built. But Deku needs NodeJS to run, so for now let's use the Alpine image instead of scratch.

### DIFF
--- a/docker/static.Dockerfile
+++ b/docker/static.Dockerfile
@@ -3,7 +3,7 @@ RUN apk --update add ca-certificates
 
 FROM esydev/esy:nightly-alpine-latest as builder
 
-RUN apk add libexecinfo-dev
+RUN apk add libexecinfo-dev libexecinfo-static
 
 WORKDIR /app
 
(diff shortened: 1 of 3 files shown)


docker/static.Dockerfile
```diff
@@ -19,14 +19,16 @@ COPY . .
...
-RUN esy build_static
+RUN esy dune build --profile=static src/bin/deku_cli.exe src/bin/deku_node.exe
 
 # Copy the static binaries to a known location
 RUN esy cp "#{self.target_dir / 'default' / 'src' / 'bin' / 'deku_cli.exe'}" deku_cli.exe && \
     esy cp "#{self.target_dir / 'default' / 'src' / 'bin' / 'deku_node.exe'}" deku_node.exe
 RUN strip ./deku_node.exe && strip ./deku_cli.exe
 
-FROM scratch as runtime
+FROM alpine:latest as runtime
```
ulrikstrid, line 29: We don't need alpine when running I think so we might as well just use a slim node image since they are more secure
```suggestion
FROM node:lts-slim as runtime
```

EduardoRFS: Can you explain why they're more secure?

EduardoRFS: fair

d4hines (approved):
LGTM

ulrikstrid (approved):
Looks good

PR #524, opened 2022-04-01
Add deku-cli and deku-node as apps in nix

ulrikstrid:
## Problem

Now that we have flakes it would be nice to have a simple way of running our tools with it

## Solution

Adding the `deku-cli` and `deku-node` as apps.
They can now be run from the repository like `nix run .#deku-cli` or `nix run .#deku-node`.
We should also be able to run `nix run github:marigold-dev/deku#deku-cli` from anywhere.

## Testing

Checkout branch and run  `nix run .#deku-cli` or `nix run .#deku-node`

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -41,5 +41,16 @@
       in {
         devShell = import ./nix/shell.nix { inherit pkgs deku npmPackages; };
         packages = { inherit deku; };
+
+        apps = {
+          deku-cli = {
+            type = "app";
+            program = "${deku}/bin/deku-cli";
+          };
+          deku-node = {
+            type = "app";
+            program = "${deku}/bin/deku-node";
+          };
+        };
       });
 }
--- a/flake.nix
+++ b/flake.nix
@@ -3,10 +3,8 @@
 
   # Setup trusted binary caches
   nixConfig = {
-    trusted-substituters = [
-      "https://cache.nixos.org/"
-      "https://deku.cachix.org"
-    ];
+    trusted-substituters =
+      [ "https://cache.nixos.org/" "https://deku.cachix.org" ];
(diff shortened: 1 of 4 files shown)


d4hines (changes requested):
What's going on with the formatting? I'd appreciate if we could separate formatting into separate commits. 

Besides that, change looks good.

gsebil08 (commented):
Does it worth it to add some documentation?

flake.nix
```diff
@@ -18,29 +16,41 @@
...
+        };
+      in {
+        devShell = import ./nix/shell.nix { inherit pkgs deku npmPackages; };
+        packages = { inherit deku; };
 
-          deku = pkgs.callPackage ./nix/deku.nix {
-            doCheck = true;
-            nodejs = pkgs.nodejs-12_x;
-            inherit npmPackages;
+        app = {
```
d4hines, line 45:
```suggestion
        apps = {
```

d4hines (changes requested):
There's a typo in flake.nix

d4hines (approved):
LGTM

gsebil08:
> > Does it worth it to add some documentation?
> 
> @gsebil08, personally I don't think so: EDIT: I had the command wrong the first time. Fixed.
> 
> ```
> nix run .#deku-cli -- --help
> nix run .#deku-node -- --help
> ```
> 
> will print out quite a bit of documentation.
> 
> What kind of docs are you looking for?

Basic simple user documentation, like in our `readme.md` add something like:
> ## Run a node with nix
> Simply run the following commands to run a deku-node:
>```shell script
> $ nix run .#deku -- data/0
> ```

PR #525, opened 2022-04-01
Hack to expose metrics for prometheus

### LINKED ISSUE #327: Observability: metrics

ulrikstrid:
## Problem

The infra team wants metrics so that we can be alerted when things go bad

## Solution

Adding prometheus as a separate dream server that we can scrape for data

## Related

- Closes #327

### DIFF
--- a/nix/deku.nix
+++ b/nix/deku.nix
@@ -42,6 +42,7 @@ pkgs.ocamlPackages.buildDunePackage {
     domainslib
     utop
     reason
+    prometheus
   ]);
 
   propagatedBuildInputs = [
--- a/nix/overlay.nix
+++ b/nix/overlay.nix
@@ -1,6 +1,7 @@
 final: prev:
-let disableCheck = package:
-  package.overrideAttrs (o: { doCheck = false; });
+let 
+  inherit (prev) lib stdenv;
+  disableCheck = package: package.overrideAttrs (o: { doCheck = false; });
 in
 {
   ocamlPackages = prev.ocaml-ng.ocamlPackages_5_00.overrideScope' (oself: osuper: rec {
@@ -38,5 +39,41 @@ in
         sha256 = "13c53b1cxkq2nj444655skw5a1mcxzbaqwqsqjf7jbwradb3hmxa";
       };
     });
+
+    asetmap = stdenv.mkDerivation rec {
+      version = "0.8.1";
+      pname = "asetmap";
+      src = prev.fetchurl {
+        url = "https://github.com/dbuenzli/asetmap/archive/refs/tags/v0.8.1.tar.gz";
+        sha256 = "051ky0k62xp4inwi6isif56hx5ggazv4jrl7s5lpvn9cj8329frj";
+      };
+
+      nativeBuildInputs = with osuper; [
+        topkg
+        findlib
+        ocamlbuild
+        ocaml
+      ];
+
+      inherit (osuper.topkg) buildPhase installPhase;
+    };
+
+    prometheus = osuper.buildDunePackage rec {
+      version = "1.1.0";
+      pname = "prometheus";
+      src = prev.fetchurl {
+        url = "https://github.com/mirage/prometheus/releases/download/v1.1/prometheus-v1.1.tbz";
+        sha256 = "1r4rylxmhggpwr1i7za15cpxdvgxf0mvr5143pvf9gq2ijr8pkzv";
+      };
+
+      propagatedBuildInputs = with osuper; [
+        astring
+        asetmap
+        fmt
+        re
+        lwt
+        alcotest
+      ];
+    };
   });
 }
--- a/src/bin/deku_node.ml
+++ b/src/bin/deku_node.ml
@@ -122,23 +122,26 @@ let node folder =
   Node.Server.start ~initial:node;
   Dream.initialize_log ~level:`Warning ();
   let port = Node.Server.get_port () |> Option.get in
-  Dream.run ~interface:"0.0.0.0" ~port
-  @@ Dream.router
-       [
-         handle_block_level;
-         handle_received_block_and_signature;
-         handle_received_signature;
-         handle_block_by_hash;
-         handle_protocol_snapshot;
-         handle_request_nonce;
-         handle_register_uri;
-         handle_receive_user_operation_gossip;
-         handle_receive_consensus_operation;
-         handle_withdraw_proof;
-         handle_ticket_balance;
-         handle_trusted_validators_membership;
-       ]
-  @@ Dream.not_found
+  Lwt.all [
+    Dream.serve ~interface:"0.0.0.0" ~port
+    @@ Dream.router
+        [
+          handle_block_level;
+          handle_received_block_and_signature;
+          handle_received_signature;
+          handle_block_by_hash;
+          handle_protocol_snapshot;
+          handle_request_nonce;
+          handle_register_uri;
+          handle_receive_user_operation_gossip;
+          handle_receive_consensus_operation;
+          handle_withdraw_proof;
+          handle_ticket_balance;
+          handle_trusted_validators_membership;
+        ]
+    @@ Dream.not_found;
+    Metrics.Dream_app.serve (Some 9000);
+  ] |> Lwt_main.run |> ignore
 
 (* TODO: https://github.com/ocaml/ocaml/issues/11090 *)
 let () = Domain.set_name "deku-node"
--- a/src/metrics/blocks.ml
+++ b/src/metrics/blocks.ml
@@ -1,0 +1,10 @@
+open Prometheus
+open Config
+
+let blocks_produced_total =
+  let help = "Total number of blocks produced" in
+  Counter.v ~help ~namespace ~subsystem "blocks_produced_total"
+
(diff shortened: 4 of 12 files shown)


sandbox.sh
```diff
@@ -173,7 +173,7 @@ start_deku_cluster() {
   SERVERS=()
   echo "Starting nodes."
   for i in ${VALIDATORS[@]}; do
-    deku-node "$data_directory/$i" --listen-prometheus=9000 &
+    deku-node "$data_directory/$i" --listen-prometheus=900$i &
```
ulrikstrid, line 176: We might want to revert this before merge, but that is something @Zett98 and @aguillon should decide on.  And maybe @d4hines should have a word on it.

d4hines: I'm ok with this change.

src/metrics/dream_app.ml
```diff
@@ -0,0 +1,51 @@
+(* Most of this code is borrowed from https://github.com/mirage/prometheus/blob/master/app/prometheus_unix.ml which is under the Apache license *)
+
+open Prometheus
+open Prometheus_app
+
+module Unix_runtime = struct
```
d4hines, line 6: Could we move this module to `prometheus_app` and deduplicate some of the code? Unless there's some reason not to, in which case it would be good to explain with a comment.

src/metrics/dream_app.ml
```diff
@@ -0,0 +1,51 @@
...
+  let open! Cmdliner in
+  let doc =
+    Arg.info ~docs:"MONITORING OPTIONS" ~docv:"PORT"
+      ~doc:"Port on which to provide Prometheus metrics over HTTP."
+      ["listen-prometheus"] in
+  Arg.(value @@ opt (some int) None doc)
+
+let opts = listen_prometheus
+
+let () =
```
d4hines, line 38: Same as above - I think we can move this into `./promtheus_app.ml` where we do similar.

d4hines (commented):
@ulrikstrid, in your PR description, you mention the license. I saw your note in at least one of the files - is that sufficient? Also, what kind of cleanup do you have in mind?

If we include this we have to be careful to instruct users not to expose this port over the internet, since it would give way too much data to attackers.

In general I want to think carefully about the security implications of adding something like this. I'd like to structure code in such a way as to make it impossible that we make a mistake down the road and accidentally leak secret data.

ulrikstrid:
The license is handled since we broke out the implementation to another repo.

The cleanup is whatever reviewers want, I threw the initial thing together so I expect there to be some things you want changed.

nix/overlay.nix
```diff
@@ -1,42 +1,49 @@
...
-    cmdliner = osuper.cmdliner.overrideAttrs (_: {
-      src = builtins.fetchurl {
-        url = "https://github.com/dbuenzli/cmdliner/archive/refs/tags/v1.0.4.tar.gz";
-        sha256 = "13c53b1cxkq2nj444655skw5a1mcxzbaqwqsqjf7jbwradb3hmxa";
-      };
-    });
-  });
+let disableCheck = package: package.overrideAttrs (o: { doCheck = false; });
+in {
+  ocaml-ng = builtins.mapAttrs (_: ocamlVersion:
```
ulrikstrid, line 4: This change should probably be moved to a separate PR. cc @d4hines and @gsebil08 

It makes it so that we overlay all ocaml versions which should (in theory at least) allow us to use these changes for static compilation and cross-compiling for arm64

src/metrics/dream_app.ml
ulrikstrid: This code is moved to https://github.com/marigold-dev/prometheus-web

ulrikstrid:
@aguillon and @d4hines I'm quite happy with the OCaml parts now, we might want to have a chat about the nix stuff and then we can either break it out or keep it here

nix/deku.nix
```diff
@@ -20,35 +22,32 @@ pkgs.ocamlPackages.buildDunePackage {
...
-  checkInputs = with pkgs.ocamlPackages; [
-    alcotest
-    qcheck
-    qcheck-alcotest
-    rely
-  ];
+  nativeBuildInputs = [ nodejs npmPackages ]
+    ++ (with ocamlPackages; [ utop reason ]);
+
+  propagatedBuildInputs = with ocamlPackages;
```
ulrikstrid, line 28: Most of our dependencies should NOT be `nativeBuildInputs`, only executables are "native"

nix/deku.nix
```diff
@@ -20,35 +22,32 @@ pkgs.ocamlPackages.buildDunePackage {
...
+      hex
+      tezos-micheline
+      digestif
+      ppx_blob
+      secp256k1-internal
+      bigstring
+      domainslib
+      prometheus
+      prometheus-dream
+    ] ++ checkInputs ++ [ npmPackages ];
```
ulrikstrid, line 50: Adding checkInputs here is needed if we want to cross-compile or use `strictDeps = true`

nix/overlay.nix
```diff
@@ -41,5 +41,9 @@ in {
           sha256 = "13c53b1cxkq2nj444655skw5a1mcxzbaqwqsqjf7jbwradb3hmxa";
         };
       });
-    });
+      odoc-parser = osuper.odoc-parser.overrideAttrs (o: {
```
ulrikstrid, line 44: Not sure why I suddenly needed this, but it broke without it

src/metrics/blocks.mli
```diff
@@ -0,0 +1,3 @@
+val inc_block_produced : unit -> unit
```
ulrikstrid, line 1: If we keep using a API like this we can be sure we're not exposing anything we shouldn't by just looking in the `metrics` library

src/node/flows.ml
```diff
@@ -165,11 +165,13 @@ let try_to_produce_block state update_state =
   let signature = sign ~key:state.identity.secret block in
   let state = append_signature state update_state ~signature ~hash:block.hash in
   broadcast_block_and_signature state ~block ~signature;
+  Metrics.Blocks.inc_block_produced ();
```
ulrikstrid, line 168: Are these at the correct place?

src/metrics/config.ml
```diff
@@ -0,0 +1,2 @@
+let namespace = "Deku"
+let subsystem = "node"
```
ulrikstrid, line 2: I think the subsystem maybe should be per module, so `block`should be one, but this is something we can tweak as we go along

src/bin/deku_node.ml
```diff
@@ -150,5 +156,5 @@ let node =
     let open Arg in
     required & pos 0 (some string) None & info [] ~doc ~docv in
   let open Term in
-  const node $ folder_node
+  const node $ folder_node $ Prometheus_dream.opts
```
ulrikstrid, line 159: Add the `--listen-prometheus=<port>` option, if it's not given we won't start the prometheus server

flake.nix
```diff
@@ -13,17 +13,26 @@
...
     nix-npm-buildpackage.url = "github:serokell/nix-npm-buildpackage";
   };
 
-  outputs =
-    { self, nixpkgs, flake-utils, nix-npm-buildpackage, ocaml-overlays }:
+  outputs = { self, nixpkgs, flake-utils, nix-npm-buildpackage, ocaml-overlays
+    , prometheus-web }:
     with flake-utils.lib;
-    eachSystem [ system.x86_64-linux ] (system:
+    eachSystem defaultSystems (system:
```
ulrikstrid, line 27: This doesn't really belong in this PR but it expands to

It's basically the same as
mac_x86
mac_aarch64
linux_x86
linux_aarch64

src/metrics/blocks.ml
```diff
@@ -0,0 +1,14 @@
...
+open Config
+
+let blocks_produced_total =
+  let help = "Total number of blocks produced" in
+  Counter.v ~help ~namespace ~subsystem "blocks_produced_total"
+
+let blocks_signed_total =
+  let help = "Total number of blocks signed" in
+  Counter.v ~help ~namespace ~subsystem "blocks_signed_total"
+
```
renatoalencar, line 11: I think it'd be useful to also have total transactions received and included, so we would be able to measure throughput and delay for a transaction to be confirmed.

ulrikstrid: The goal of this PR is to get the bare minimum to just have dashboards. Please add this to upcoming PRs :smile:
Would also love it if @gsebil08 and @rosalogia could add some metrics to the pollinate library that we can expose via deku.

sandbox.sh
aguillon: I'm ok with this change but we should probably move it to a proper configuration file one day.

src/node/flows.ml
aguillon: Yup.

aguillon (approved):
I'll merge if @d4hines approves.

d4hines (approved):
Thanks for this @ulrikstrid, LGTM

src/metrics/blocks.ml
rosalogia: Sounds like a great idea :smile:

PR #527, opened 2022-04-04
[d4hines] Dummy-ticket

d4hines:
## Problem

There's no easy way to test ticket deposits in Deku.

## Solution

This PR adds
- A dummy contract that forges tickets and deposits them to Deku.
- A command to sandbox.sh for deploying it.
- A command to sandbox.sh for automatically doing the deposit.

### Stacked PR Chain: d4hines
|#558|Simplify Sandbox.sh||-|
|#527|👉Dummy-ticket||#558|
|#536|feat: add block-by-level rpc||#527|
|#537|Add block rate and operation rate metrics||#536|
|#539|Benchmarking: ticket transfers||#537|
|#540|*(Draft)POC: Parametric Deku*||#539|


dummy_ticket.mligo
```diff
@@ -0,0 +1,33 @@
...
+    (* Content of ticket is just a packed int *)
+    let ticket = Tezos.create_ticket (Bytes.pack counter) 1n in 
+    let param  = { ticket = ticket; address = deku_recipient } in
+    let op = Tezos.transaction param 0mutez contract in
+    (* We increment the counter to add some variety to the data *)
+    [op], (counter + 1, stored_tickets)
+   | None -> failwith "Entrypoint doesn't exist")
+  | Withdraw ticket ->
+    let ops : operation list = [] in
+    ops , (counter, ticket :: stored_tickets)
```
aguillon, line 32: Just to be sure: calling this contract's "deposit" entrypoint amounts to creating a ticket belonging to `deku_recipient`, who will ultimately be the dummy_ticket contract, and sending it to the Deku vault contract. This ticket only contains the current counter value. Correct?

Withdrawing a ticket should only be done through `deku_cli withdraw`. Right?

sandbox.sh
```diff
@@ -194,6 +194,28 @@ start_deku_cluster() {
   done
 }
 
+deploy_dummy_ticket() {
+  contract=$(ligo compile contract ./dummy_ticket.mligo)
+  tezos-client --endpoint $RPC_NODE originate contract "dummy_ticket" \
+    transferring 0 from myWallet \
```
aguillon, line 200: This may be a bit late to notice, but: where does the alias `myWallet` come from? I didn't see it in the docs.

dummy_ticket.mligo
d4hines: Not quite.

The `Deposit` entrypoint of dummy ticket contract forges a ticket, and then calls the consensus contract's `deposit` entrypoint with the parameters. ` {ticket = ticket; address = deku_recipient}`. At that point, you'll have to read `consensus.mligo` to see what it does with it, but I can summarize it by saying it locks the ticket in a vault and tags it as being owned by the `address` specified. When the deku nodes hear this transaction, they forge the ticket on the Deku side of things, and the address originally listed in `deku_recipient` can now use the asset. 

Withdraw is more complex, please see https://hackmd.io/L74tCg4cTjq9AO7JIQq2BA

sandbox.sh
d4hines: `myWallet` is created during `./sandbox.sh setup`.

d4hines: There is an implicit dependency graph in the commands - not every order you could run them in works. Right now I think it's best if our team just understands this dependency order (it's very intuitive if you understand what the commands are doing), and runs them manually. We should eventually have a `Makefile` like dependency graph for tasks, but I think we should rewrite `./sandbox.sh` in OCaml before we do that.

d4hines: I say this because if you run `./sandbox.sh deploy-dummy-ticket` before running `./sandbox.sh setup` you will get an error.

d4hines:
@aguillon, when I actually used this contract for benchmarking, I founds to be less useful than I thought, so I added a new commit that I believe improves them.

anchpop (approved):
LGTM! #450 modifies ./sandbox.sh to include a function for deploying contracts, once that gets merged it would be nice to reuse it for deploying the dummy_ticket.mligo

lykimq:
In the `sandbox.sh` at the function 
```
assert_deku_state() {
  local current_state_hash
  current_block_height=$(deku_state_hash)
```

the function for `current_block_height` I think it should be `current_state_hash`?

ulrikstrid:
> Regarding adding this to CI e2e tests: do you think it's possible that it fails sometimes, as several operations are running together with the same private key and share the same counter? If yes, it would help to have distinct private keys for deku hash saving and the dummy ticket, right?

Doesn't the e2e test start flextesa so it's a sandboxed tezos that doesn't know about each other?

aguillon:
I'm stuck here for now: it works fine on my system, but I can't reproduce the CI because it uses nix, which asks me to download several GB of stuff, which I can't do for the moment.

ulrikstrid:
Adding this:
```diff
diff --git a/src/node/flows.ml b/src/node/flows.ml
index 40d8dacd..78925ddf 100644
--- a/src/node/flows.ml
+++ b/src/node/flows.ml
@@ -415,6 +415,11 @@ let register_uri state update_state ~uri ~signature =
       } in
   Ok ()
 let request_withdraw_proof state ~hash =
+  let () =
+    if BLAKE2B.Map.is_empty state.Node.recent_operation_receipts then
+      print_endline "recent_operation_receipts is empty"
+    else
+      print_endline "recent_operation_receipts has receipts" in
   match state.Node.recent_operation_receipts |> BLAKE2B.Map.find_opt hash with
   | None -> Network.Withdraw_proof.Unknown_operation
   | Some (Receipt_tezos_withdraw withdrawal_handle) ->
```
Will have it print `recent_operation_receipts is empty` when running the test, I'm surprised this works anywhere and seems like a bug in deku

aguillon:
The ticket deposit on Tezos is not seen by Deku, hence the test fails on the CI. I have no idea what's going on, I tried upgrading Taquito (why not). I can't run the nix version locally for now because it wants to download too much stuff.

src/protocol/protocol.ml
```diff
@@ -22,6 +22,9 @@ let apply_core_tezos_operation state tezos_operation =
   in
   let core_state =
     Core.State.apply_tezos_operation state.core_state tezos_operation in
+  Printf.eprintf "Received Tezos operation, core state is now:\n%!";
+  Printf.eprintf "%s\n%!"
+    (Core.State.to_yojson state.core_state |> Yojson.Safe.to_string);
```
aguillon, line 27: remove me

sandbox.sh
```diff
@@ -385,10 +388,67 @@ start_tezos_node() {
...
+
+  sleep 10
+
+  echo "{\"address\": \"$DEKU_ADDRESS\", \"priv_key\": \"$DEKU_PRIVATE_KEY\"}" > wallet.json
+
+  DUMMY_TICKET=$(tezos-client show known contract dummy_ticket | tr -d '\t\n\r')
+
+  # # We can withdraw 10 tickets from deku
+  OPERATION_HASH=$(deku-cli withdraw data/0 ./wallet.json "$DUMMY_TICKET" 10 "Pair \"$DUMMY_TICKET\" 0x" | awk '{ print $2 }' | tr -d '\t\n\r')
+  echo "OPERATION HASH IS $OPERATION_HASH"
```
aguillon, line 433: remove me

sandbox.sh
```diff
@@ -385,10 +388,67 @@ start_tezos_node() {
...
+deposit_withdraw_test() {
+  # Deposit 100 tickets
+  deposit_ticket | grep tezos-client | tr -d '\r'
+  echo "Checking that the deposit has been made:"
+  TICKET_ADDRESS="$(tezos-client --endpoint $RPC_NODE show known contract dummy_ticket | grep KT1 | tr -d '\r')"
+  vault_big_table=$(tezos-client --endpoint $RPC_NODE get contract storage for consensus | tail -n 1 | tr -d " \r")
+  hash=$(tezos-client hash data "Pair \"$TICKET_ADDRESS\" 0x" of type "pair address bytes" | sed -n "s/^Script-expression-ID-Hash: \(.*\)$/\1/p" | tr -d '\r')
+  echo "Hash is $hash"
+  vault=$(tezos-client get element "$hash" of big map "$vault_big_table")
+  echo "vault is $vault"
```
aguillon, line 423: remove me

src/protocol/protocol.ml
ulrikstrid: Could use `Logs.debug` to have an easy way of re-adding them back if it's needed

aguillon:
Finally. Thank you @ulrikstrid!

PR #528, opened 2022-04-04
chore: add nixfmt ci

d4hines:
## Problem

nix code formatting isn't checked.

## Solution

Check formatting in CI, just like ocaml code.

We may want to use e.g. https://github.com/numtide/treefmt in the future.


ulrikstrid (approved):

ulrikstrid:
#524 fixes the formatting so we should merge that first

gsebil08 (approved):

PR #529, opened 2022-04-04
add auto-build to ./sandbox.sh

d4hines:
## Problem

You don't want to run `./sandbox.sh` with old code - you'll inevitably get confused and lose time debugging.

Esy users dodge this because `esy x` rebuilds the environment. Nix users hit this.

## Solution

I'm not sure if this is a great solution, but it's simple enough. What do you guys think?


d4hines:
I tested `time ./sandbox.sh` before and after the change, and with a warm cache it seems to add less than 0.5 seconds of overhead.

aguillon (approved):

PR #530, opened 2022-04-05
Tendermint: running the thing

aguillon:
## Depends

- [ ] #434 

## Problem

The consensus algorithm is now triggered by other consensus messages, so we have to send the first one manually.

Also, please note that Tendermint requires at least 4 validators (see changes to `sandbox.sh` in the previous PR) to run properly.

## Solution

We add the right handler to `deku_node`. As the one defined in flows.ml explicitly uses Lwt, we have to define a new generic handler function. Looks good though? Should we switch all the existing, result-based ones?

We also add a `start-consensus` command to the CLI, as well as a small script to start the consensus.

I recommend the following workflow:
- open 4 terminals
- start 3 deku instances on data/0, data/1, data/2
- run `start_consensus.sh`
- watch them struggle forward (be sure to wait long enough, like 2 minutes or so)
- start the remaining deku instance on data/3. Does it catch the others? Is it better?

## Reviews and future works

You can change the timeouts for each step in `tendermint_internals`. It's probably better if you want to test stuff.

Please make sure that you kill and start nodes, and that the nodes can recover from being killed (both by reading the past state from the disk and from catching the others).

Please note that, if you kill 2 (on a total of 4) nodes and the consensus stops, you'll have to manually restart it. This is likely *the intended behaviour*, as a freshly reconnected Tendermint node expects the networking layers (of other nodes) to send the missed messages… and the current HTTP implementation does not take care of this.

### DIFF
--- a/src/node/tendermint.ml
+++ b/src/node/tendermint.ml
@@ -1,0 +1,206 @@
+open Helpers
+open Tendermint_internals
+open Tendermint_data
+open Tendermint_processes
+open Tendermint_helpers
+
+type t = {
+  identity : State.identity;
+  clocks : clock list IntSet.t;
+  consensus_states : consensus_state IntSet.t;
+  (* TODO: make non-mutable *)
+  mutable procs : (height * process) list;
+  node_state : State.t;
+  input_log : input_log;
+  output_log : OutputLog.t;
+}
+
+type should_restart_tendermint =
+  | DontRestart
+  | RestartAtHeight of height
+  | RestartAtRound  of round
+
+let make node_state current_height =
+  let identity = node_state.State.identity in
+  let clocks = IntSet.create 0 in
+  let new_state = fresh_state current_height in
+  let states = IntSet.create 0 in
+  IntSet.add states current_height new_state;
+  let procs = List.map (fun x -> (current_height, x)) all_processes in
+  let input_log = empty () in
+  let output_log = OutputLog.empty () in
+  {
+    identity;
+    clocks;
+    consensus_states = states;
+    node_state;
+    procs;
+    input_log;
+    output_log;
+  }
+
+let current_height node =
+  List.fold_left max 0L (IntSet.to_seq_keys node.consensus_states |> List.of_seq)
+
+(* FIXME: bad design *)
+let () =
+  produce_value := fun state -> block (Building_blocks.produce_block state)
+
+(** Process messages in the queue and decide actions; pure function, to be interpeted in Lwt later.
+    TODO: Ensures that messages received over network go through signature verification before adding them to input_log
+    FIXME: we're not empyting the input_log atm *)
+let tendermint_step node =
+  let open Tendermint_processes in
+  let rec exec_procs processes still_active network_actions =
+    match processes with
+    | ((height, process) as p) :: rest -> begin
+      let consensus_state = IntSet.find node.consensus_states height in
+      let round = consensus_state.Tendermint_internals.round in
+      let message_log = node.input_log in
+      let output = node.output_log in
+      match
+        process height round consensus_state message_log output node.node_state
+      with
+      (* The process precondition hasn't been activated *)
+      | None -> exec_procs rest (p :: still_active) network_actions
+      (* The process terminates with a network event *)
+      | Some (Broadcast t) -> exec_procs rest still_active (t :: network_actions)
+      (* The process terminates silently *)
+      | Some DoNothing -> exec_procs rest still_active network_actions
+      (* We accepted a block for the height *)
+      | Some (RestartTendermint (height, 0)) ->
+        (* Start new processes and forget about the older ones *)
+        (* TODO: this is only valid for the (current) restricted version of Tendermint which does not
+           support several cycles/heights running in parallel. *)
+        ([], [], RestartAtHeight height)
+      | Some (RestartTendermint (_height, round)) ->
+        ([], network_actions, RestartAtRound round)
+      (* Add a new clock to the scheduler *)
+      | Some (Schedule c) ->
+        let cs =
+          IntSet.find_opt node.clocks height |> Option.value ~default:[] in
+        IntSet.add node.clocks height (c :: cs);
+        exec_procs rest still_active network_actions
+    end
+    | [] -> (still_active, network_actions, DontRestart) in
+  let still_active, network_actions, should_restart =
+    exec_procs node.procs [] [] in
+  (* TODO: make non-mutable *)
+  node.procs <- still_active;
+  ({ node with procs = still_active }, List.rev network_actions, should_restart)
+(* List.rev: Do we really need order on the network? *)
+
+let add_to_input input_log height step content =
+  let index = (height, step) in
+  add input_log index content
+
+let is_valid_consensus_op state consensus_op =
+  (* TODO: ConsensusStep2 this is a filter for input log since it's only optimization (don't keep stuff from past)*)
+  let open Result in
+  let _all_operations_properly_signed = function
+    | _ -> true in
+  let h = height consensus_op in
+  let current_height = state.State.protocol.block_height in
+  if current_height > h then
+    let s =
+      Printf.sprintf
+        "new block has a lower height (%Ld) than the current state (%Ld)" h
+        current_height in
+    error s
+  else
+    ok ()
+
+(** Hashes stuff and broadcast the corresponding consensus operation *)
+let broadcast_op state consensus_op =
+  (* We need two hashes: one corresponding to the block, height and round, which will ultimately get
+     posted on Tezos; and one of the consensus operation *)
+  let node_address = state.State.identity.t in
+  let secret_key = state.State.identity.secret in
+  (* Hashing the operation *)
+  let operation_hash = hash_of_consensus_op consensus_op node_address in
+  let operation_signature =
+    Protocol.Signature.sign ~key:secret_key operation_hash in
+  (* Hash the value+height+round even if it's nil *)
+  let block_hash = hash_of_consensus_value consensus_op in
+  let block_signature = Protocol.Signature.sign ~key:secret_key block_hash in
+  Lwt.async (fun () ->
+      let%await () = Lwt_unix.sleep 0.3 in
+      Networking.broadcast_consensus_op state
+        {
+          operation = consensus_op;
+          sender = node_address;
+          hash = block_hash;
+          block_signature;
+          operation_signature;
+        })
+
+let add_consensus_op node sender op =
+  let input_log = node.input_log in
+  let input_log =
+    add_to_input input_log (height op) (step_of_op op) (content_of_op sender op)
+  in
+  Lwt.return { node with input_log }
+
+let rec exec_consensus node =
+  let node, network_actions, should_restart = tendermint_step node in
+
+  (* Send all actions over the network *)
+  List.iter (broadcast_op node.node_state) network_actions;
+  match (node.procs, should_restart) with
+  | _, RestartAtRound r ->
+    let cur_height = current_height node in
+    let cur_state = IntSet.find node.consensus_states cur_height in
+    cur_state.round <- r;
+    let new_processes = List.map (fun p -> (cur_height, p)) all_processes in
+    exec_consensus { node with procs = new_processes }
+  | [], RestartAtHeight new_height ->
+    let new_state = fresh_state new_height in
+    IntSet.add node.consensus_states new_height new_state;
+    let new_processes = List.map (fun p -> (new_height, p)) all_processes in
+    { node with procs = new_processes }
+  | _ ->
+    (* Start the clocks *)
+    IntSet.map_inplace
+      (fun _height clocks ->
+        List.map
+          (fun c -> start_clock node c.Clock.height c.Clock.round c)
+          clocks)
+      node.clocks;
+    node
+
+(* TODO: don't use async and global state,
+   maintain an internal list of clocks started at a given time and periodically call exec_consensus
+   from flows again? *)
+
+and start_clock node clock_height clock_round clock =
+  let open Lwt in
+  let open Tendermint_processes in
+  if clock.Clock.started then
+    clock
+  else begin
+    async (fun () ->
+        Lwt_unix.sleep (float_of_int clock.Clock.time) >>= fun () ->
+        (* Checks that the clock is still relevant *)
+        let current_height = current_height node in
+        (* FIXME: fragile code; we should fix the whole way we handle clocks with Lwt *)
+        let current_round =
+          (IntSet.find node.consensus_states current_height)
+            .Tendermint_internals.round in
+        if current_height > clock_height || current_round > clock_round then
+          Lwt.return_unit
+        else
+          let input_log = node.input_log in
+          let _ = add_to_input input_log clock_height clock.Clock.step Timeout in
+          let new_node = exec_consensus node in
+          (* TODO: make non-mutable *)
+          node.procs <- new_node.procs;
+          Lwt.return_unit);
+    { clock with started = true }
+  end
+
+let make_proposal height round b = ProposalOP (height, round, block b, -1)
+
+let get_block_opt cstate height =
+  match OutputLog.get cstate.output_log height with
+  | Some (Block b, round) -> Some (b, round)
+  | _ -> None
--- a/src/node/tendermint_data.ml
+++ b/src/node/tendermint_data.ml
@@ -1,0 +1,284 @@
+open Crypto
+open State
+open Tendermint_helpers
+open Tendermint_internals
+
+type elmt = value * round
+
+module MySet = Set.Make (struct
+  type t = elmt
+
+  let compare (v1, r1) (v2, r2) = compare (v1, r1) (v2, r2)
+end)
+
+type node_identifier = Key_hash.t
+
+type index_step = Tendermint_internals.consensus_step
+
+type index = height * index_step
+
+type proposal_content = {
+  process_round : round;
+  proposal : value;
+  process_valid_round : round;
+  sender : node_identifier;
+}
+
+type prevote_content = {
+  process_round : round;
+  repr_value : value;
+  sender : node_identifier;
+}
+
+type precommit_content = prevote_content
+
+type content =
+  | ProposalContent  of proposal_content
+  | PrevoteContent   of prevote_content
+  | PrecommitContent of precommit_content
+  | Timeout
+
+(* TODO: ensure PrevoteOP and PrecommitOP contains repr_values and not values. *)
+let content_of_op sender = function
+  | Tendermint_internals.ProposalOP
+      (_, process_round, proposal, process_valid_round) ->
+    ProposalContent { process_round; proposal; process_valid_round; sender }
+  | Tendermint_internals.PrevoteOP (_, process_round, repr_value) ->
+    PrevoteContent { process_round; repr_value; sender }
+  | Tendermint_internals.PrecommitOP (_, process_round, repr_value) ->
+    PrecommitContent { process_round; repr_value; sender }
+
+(* TODO: ensure there is no data duplication in input_log *)
+type input_log = {
+  msg_log : (index, content list) Hashtbl.t;
+  timeouts : (index, content) Hashtbl.t;
+      (* Making timeouts a separate queue because it's easier at the moment... *)
+}
+let empty () : input_log =
+  { msg_log = Hashtbl.create 0; timeouts = Hashtbl.create 0 }
+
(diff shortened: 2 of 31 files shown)


src/node/tendermint.ml
```diff
@@ -0,0 +1,206 @@
...
+      | Some (Broadcast t) -> exec_procs rest still_active (t :: network_actions)
+      (* The process terminates silently *)
+      | Some DoNothing -> exec_procs rest still_active network_actions
+      (* We accepted a block for the height *)
+      | Some (RestartTendermint (height, 0)) ->
+        (* Start new processes and forget about the older ones *)
+        (* TODO: this is only valid for the (current) restricted version of Tendermint which does not
+           support several cycles/heights running in parallel. *)
+        ([], [], RestartAtHeight height)
+      | Some (RestartTendermint (_height, round)) ->
```
gsebil08, line 76: Where is this `_height` coming from?

gsebil08 (commented):
Massive work! Asked some questions...

src/node/tendermint_processes.ml
```diff
@@ -0,0 +1,367 @@
...
+    None
+
+let start_round (height : height) (round : round)
+    (consensus_state : consensus_state) (_msg_log : input_log) _dlog
+    global_state =
+  consensus_state.round <- round;
+  consensus_state.step <- Proposal;
+  (* If we start a round > 0 and already have a valid/locked value, we update their hash
+     with the correct round *)
+  (if consensus_state.valid_value <> nil then
```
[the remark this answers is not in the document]
aguillon: `valid_value` and `locked_value` only make sense when we failed last round but were already "precommitted" to a value. As this value is carried over for the new round, we need to update its round (see block.ml) and its hash in case it gets committed. If these values are nil, there's nothing to update.

src/node/tendermint_processes.ml
```diff
@@ -0,0 +1,367 @@
...
+                  consensus_state.valid_value,
+                  consensus_state.valid_round )))
+      else
+        Some
+          (Broadcast
+             (ProposalOP
+                ( consensus_state.height,
+                  consensus_state.round,
+                  (* FIXME: remove depdendency to full global_state *)
+                  !produce_value global_state,
```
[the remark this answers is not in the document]
aguillon: In case we're the producer for this round but there's no `valid_value` from last round (e.g., we're in a new round), we have to produce one. 

Sidenote: if you observe the algorithm, there shouldn't be a case where we have a valid_value that isn't nil, and where locked_value is. @AngryStitch this is an opportunity for a test! :p

src/node/tendermint_processes.ml
```diff
@@ -0,0 +1,367 @@
...
+      DoNothing in
+  if contains_timeout msg_log height Precommit then
+    Some (do_something consensus_state)
+  else
+    None
+
+let response_to_proposal (height : height) (round : round)
+    (consensus_state : consensus_state) (msg_log : input_log) _dlog global_state
+    =
+  let _height = consensus_state.height in
```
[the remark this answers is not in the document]
aguillon: Yes lol, thank you. They're most likely the remains of a test, as they should be equal to `height` and `round` (the arguments).

src/node/tendermint_processes.ml
```diff
@@ -0,0 +1,367 @@
...
+  let common_set = extract_common_set left_set right_set in
+  if
+    left_set = MySet.empty
+    || right_set = MySet.empty
+    || step <> Proposal
+    || common_set = MySet.empty
+  then
+    None
+  else
+    Some (do_something consensus_state common_set)
```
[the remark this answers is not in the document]
aguillon: Which one?

src/node/tendermint_processes.ml
```diff
@@ -0,0 +1,367 @@
...
+
+let prepare_default_precommit_prevote_phase (height : height) (round : round)
+    (consensus_state : consensus_state) (msg_log : input_log) _dlog global_state
+    =
+  let take_all_prevotes = function
+    | PrevoteContent content when content.process_round = consensus_state.round
+      ->
+      Some ((nil, content.process_round), get_weight global_state content.sender)
+    | PrevoteContent _ -> None
```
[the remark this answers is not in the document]
AngryStitch: I don't get the question. `count_prevotes` takes a `prevote_selection` that is defined here where what matters is that round match.

src/node/tendermint_processes.ml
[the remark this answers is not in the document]
AngryStitch: No? PrevoteTimeout are triggered when you reach 2 f + 1 prevotes (no matter the value) at (h, r)

src/node/tendermint_data.ml
```diff
@@ -0,0 +1,284 @@
...
+  let prevote_selection =
+    match prevote_selection with
+    | None -> (
+      function
+      | PrevoteContent content
+        when content.process_round = consensus_state.round ->
+        Some
+          ( (content.repr_value, content.process_round),
+            get_weight global_state content.sender )
+      | PrevoteContent _ -> None
```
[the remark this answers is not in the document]
AngryStitch: Actually, @aguillon I think there might be an issue. 
In `precommit_failed_previous_round` I think we are supposed to match `vr` but `count_prevotes` enforces a match on `process_round`, that's an issue?

src/node/tendermint.ml
AngryStitch: In `tendermint_processes.ml` we have:
```
type consensus_action =
  | Broadcast         of sidechain_consensus_op
  | Schedule          of consensus_action option Clock.t
  | RestartTendermint of (height * round)
  | DoNothing
  ```
  
  Here, we discard `_height`

File: src/node/tendermint_processes.ml (hunk 3 above)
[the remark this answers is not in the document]
Comment (AngryStitch) [reply src/node/tendermint_processes.ml:122]: Yes. I'm gonna enforce stuff on the new consensus repo. Thanks :)

File: src/node/tendermint_data.ml [hunk 8]
```diff
@@ -0,0 +1,284 @@
...
+}
+let empty () : input_log =
+  { msg_log = Hashtbl.create 0; timeouts = Hashtbl.create 0 }
+
+let add (input_log : input_log) (index : index) (c : content) =
+  match c with
+  | Timeout ->
+    Hashtbl.replace input_log.timeouts index c;
+    input_log
+  | c -> (
```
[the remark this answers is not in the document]
AngryStitch: Nope, it needs to be `c`, take a look at line 70:
```
match List.find_opt (fun elem -> elem = c) previous with
```

we're using `c`

src/node/tendermint_data.ml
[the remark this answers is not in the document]
aguillon: Indeed, will fix later. Thanks!

src/node/tendermint_data.ml
```diff
@@ -0,0 +1,284 @@
...
+
+let remove (input_log : input_log) (index : index) =
+  Hashtbl.remove input_log.msg_log index;
+  Hashtbl.remove input_log.timeouts index
+
+let prune (input_log : input_log) (height : height) =
+  let _ =
+    [Proposal; Prevote; Precommit]
+    (* Should we make sure all h < height are emptied here? *)
+    |> List.map (fun step -> (height, step))
```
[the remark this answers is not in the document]
AngryStitch: There are 3 steps: `Proposal`, `Prevote`, `Precommit`.
For a given `height` we re-create each possible `index` (`type index = height * index_step`) and remove from `input_log` stuff at that index

d4hines:
During testing, the chain got stuck in a situation where I don't think it should have:
- start 3 nodes, start_consensus.sh
- wait a while, then start 4th node
- wait a while, kill block producer (node 0)
- wait a while, bring node 0 back online
- wait a while, kill node 2

On the last step, block production stopped.

src/protocol/validators.ml
```diff
@@ -52,3 +52,18 @@ let remove validator t =
...
+  let n = List.length validators in
+  (* n should stay small enough *)
+  let i1 = Int64.rem height (Int64.of_int n) in
+  let i1 = Int64.to_int i1 in
+  let i = (i1 + (round mod n)) mod n in
+  List.nth validators i
+
+(** Verifies if someone is a validator as required in Tendermint consensus. *)
+let is_validator t (x : Key_hash.t) =
+  let addresses = List.map (fun v -> v.address) t.validators in
```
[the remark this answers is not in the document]
d4hines: @EduardoRFS am I wrong on this?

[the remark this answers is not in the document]
d4hines: Also if you use (polymorphic) structural equality, (e.g. List.mem), @Zett98 is going to come after you.

src/node/tendermint_internals.mli
```diff
@@ -0,0 +1,94 @@
...
+type round = int [@@deriving yojson]
+(** At a specific (chain) height, Tendermint's consensus algorithm may run several rounds.*)
+
+(** Tendermint sometimes decides on a `Nil` value. *)
+type value =
+  | Block of Protocol.Block.t
+  | Nil
+[@@deriving yojson]
+
+val repr_of_value : value -> value
```
d4hines, line 14: This function does not make sense. As it is, it is pointless.

src/node/tendermint_internals.mli
```diff
@@ -0,0 +1,94 @@
...
+val is_valid : State.t -> value -> bool
+(** Ensures a value going through Tendermint is valid. TODO: ConsensusStep2 actually verify signatures, logging system. *)
+
+val on_block : nil_default:'a -> (Protocol.Block.t -> 'a) -> value -> 'a
+
+val block : Protocol.Block.t -> value
+
+val nil : value
+
+val update_value : value -> round -> value
```
d4hines, line 29: Hmmmm. Maybe `value -> round -> value result`? May change my mind when I better understand it.

src/node/tendermint_internals.mli
```diff
@@ -0,0 +1,94 @@
...
+
+type consensus_step =
+  | Proposal
+  | Prevote
+  | Precommit
+      (** Tendermint's consensus goes through 3 steps at each round. Used inside a consensus instance in a node. *)
+
+val string_of_step : consensus_step -> string
```
d4hines, line 37: Use `deriving show`

src/node/tendermint_internals.mli
```diff
@@ -0,0 +1,94 @@
...
+
+(** Tendermint's consensus step-communication with other nodes. Two first fields are height and
+    round of when the operation was sent. *)
+type sidechain_consensus_op =
+  | ProposalOP  of (height * round * value * round)
```
d4hines, line 42: Nitpick: `ProposalOP` expands to 3 words: `Proposal O* P*`. I'm guessing this is short for `ProposalOperation`, in which case it should be `ProposalOp`. Or just `ProposalOperation`.

src/node/tendermint_internals.mli
```diff
@@ -0,0 +1,94 @@
...
+  | PrevoteOP   of (height * round * value)
+  | PrecommitOP of (height * round * value)
+[@@deriving yojson]
+
+val step_of_op : sidechain_consensus_op -> consensus_step
+
+val string_of_op : sidechain_consensus_op -> string
```
d4hines, line 49: `deriving show` would be nice again.

src/node/tendermint_data.ml
```diff
@@ -0,0 +1,284 @@
...
+    |> List.map (fun index -> remove input_log index) in
+  ()
+
+let map_option f ls =
```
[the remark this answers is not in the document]
Zett98: This is List.filter_map

src/node/tendermint_processes.ml
```diff
@@ -0,0 +1,367 @@
...
+    else
+      Some
+        (Schedule
+           (Clock.make proposal_timeout Proposal height round
+              (on_timeout_propose height round))) in
+  return_action
+
+let on_timeout_precommit (height : height) (round : round)
+    (consensus_state : consensus_state) (msg_log : input_log) _dlog
+    _global_state =
```
Zett98, line 133: is _dlog and _global_state useless?

src/node/tendermint_processes.ml
```diff
@@ -0,0 +1,367 @@
...
+        (PrecommitOP (consensus_state.height, consensus_state.round, nil)))
+    else
+      DoNothing in
+  if contains_timeout msg_log height Prevote then
+    Some (do_something consensus_state)
+  else
+    None
+
+let start_round (height : height) (round : round)
+    (consensus_state : consensus_state) (_msg_log : input_log) _dlog
```
Zett98, line 93: _msg_log and _dlog?

src/node/tendermint_processes.ml
```diff
@@ -0,0 +1,367 @@
...
+    else
+      DoNothing in
+  if contains_timeout msg_log height Proposal then
+    Some (do_something consensus_state)
+  else
+    None
+
+let on_timeout_prevote (height : height) (round : round)
+    (consensus_state : consensus_state) (msg_log : input_log) _dlog
+    _global_state =
```
Zett98, line 75: _dlog and _global_state

src/node/tendermint_internals.ml
```diff
@@ -0,0 +1,158 @@
...
+
+let repr_of_value v = v
+
+let produce_value : (State.t -> value) ref = ref (fun _ -> assert false)
+
+let is_valid state value =
+  match value with
+  | Nil -> false
+  | Block block ->
+    let all_operations_properly_signed _block = true in
```
Zett98, line 23: ?

src/node/tendermint_processes.ml
```diff
@@ -0,0 +1,367 @@
...
+      global_state in
+  let extract_valid_set (found_set : MySet.t) =
+    MySet.filter (fun (_, r) -> r = round) found_set in
+  let valid_set = extract_valid_set found_set in
+  if valid_set = MySet.empty || step <> Proposal then
+    None
+  else
+    Some (do_something consensus_state valid_set)
+
+let precommit_failed_previous_round (_height : height) (_round : round)
```
Zett98, line 172: _round _height _dlog

Zett98 (commented):
Take a quick look, Also please don't use Polymorphic comparisons so much aka `=` and `<>`. 
Please remove unused arguments and use named args when `num of args >= 4`

src/node/tendermint_internals.mli
AngryStitch: Yes, as it is it's pointless. However, it shouldn't be. Tendermint is meant to use `repr_of_value` and not the actual value, as a way to minimize what's sent on the network. We used `repr_of_value` in the code to remind ourselves of that so we'll implement it in the future.

src/node/tendermint.ml
```diff
@@ -0,0 +1,206 @@
...
+(* List.rev: Do we really need order on the network? *)
+
+let add_to_input input_log height step content =
+  let index = (height, step) in
+  add input_log index content
+
+let is_valid_consensus_op state consensus_op =
+  (* TODO: ConsensusStep2 this is a filter for input log since it's only optimization (don't keep stuff from past)*)
+  let open Result in
+  let _all_operations_properly_signed = function
```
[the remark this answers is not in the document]
AngryStitch: As of right now, yes. However something should be done here, hence the `TODO: ConsensusStep2`

src/node/tendermint_processes.ml
AngryStitch: Here, yes. However we need all the tendermint process functions to have the same signature.

src/node/tendermint_processes.ml
AngryStitch: Unused here, however, we need all the Tendermint process functions to have the same signature.

src/node/tendermint_processes.ml
AngryStitch: Unused here, however, we need all the Tendermint process functions to have the same signature.

src/node/tendermint_processes.ml
AngryStitch: Unused here, however, we need all the Tendermint process functions to have the same signature.

src/node/tendermint_processes.ml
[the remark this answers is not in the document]
gsebil08: I didn't know the `when` syntax right before, so I was asking if this pattern was useful because I thought line222 would be the same as line219.
My bad.

src/node/tendermint_processes.ml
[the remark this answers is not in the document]
gsebil08: Thanks, totally forgot the `2 f + 1`! Make sens.

src/node/tendermint_internals.mli
d4hines: Yeah, Ulrik beat me to it, but what I think you want is an abstract type `repr`. Then `repr_of_value` makes total sense. Under the hood, you can implement it as the identity function, and later you can change it to be more efficient without changing the signature at all.

src/node/tendermint_internals.ml
d4hines: My recommendation is a `FIXME:` comment to preempt our internal alarm bells.

src/node/tendermint_internals.ml
```diff
@@ -0,0 +1,158 @@
...
+type height = int64 [@@deriving yojson]
+
+type round = int [@@deriving yojson]
+
+type consensus_step =
+  | Proposal
+  | Prevote
+  | Precommit
+
+let string_of_step = function
```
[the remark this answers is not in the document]
d4hines: +1

src/node/tendermint_internals.ml
```diff
@@ -0,0 +1,158 @@
...
+  | Proposal -> "Proposal"
+  | Prevote -> "Prevote"
+  | Precommit -> "Precommit"
+
+type sidechain_consensus_op =
+  | ProposalOP  of (height * round * value * round)
```
d4hines, line 54: There are two values of the same type. This is begging for them to be confused and then misused. I suggest a record instead of a tuple.

src/node/tendermint_internals.ml
```diff
@@ -0,0 +1,158 @@
...
+  | Block b -> b.Block.hash
+
+let hash_of_consensus_op consensus_op sender =
+  let s1 = string_of_op consensus_op in
+  let s2 = Crypto.Key_hash.to_string sender in
+  Crypto.BLAKE2B.hash (s1 ^ s2)
+
+let hash_of_consensus_value consensus_op =
+  let _, _, value =
+    match consensus_op with
```
d4hines, line 103: this should just be `let hash_of_consensus_value consensus_op = value_of_op consensus_op |> hash_of_value`, right?

src/node/tendermint_internals.ml
```diff
@@ -0,0 +1,158 @@
...
+let debug state msg =
+  let self = Crypto.Key_hash.to_string state.State.identity.t in
+  let self = String.sub self (String.length self - 6) 6 in
+  prerr_endline ("*** " ^ self ^ "   " ^ msg)
+
+let is_allowed_proposer (global_state : State.t) (height : height)
+    (round : round) (address : Key_hash.t) =
+  let protocol_state = global_state.protocol in
+  let proposer = proposer protocol_state.validators height round in
+  let b = Key_hash.equal address proposer.address in
```
d4hines, line 144: I don't believe `let b = <bool expr> in b` is any more meaningful than simply `<bool expr>` in this case.

src/node/tendermint_internals.mli
```diff
@@ -0,0 +1,94 @@
...
+val is_allowed_proposer : State.t -> height -> round -> Key_hash.t -> bool
+(** Verifies that the Key_hash.t matches the one expected by protocol for this height and round. *)
+
+val i_am_proposer : State.t -> height -> round -> bool
+(** True if the node is the proposer for this (height, round) according to proposer selection. *)
+
+val get_weight : State.t -> Key_hash.t -> int
+(** Implemented as Proof-of-Authority. *)
+
+val proposal_timeout : int
```
d4hines, line 92: units of time would be helpful here. Is this seconds, minutes, etc?

src/node/tendermint_data.mli
```diff
@@ -0,0 +1,115 @@
...
+open Tendermint_internals
+
+type elmt = value * round
+
+module MySet : Set.S with type elt := elmt
+
+(* FIXME: ConsensusStep2 Tendermint we could do better *)
+type node_identifier = Key_hash.t
+
+type index_step = consensus_step
```
d4hines, line 14: I really prefer to avoid type aliases unless it's for the purpose of abstraction. Why add the extra mental pointer-lookup?

PR #531, opened 2022-04-05
Add core_bench + reqs to esy.json

InfiniteSwerve:
## Problem

We want to use [core_bench](https://github.com/janestreet/core_bench) to create benchmarks for deku.

## Solution

This PR adds  core_bench and it's requirements to esy.json. It also changes the name of the local library 'core' to 'deku_core' to avoid conflicts with the janestreet library core that core_bench requires.


aguillon:
As you listed it as a resolution, could you explain https://github.com/InfiniteSwerve/core/commit/9c8f9f90ac74fe33aae13c7c040951a5ab78b064 please? Also, are all those resolutions really necessary?

esy.json
```diff
@@ -186,6 +190,80 @@
...
+    "@opam/textutils_kernel": "janestreet/textutils_kernel:textutils_kernel.opam#99e6218f33e51d37bbed9ab620ec8260ad89a66d",
+    "@opam/core_kernel": "janestreet/core_kernel:core_kernel.opam#331c34e99e3f15116f9f0cd6235f1d016adca6d7",
+    "@opam/int_repr": "janestreet/int_repr:int_repr.opam#52e9ba15ecef9d6ae1346b9503c6534bbbab212e",
+    "@opam/core_unix": "EduardoRFS/core_unix:core_unix.opam#e765288323c8d54732e253e34daea6da9b824173",
+    "@opam/timezone": "janestreet/timezone:timezone.opam#e7a86ae5c833de8615265b9d8f598653b5593b91",
+    "@opam/ocaml_intrinsics": "janestreet/ocaml_intrinsics:ocaml_intrinsics.opam#98af7a639c1b4a50c9e4b04e9b070a28129ca6a9",
+    "@opam/sexp_pretty": "EduardoRFS/sexp_pretty:sexp_pretty.opam#50f666f9c464ab8e080f585a0ce72d5b0da43452"
   }
-}
+}
```
d4hines, line 269: Can you put back this newline char?

d4hines (changes requested):
This looks fine to me, but a few comments: 
- it looks like your missing changes to the `esy.lock` folder.
- Not a blocker for this PR, but we'll also need the corresonding changes in nix. You've done the hard work for finding what combination of packages works, and myself or @ulrikstrid would just need to translate that into nix expressions.

esy.json
InfiniteSwerve: Done. For anyone else confused as to why this can matter: https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

lykimq:
@InfiniteSwerve
Do you think we can rebase this branch with the current main? or it will be easier to close this PR and fork a new branch from main and add these changes?

PR #534, opened 2022-04-07
remove annoying log message

d4hines:
## Problem

This one log is particularly annoying, and  I don't find it valuable.
It also doesn't use `Format` module, which means I miss it when I grep for our other logs by looking for `Format\..*`

## Solution

Remove it.


anchpop (approved):
LGTM :p

Issue #535, opened 2022-04-07
Deku Indexer


d4hines:
We'll need an indexer soon.

## Idea for Architecture

@EduardoRFS and I were thinking about this. It would be nice to keep `deku-node` is minimal as possible, making it easier to test and verify. To this end, one idea is to have `deku-node` only maintain the absolute bare minimum state (because we are using deterministic consensus we should only need the most recent finalized block). The archiving and indexing is done in a separate process, call it `deku-indexer`. `deku-node` communicates with `deku-indexer` via stdout or a message queue or something (we have to think about how to guarantee delievery while not compromising simplicity and performance of `deku-node`). We want independence between the two processes - I don't think DOS'ing `deku-indexer` should be allowed to affect `deku-node`, but at the same time it's important that `deku-indexer` receive the contents of every block from `deku-node` at least once.

PR #536, opened 2022-04-07
feat: add block-by-level rpc

d4hines:
## Problem

There is no way to inspect the contents of blocks from outside a deku node.

## Solution

Add a `block-by-level` rpc that takes a block level as an int and returns the block with that block height, or `null` if it doesn't know the contents of that block.

This is a workaround to the fact we have no indexer, see #535. But it's pretty essential for using Deku at all, so I think the work-around is justified for now.

### DIFF
--- a/src/bin/deku_node.ml
+++ b/src/bin/deku_node.ml
@@ -53,6 +53,17 @@ let handle_block_by_hash =
     (fun _update_state request ->
       let block = Flows.find_block_by_hash (Server.get_state ()) request.hash in
       Ok block)
+let handle_block_by_level =
+  handle_request
+    (module Networking.Block_by_level_spec)
+    (fun _update_state request ->
+      let state = Server.get_state () in
+      let block =
+        List.find_opt
+          (fun block ->
+            Int64.equal block.Protocol.Block.block_height request.level)
+          state.applied_blocks in
+      Ok block)
 let handle_block_level =
   handle_request
     (module Networking.Block_level)
@@ -131,6 +142,7 @@ let node folder prometheus_port =
              handle_received_block_and_signature;
              handle_received_signature;
              handle_block_by_hash;
+             handle_block_by_level;
              handle_protocol_snapshot;
              handle_request_nonce;
              handle_register_uri;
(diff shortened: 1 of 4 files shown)


src/node/networking.ml
```diff
@@ -64,6 +64,11 @@ module Block_by_hash_spec = struct
   type response = Block.t option [@@deriving yojson]
   let path = "/block-by-hash"
 end
+module Block_by_level_spec = struct
+  type request = { level : int64 } [@@deriving yojson]
+  type response = Block.t option [@@deriving yojson]
+  let path = "/block-by-level"
+end
```
anchpop, line 71: It's a bit late since we already have `Block_by_hash_spec`, but why do these end with `spec` while others don't?

anchpop (approved):

src/bin/deku_node.ml
```diff
@@ -53,6 +53,17 @@ let handle_block_by_hash =
...
+let handle_block_by_level =
+  handle_request
+    (module Networking.Block_by_level_spec)
+    (fun _update_state request ->
+      let state = Server.get_state () in
+      let block =
+        List.find_opt
+          (fun block ->
+            Int64.equal block.Protocol.Block.block_height request.level)
+          state.applied_blocks in
```
anchpop, line 65: Fine for now, but it's kind of sad that we can't just index into the list

lykimq:
I would prefer if the implementation accompany with the tests for it.

src/bin/deku_node.ml
d4hines: I believe the problem isn't indexing this set of blocks, but that the `state.applied_blocks` is unbounded currently. I made a TODO related to this in `./node/state.ml` and referenced the relevant issue (#535)

d4hines: When we have an indexer, we can discard any block that is below the most recently finalized blocks (at least, this is the case once we have deterministic consensus with tendermint).

d4hines:
@lykimq, are you asking me to write tests for this code? I would say that while we should have more tests in Deku, this functionality is only temporary and isn't worth automated tests. I believe the right approach for this PR is a manual test. Simply run the sandbox and the do:
```
curl -X POST -d '{"level": 1}' http://localhost:4441/block-by-level
```
and inspect the output.
(That would have been helpful to put in the PR description :thinking: - my bad)

d4hines:
Closed to consolidate PR's

PR #533, opened 2022-04-07
feat: add prometheus to docker-compose

d4hines:
## Problem

We'd like to make prometheus stats available to
devs when testing locally.

## Solution

Add Prometheus server to docker-compse. 
This makes the dashboard available on localhost:9090.

Please test it out! It's super cool :)

## How to use Prometheus

Ask Infra, I'm a prometheus noob. But here's a screenshot: 
image

If you click the "world" button you can select from available metrics. Then click "execute"
image

PR #537, opened 2022-04-07
Add block rate and operation rate metrics

d4hines:
## Depends

- [ ] #533

## Problem

We want metrics for block rate and operations per block (from which you can derive operations rate - i.e tps - in prometheus dashboard)

## Solution

Adds these metrics.

### DIFF
--- a/prometheus.yml
+++ b/prometheus.yml
@@ -1,0 +1,8 @@
+scrape_configs:
+  - job_name: 'deku'
+    scrape_interval: 1s
+    static_configs:
+      # TODO: currently these are hard-coded for development/local benchmarking
+      # But we might want to do this more dynamically. Maybe Prometheus has some
+      # dynamic config?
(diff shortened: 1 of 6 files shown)


anchpop (approved):

prometheus.yml
```diff
@@ -0,0 +1,8 @@
+scrape_configs:
+  - job_name: 'deku'
+    scrape_interval: 1s
+    static_configs:
+      # TODO: currently these are hard-coded for development/local benchmarking
+      # But we might want to do this more dynamically. Maybe Prometheus has some
+      # dynamic config?
```
renatoalencar, line 7: It depends on the type or orchestration you're using, for this case you may want to try it for Docker.

https://prometheus.io/docs/prometheus/latest/configuration/configuration/#docker_sd_config

d4hines:
Closed to consolidate PR's

PR #543, opened 2022-04-13
Serialization at interop.go and docker error

MohamadAlkhaldi:
The two commits in this pull request tackle two issues:

1- Serialization at Set method in interop.go: As mentioned before in Slack, when set is called, "value" is already serialized(type []byte). So Set should expect []byte "value" and should not serialize it again.
 
2- Port mapping and "host" network_mode in docker-compose for prometheus: 
In recent Docker versions this configuration throws a breaking error: 'docker.errors.InvalidArgument: "host" network_mode is incompatible with port_bindings'. In older versions port mapping is just ignored if network_mode is "host", so removing port mapping will make it work for new docker versions and it should not break anything in older versions.

### DIFF
--- a/go_interop/interop.go
+++ b/go_interop/interop.go
@@ -2,7 +2,6 @@ package deku_interop
 
 import (
 	"encoding/binary"
-	"encoding/json"
 	"fmt"
 	"os"
 )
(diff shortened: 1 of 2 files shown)


d4hines (approved):
Thanks for the PR! Looks good :+1:

PR #545, opened 2022-04-14
Rework development workflow

ulrikstrid:
## Problem

Even with `sandbox.sh` the setup still requires some steps, we want to get this down to as few commands as possible.

## Solution

This PR implements a [Tilt](https://tilt.dev/) setup which allows us to run the network with just 1 command (`tilt up`) and have it automatically rebuild and restart when making changes.

It has 2 dependencies which are the `tilt` command and `docker-compose` because it currently doesn't work with `docker compose`. These are added in the `nix` setup for convenience but needs to be installed by other means if not using `nix`.

Tilt also includes a nice dashboard where you get a overview of what is running and you can do some manual tasks.

image

image

Commit f2fd6ea1a2d31a8cd5b24cbc27138e007fce24bc, 28f50a1bd7882b013dc4a495665fc59ba114224c and aafc25b9d507bdee6d97f61a025af6547d066d9e could go in separately since they are just fixes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ In either case, if all goes well, you should see the block height displayed in t
 
 #### Experimental: Tilt
 
-Our experimental workflow uses [Tilt](https://tilt.dev) which is an advanced orchestrator. If you're using `nix` everything is already provided. If not, you will have to install the things in [requirements](#requirements) + `tilt` and `docker-compose`.
+Our experimental workflow uses [Tilt](https://tilt.dev) which is an advanced orchestrator. If you're using `nix` everything is already provided. If not, you will have to install the things in [requirements](#requirements) + [tilt](https://docs.tilt.dev/install.html) and [docker-compose](https://docs.docker.com/compose/install/).
 
 When the requirements are in place you should just have to run `tilt up` and everything will be running.
 
--- a/README.md
+++ b/README.md
@@ -113,6 +113,12 @@ You may have to stop the session with `killall deku-node` to fully shutdown the 
 
 In either case, if all goes well, you should see the block height displayed in the terminal and increasing every second or so.
 
+#### Experimental: Tilt
+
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -110,6 +110,12 @@ You may have to stop the session with `killall deku-node` to fully shutdown the
 
 In either case, if all goes well, you should see the block height displayed in the terminal and increasing every second or so.
 
+#### Tilt
+
+We have a experimental workflow that uses [Tilt](https://tilt.dev) which is a advanced orcastrator. If you're using `nix` you will have all the tools available to run this workflow. If not, you will have to install the things in [requirements](#requirements) + `tilt` and `docker-compose`.
```
aguillon, line 115: orchestrator*

aguillon (approved):
I'm not sure to see the point of this. At the moment, I'd be happy to be sure that we all actually engage with running the project and that we don't have a too systematic way of doing things, as I want us to add chaos to our testing. Having a tool that automatically restarts the nodes from a clean state will not help that.

But what do I know, eh. If you find it that useful, I'm not against it.

ulrikstrid:
@aguillon it will actually not start from a clean state, it only rebuilds and restarts the container (s). The UI also allows you to restart nodes with just a click which would help this testing

sandbox.sh
```diff
@@ -130,7 +147,7 @@ EOF
     --burn-cap 2 \
     --force
 
-  for VALIDATOR in ${VALIDATORS[@]}; do
+  for VALIDATOR in "${VALIDATORS[@]}"; do
```
gsebil08, line 150: What's the reason of adding these?

gsebil08 (commented):
I tested the terminal view from my debian, seems great. I had to run it twice to get the `deku-net` ok, don't know why :/

sandbox.sh
ulrikstrid: While I was in there I fixed some shellcheck stuff. We should probably add it to CI

anchpop:
@ulrikstrid just tested it, works great for me!

anchpop (approved):

d4hines (approved):
Thanks for this @ulrikstrid 
Wailord approves

PR #546, opened 2022-04-14
Improve docker file generation in CI

ulrikstrid:
## Problem

We can't test a PR in a nice way using docker images.
Our tags are not very descriptive when we just have a long git sha as tag

## Solution

We add another step that only triggers if a PR has the label "docker" and that will push with a special tag name.
Sets the tag to `YYYY-mm-dd-short_sha` and for the PR it will be prefixed with `pr-`

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -23,6 +23,12 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Set outputs
+        id: vars
+        run: |
+          echo "::set-output name=date::$(date +%Y-%m-%d)"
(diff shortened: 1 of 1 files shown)


d4hines (commented):
I had trouble with this:
```
docker run ghcr.io/marigold-dev/deku:pr-2022-04-14-4a14a59
Unable to find image 'ghcr.io/marigold-dev/deku:pr-2022-04-14-4a14a59' locally
docker: Error response from daemon: manifest unknown.
See 'docker run --help'.
```

Ideas?

anchpop:
Can this be merged?

ulrikstrid:
From my pov we can squash merge it 👍

ulrikstrid:
Fixed the history, this should now be good go

d4hines (approved):
LGTM 
wailmer approves

PR #547, opened 2022-04-14
Fix double esy install

d4hines:
## Problem
 In our docker build `esy install` runs twice and isn't cached properly.

## Solution

Split the copy commands. Not sure why this fixes it, maybe @ulrikstrid can explain. 

## Related

I cherry-picked this commit from #545 because it's super useful by itself.

### DIFF
--- a/docker/static.Dockerfile
+++ b/docker/static.Dockerfile
@@ -8,7 +8,8 @@ RUN apk add libexecinfo-dev libexecinfo-static
 WORKDIR /app
 
 # Add things that doesn't change much, or should bust cache when it does
-COPY ./esy.lock esy.json ./
+COPY ./esy.lock ./esy.lock
+COPY esy.json ./
 
 RUN esy install
 RUN esy build-dependencies --release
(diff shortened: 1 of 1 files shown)


ulrikstrid:
It was actually @callistonianembrace that found this, but the reason is that the lock directory is not properly copied over in this case. So when we run `esy install` we generate a new lock directory and then copy over it with `COPY . .`.

ulrikstrid (approved):
Since this is my commit I obviously approve

gsebil08 (approved):

docker/static.Dockerfile
```diff
@@ -8,16 +8,15 @@ RUN apk add libexecinfo-dev libexecinfo-static
 WORKDIR /app
 
 # Add things that doesn't change much, or should bust cache when it does
-COPY ./esy.lock esy.json ./
+COPY ./esy.lock ./esy.lock
+COPY  esy.json ./
```
EduardoRFS, line 12:
```suggestion
COPY esy.json ./
```

EduardoRFS (approved):
anime girl approve this PR

ulrikstrid:
Committed @EduardoRFS's fix and rebased, this should be good to go. @aguillon

PR #550, opened 2022-04-18
chore: Update michelson_v1_primitives.ml and tezos_micheline

Zett98:
## Problem

We've fallen a bit behind tezos mainline repo wrt michelson_v1_primitives
## Solution

update the file and dependency for tezos_micheline

### DIFF
--- a/esy.json
+++ b/esy.json
@@ -4,7 +4,10 @@
   "esy": {
     "build": "dune build -p #{self.name}",
     "release": {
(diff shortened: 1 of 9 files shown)


d4hines (approved):
LGTM, but I've created https://github.com/marigold-dev/deku/issues/552 to address a difficulty I had in comparing to the source.

anchpop (approved):

PR #551, opened 2022-04-19
Deku-cli: Use random nonce

### LINKED ISSUE #538: Deku CLI re-uses nonces incorrectly
This line is causing transactions identical content to only be executed 1 per block: https://github.com/marigold-dev/deku/blob/875c7211d0240bb432f0117cec120ce7753b3351/src/bin/deku_cli.ml#L151. If you wanted to submit similar transactions in the same block with cli, you can't do it.

anchpop:
## Problem

`deku-cli` always uses `0` as the nonce for every operation it submits. This makes it impossible to submit the same transaction twice in one block, which is bad because that's sometimes desirable (e.g. for testing). 

## Solution

Use a random nonce when submitting operations

## Related

- Closes #538

### DIFF
--- a/src/bin/deku_cli.ml
+++ b/src/bin/deku_cli.ml
@@ -157,7 +157,8 @@ let create_transaction node_folder sender_wallet_file received_address amount
   let block_level = block_level_response.level in
   let%await wallet = Files.Wallet.read ~file:sender_wallet_file in
   let transaction =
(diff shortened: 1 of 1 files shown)


ulrikstrid:
It should be a "number once" so this change makes sense 👌

d4hines (approved):
I tested this change by porting it over to my ticket transfer benchmark and running. I was using a counter that increased before, but random numbers work too. Looks good :+1:

PR #517, opened 2022-03-28
Smart contracts: Contract invocation

Zett98:
## Depends

- [x] #506
- [x] #515
## Problem
We need to have a way to invoke a contract
## Solution
add a user operation to invoke a contract and expose cli interface to it using already existing `transfer` command

PR #553, opened 2022-04-20
Add dummy vm

Zett98:
## Depends
- [x] #517

## Problem

We need to support multiple vm's, so this is mainly a pr about having interface to multiple vms
## Solution

Add a dummy vm

### DIFF
--- a/src/bin/deku_cli.ml
+++ b/src/bin/deku_cli.ml
@@ -129,6 +129,20 @@ let ticket =
     Format.fprintf fmt "%S" (Tezos.Ticket_id.to_string ticket) in
   let open Arg in
   conv ~docv:"A ticket" (parser, printer)
+let vm_flavor =
+  let parser string =
+    (match string with
+    | "Lambda" -> Some `Lambda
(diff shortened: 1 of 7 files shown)


src/bin/deku_cli.ml
```diff
@@ -129,6 +129,20 @@ let ticket =
     Format.fprintf fmt "%S" (Tezos.Ticket_id.to_string ticket) in
   let open Arg in
   conv ~docv:"A ticket" (parser, printer)
+let vm_flavor =
+  let parser string =
+    (match string with
+    | "Lambda" -> Some `Lambda
```
gsebil08, line 135: Naive question: Does it have to be polymorphic?
Why not defining a simple
```ocaml
type vm_type =
  | Lambda 
  | Dummy
```

Zett98: honest answer: I was too lazy to do that and we already use polymorphic variants freely in the cli

anchpop (approved):
Can't test locally right now, but LGTM

PR #554, opened 2022-04-20
chore: add shellcheck linting

### LINKED ISSUE #548: Lint Shell files with shellcheck
We should lint out shell scripts with https://www.shellcheck.net/

d4hines:
## Problem

Bash is really really hard.

## Solution

Add the https://www.shellcheck.net/ linter to our code and CI

Currently I only add it to the nix setup. We can defer merging this until everyone is on Nix if we want.

Closes #548

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -18,6 +18,8 @@ jobs:
         run: nix flake check
       - name: "Check nix formatting"
         run: nix develop -c nixfmt --check ./flake.nix ./nix/*.nix
+      - name: "Check shell formatting"
+        run: nix develop -c shellcheck *.sh
       - name: "Run tests"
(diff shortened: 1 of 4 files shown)


ulrikstrid:
Would be nice to have this merged just to run it in CI

ulrikstrid (approved):

gsebil08 (approved):
Regarding the last improvements we have done _(build with nix, nixfmt, shellcheck, etc.)_, I think it would be great to rework a bit our [contributing guide](https://github.com/marigold-dev/deku/wiki), don't you think?
Should be done in a dedicated PR of course.

ulrikstrid:
I think this PR is orthogonal to nix, @d4hines just picked nix because it was the easiest way to get it into CI since we have the infra for it already

PR #555, opened 2022-04-20
[p] Remove Esy from sandbox scripts

d4hines:
Created in draft mode until we transition the whole team to Nix.

## Depends

- [x] #554

## Problem

If we remove esy from our sandbox script, this yields:
- simpler scripts - no branching based on package manager
- a lot of leverage in packaging the script for clients. This has been really useful to me for parametric deku.

## Solution

Remove esy from the scripts.

PR #556, opened 2022-04-21
chore: Build docker images with nix

ulrikstrid:
## Problem

If we are to move to nix we need to build images with nix as well

## Solution

Build docker images with nix, we also add static builds because we want to use them in docker

## Related

- #555

### DIFF
--- a/nix/overlay.nix
+++ b/nix/overlay.nix
@@ -3,6 +3,11 @@ let disableCheck = package: package.overrideAttrs (o: { doCheck = false; });
 in {
   ocaml-ng = builtins.mapAttrs (_: ocamlVersion:
     ocamlVersion.overrideScope' (oself: osuper: rec {
+      lwt_react = osuper.lwt_react.overrideAttrs
+        (o: { nativeBuildInputs = o.nativeBuildInputs ++ [ oself.cppo ]; });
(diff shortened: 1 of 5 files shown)


nix/overlay.nix
```diff
@@ -3,6 +3,12 @@ let disableCheck = package: package.overrideAttrs (o: { doCheck = false; });
 in {
   ocaml-ng = builtins.mapAttrs (_: ocamlVersion:
     ocamlVersion.overrideScope' (oself: osuper: rec {
+      lwt_react = osuper.lwt_react.overrideAttrs
+        (o: { nativeBuildInputs = o.nativeBuildInputs ++ [ oself.cppo ]; });
+      utop = osuper.utop.overrideAttrs (o: {
+        # nativeBuildInputs = o.nativeBuildInputs ++ (with oself; [ dune ocaml cppo ]);
```
d4hines, line 9: Should we remove this commented code?

d4hines (commented):
This is great! I just have one question about the commented-out code.

nix/overlay.nix
ulrikstrid: Yes!

d4hines (approved):
This is :ok_hand: :raised_hands: 
nixified wailord approves

PR #539, opened 2022-04-07
[d4hines] Benchmarking: ticket transfers                 d4hines/ticket-transfer-benchmarking

d4hines:
## Depends
- [ ] #647 

## Problem

We want a reproducible way of testing through the throughput of Deku. This will help us determine where our bottlenecks are and where to focus optimization effrots.

## Solution

This PR adds an e2e test similar to that in #470 that attempts to saturate the network with ticket transfer transactions. 

There is a lot that could be improved in this PR:
- I have not tested the efficiency of the load tester. Obviously, if your load tester is not fast enough, you won't fully exercise your system. So I can't say with confidence that this truly saturates the network.
- We obviously need to add more variety to the tests and the testing conditions (different network topologies, killing nodes etc.).
- I'd like to automate the execution and reporting of these benchmarks but I need help from infra to do that (cc @ulrikstrid).

Despite the above, I think this is a good start, and will already provide immediately useful information to us for debugging performance issues (it already helped me detect #538), and I'm interested in feedback to get it merged in. 

### Stacked PR Chain: d4hines
| PR | Title | Status |  Merges Into  |
|:--:|:------|:-------|:-------------:|
|#691|fix: fix prometheus with tilt||-|
|#647|feat: block rate and transaction rate metrics  d4hines/block-rate-and-tps||#691|
|#658|chore: refactor cli terms                      d4hines/batch-operations||#647|
|#681|chore: refactor config out of node state||#658|
|#668|make minimum block time configurable||#658|
|#539|👉Benchmarking: ticket transfers                 d4hines/ticket-transfer-benchmarking||#658|

PR #558, opened 2022-04-25
[d4hines] Simplify Sandbox.sh

d4hines:
## Problem

Currently the sandbox.sh supports 3 workflows: local (with nix), local (with esy), and Docker.

We can simplify the script by removing esy, and using Nix internally for the team. Users without nix can still build the software with esy and run it inside Docker (but it will be much better 

I created this PR while hacking on parametric Deku - I found it much easier to package the sandbox script using nix, so I think should be the way we do it.

## Solution

Remove esy from sandbox.sh.

Additionally, I add Ligo to the nix environment, allowing us to more easily access it during contract development.

Supporting local with and AND Docker is still not ideal. In a subsequent PR, I will simplify the scripts further by parameterizing the script by environment variables and removing the `if [ $mode = "docker" ]` branches.
### Stacked PR Chain: d4hines
|#558|👉Simplify Sandbox.sh||-|
|#527|Dummy-ticket||#558|
|#536|feat: add block-by-level rpc||#527|
|#537|Add block rate and operation rate metrics||#536|
|#539|Benchmarking: ticket transfers||#537|
|#540|*(Draft)POC: Parametric Deku*||#539|


sandbox.sh
```diff
@@ -1,36 +1,22 @@
 #! /usr/bin/env bash
-# shellcheck disable=SC2155
 
 set -e
 
-if [ "$2" = "docker" ]
-then
+if [ "${2:-local}" = "docker" ]; then
```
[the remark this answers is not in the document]
d4hines: Yes, and I think bash complains of an undefined variable without it.

.github/workflows/nix.yml
```diff
@@ -84,4 +84,20 @@ jobs:
...
+  e2e_test:
+    runs-on: ubuntu-latest
+    needs: linux
+    steps:
+      - uses: actions/checkout@v2
+      - uses: cachix/install-nix-action@v16
+      - uses: cachix/cachix-action@v10
+        with:
+          name: deku
+          authToken: "${{ secrets.CACHIX_SIGNING_KEY }}"
```
EduardoRFS, line 97: Please confirm with infra if that's a good idea due to security.

EduardoRFS (approved):
The changes are okay, but I'm not very confident in Nix, please review
tambourine girl playing tambourine
 @ulrikstrid

.github/workflows/nix.yml
d4hines: cc @ulrikstrid

ulrikstrid: This is the recommended way, we might want to not push to the cache in PRs since we only have 5gb and we could in theory push out things that are in main

ulrikstrid: But that can be fixed in a separate PR

ulrikstrid (approved):
Can review the overlays on my phone but I assume it's a formatting change?
Looks good otherwise

d4hines:
> Can't review the overlays on my phone but I assume it's a formatting change? Looks good otherwise

wrong version of nixfmt. Will fix.

ulrikstrid (approved):
Looks good to me

PR #560, opened 2022-04-26
[d4hines] remove static.Dockerfile

d4hines:
## Problem
 This build job is very slow and responsible for most of our CI time.

## Solution

    We now are building the static images with nix (which is much faster
    because of the shared build graph) and can safely drop this.


ulrikstrid (changes requested):
Please remove `.dockerignore` as well

ulrikstrid (approved):
Looks good

PR #562, opened 2022-04-28
Smart contract usage howto + bugfix

Zett98:
## Problem

We added rumintary smart-contract support to deku using lambda vm but we have no instructions how to use it.
## Solution

add the instructions.
Note: patch also includes a bugfix for contract invocation on lambda


src/lambda_vm/lambda_vm.ml
```diff
@@ -15,7 +15,11 @@ let wrap compute =
   | exception Checks.Out_of_stack -> `Limits Runtime_limits_error.Out_of_stack
   | exception Gas.Out_of_gas -> `Limits Runtime_limits_error.Out_of_gas
 
-module Ir = Ir
+module Ir = struct
+  include Ir
+  let tup_arg_and_storage arg storage =
```
renatoalencar, line 20: I'm not sure about this, I think we could either have type constraints for arg and storage, use the usual type constructors or have custom functions (like `pair`)  for creating this instead.

Zett98: changed to have a module with Value constructors

src/core/contract_vm.ml
```diff
@@ -72,7 +69,7 @@ module Lambda = struct
     type t = Raw_repr.Value.t [@@deriving yojson]
   end
   module Interpreter = struct
-    let error_to_string : Lambda_vm.Interpreter.error -> string = function
+    let pp_error : Lambda_vm.Interpreter.error -> string = function
```
renatoalencar, line 72: `pp_` functions usually expect a `Format.formatter` value and return `unit`. I think `<something>_to_string` (or `of_string`) is better in this context.

src/lambda_vm/lambda_vm.ml
```diff
@@ -15,7 +15,14 @@ let wrap compute =
   | exception Checks.Out_of_stack -> `Limits Runtime_limits_error.Out_of_stack
   | exception Gas.Out_of_gas -> `Limits Runtime_limits_error.Out_of_gas
 
-module Ir = Ir
+module Ir = struct
+  include Ir
+
+  module Value_syntax = struct
```
renatoalencar, line 21: This looks great

renatoalencar (approved):
It looks great now 🚀

PR #565, opened 2022-05-03
chore: Fix docker tagging

ulrikstrid:
## Problem

We don't tag the docker images correctly which then fails the push

## Solution

Add `ghcr.io/marigold-dev/deku:` before the tag name.
Also updated  the PR tag to include the PR number to make it easier to relate them

## Related

- #564 
- #556


d4hines (approved):
LGTM :+1:

PR #573, opened 2022-05-04
load_test type error

InfiniteSwerve:
Currently, running dune build in this branch yields 
```
61 |            ~sender:(Core.Address.of_key_hash sender.key_hash)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: The function applied to this argument has type
         source:Crypto.Key_hash.t -> Core__User_operation.t
This argument cannot be applied with label ~sender
```
This change fixes said error.


d4hines:
Thanks for the fix! I applied the patch and made you co-author against the latest commit in #539

PR #578, opened 2022-05-04
WIP: add tutorial

d4hines:
## Problem

Configuring Deku is intrinsically complex.

## Solution

Help our users do it with a tutorial. The tutorial is a literate bash script, so it can be tested in CI.

This is a WIP: it's old and I haven't checked to see if it's still perfectly up to date. We also need to write the github action that tests it in CI. But it should be mostly accurate.

### DIFF
--- a/tutorial/0_configuration_tutorial.sh
+++ b/tutorial/0_configuration_tutorial.sh
@@ -1,0 +1,205 @@
+#! /bin/bash
+set -e
+
+# Deku Testnet Tutorial
+
+# To deploy a Deku testnet, we need to do several things:
+# 1. Launch a Tezos testnet
+# 2. Configure the Deku nodes that will run in the cluster such
+#     that they are all aware of each other
+# 3. Start each node
+# 4. Manually produce the first block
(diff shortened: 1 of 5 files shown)


aguillon:
Should we have both `tutorial/` and `notes/`? I'd move the former into the latter.

lykimq:
> Should we have both `tutorial/` and `notes/`? I'd move the former into the latter.
I also prefer the second option: `notes/tutorial`

lykimq:
> I don't like having a `./notes` dir committed. I'd prefer to have a `./docs` folder for official documentation. I don't know where notes should be written down, but I imagine it's in the source code as comments.

I agree to change notes to docs.

d4hines:
> > I don't like having a `./notes` dir committed. I'd prefer to have a `./docs` folder for official documentation. I don't know where notes should be written down, but I imagine it's in the source code as comments.
> 
> I agree to change notes to docs. So maybe inside docs we can have notes?

`./docs` should be our public documentation.

PR #580, opened 2022-05-06
chore: Update cmdliner

ulrikstrid:
## Problem

The community is moving to cmdliner 1.1+ and we don't want to be left beind

## Solution

Update cmdliner, make sure it works with our dependencies and update our usage.


lykimq:
Why this PR is draft?

d4hines:
> Why this PR is draft?

Because it doesn't build - we have to change a bunch of things in ./src/bin and ./test/e2e to get this build correctly. But Ulrik was kind enough to do the nix parts for us.

ulrikstrid:
I was planning on doing this with @AngryStitch but we have yet to find time to actually do it :see_no_evil:

ulrikstrid:
There is some additional cleanup we can do here but this at least lets us use the latest Cmdliner

src/bin/deku_cli.ml
```diff
@@ -765,23 +763,29 @@ let remove_trusted_validator =
...
+  let exits = Cmd.Exit.defaults in
+  Cmd.info "side-cli" ~version:"v0.0.1" ~doc ~sdocs ~exits
+
 (* TODO: https://github.com/ocaml/ocaml/issues/11090 *)
 let () = Domain.set_name "deku-cli"
 
-let () =
-  Term.exit
-  @@ Term.eval_choice show_help
+let _ =
```
ulrikstrid, line 775: `Cmd.eval` returns a `int`, I think it's the exit code.

ulrikstrid:
The cleanup I wanted to do is now done so this is ready.

src/bin/deku_cli.ml
```diff
@@ -486,17 +475,19 @@ let withdraw_proof node_folder operation_hash callback =
...
     await (`Ok ())
+
+let folder_dest =
+  let docv = "folder_dest" in
+  let doc =
+    "The folder the files will be created in. The folder must exist and be \
+     empty." in
+  let open Arg in
+  required & pos 0 (some string) None & info [] ~doc ~docv
+
```
aguillon, line 486: I would even
1. regroup folder_dest and folder_node (and similar declarations) together in the file
2. factorize them in a function, which could help giving more precise help messages for each
command

Am I missing something, or does folder_dest (which describes a destination folder, which must be
empty) only make sense for the setup_tezos command and not for withdraw_proof, which doesn't write
any file but just prints stuff on stdout?

d4hines: I think `folder_dest` can just be deleted and replaced with `folder_node`. I made a commit that does this here: https://github.com/marigold-dev/deku/pull/580/commits/2a3eede547190ae1ae8fe2980f9018f328f2df72; @ulrikstrid @aguillon feel free to revert if you don't like this change.

d4hines (approved):
Thanks for this! LGTM

aguillon (approved):
Ok then

nix/overlay.nix
```diff
@@ -19,34 +19,16 @@ in {
...
       alcotest = osuper.alcotest.overrideAttrs (o: {
         propagatedBuildInputs =
           prev.lib.lists.remove osuper.uuidm o.propagatedBuildInputs;
       });
-      qcheck = osuper.qcheck.overrideAttrs (_: {
-        src = prev.fetchurl {
-          url = "https://github.com/c-cube/qcheck/archive/v0.18.1.tar.gz";
-          sha256 = "1jzhwrzsf5290rs7hsa1my5yh1x95sh2sz92c4svd8yahzdlny7m";
-        };
-      });
```
ulrikstrid, line 32: Moving a question here since this is the source of the diff.

From @aguillon 
> I don't usually follow what's going on with Nix scripts but I'm not sure why this was introduced in the first place, and why it has to go now? I believe qcheck is used in Pollinate and @gsebil08 and @rosalogia will probably use it more.

We needed a specific version of `qcheck` before this because they had a dependency on older `cmdliner`. `qcheck` is still available for use however :)

PR #579, opened 2022-05-05
Nix: webpack support

r-muhairi:
## Problem

Webpack fails to work on nix

## Solution

Implements support for webpack for nix builds using dream2nix

@d4hines

PR #581, opened 2022-05-06
chore: use esbuild instead of webpack

ulrikstrid:
## Problem

Webpack is slow and problematic in nix

## Solution

Use [esbuild](https://esbuild.github.io) instead

## Related

- #579


ulrikstrid:
@d4hines It can't find dependencies installed with `esy` because it uses a non-standard way of providing paths to the node packages.
It would be good if @EduardoRFS or @callistonianembrace could implement `NODE_PATH` upstream in `esy`, that way things would work out of the box.

callistonianembrace:
First off - apologies for the shabby JS code. Had to do this in a jiffy.

1. I added a esbuild plugin (light tweaks to the yarn pnp plugin)
2. Disabled sandbox on the CI because esbuild does sneaky file moving in the sources area (which violates immutability assumptions esy makes)

With this, constructing a NODE_PATH is unnecessary.

Happy to explain the plugin offline too. It isn't much - it captures the require statements (like the yarn pnp plugin did) and uses esy's pnp js to resolve them.

callistonianembrace:
I can rework the plugin code sometime later this week if this indeed is the direction you all wish to take. These commits could be considered a PoC. Let know what you think.

callistonianembrace:
Also, I'm not entirely sure why Nix is failing here. Seems like we need something around Dune rules and not much to do with module resolution per se.

ulrikstrid:
@callistonianembrace making `NODE_PATH` available in `esy` would be more flexible. `nix` doesn't know about pnp so it will fail with this solution. But maybe we could create a custom resolver that will check if `NODE_PATH` is set and then use that, otherwise do the pnp stuff.

callistonianembrace:
@ulrikstrid I'm missing something here - this is about esy + nix right? Why doesn't Nix see esy's pnp.js file?

ulrikstrid:
@callistonianembrace nix is separate from esy so when using nix we don't even run esy

ulrikstrid:
The main problem, getting m1 to work in nix, has been resolved in another PR so we can close this

Issue #574, opened 2022-05-04
Document and validate KPIs for Deku benchmarks


lykimq:
# Summary

Define KPIs to benchmark various parts of Deku (consensus, hashing...)

# Motivation

We want to measure Deku's ability to handle an important number of transactions. This number depends on several factors (number of validators, hardware, network...). We want to establish a protocol to make this measurements and find potential bottlenecks in the project.

# Production

A document recalling these KPIs, the factors on which they depend and the libraries used, which will be added to notes/ (and HackMD?).

PR #582, opened 2022-05-06
Notes: doc benchmarks

lykimq:
issue #574


notes/benchmarks/benchmarks_kpis.md
```diff
@@ -0,0 +1,71 @@
...
+## Production
+
+We define the KPIs into 3 states, and by priority order.
+
+### State 1
+
+Focus on getting the number of TPS, hardware configurations, etc.
+ 
+There are two enviroments we will do the benchmark:
+- Local environment existing on a single machine
```
aguillon, line 22: For development only, right?

notes/benchmarks/benchmarks_kpis.md
```diff
@@ -0,0 +1,71 @@
...
+
+We want to measure Deku's ability to handle an important number of transactions. This number depends on several factors (number of validators, hardware, network...). We want to establish a protocol to make this measurements and find potential bottlenecks in the project.
+
+## Production
+
+We define the KPIs into 3 states, and by priority order.
+
+### State 1
+
+Focus on getting the number of TPS, hardware configurations, etc.
```
aguillon, line 19: I'd be more specific: can you list all the factors you expect to test, and be specific about the hardware you want to change as well? Others (infra team?) will probably want to read this as well.

notes/benchmarks/benchmarks_kpis.md
```diff
@@ -0,0 +1,71 @@
...
+- Normal transaction (A sends 10 xtz to B)
+- Smart contract call (A calls a smart contract C)
+
+For the latter, we need to identify what kind of smart contract we are interested to get the benchmark for. This will require an input from team leader.
+
+The scenario that we have in mind are:
+- Base case: increase the number of transactions (10, 100, 1000, etc.) with the base number of validators (e.g. minimum 4 for Tendermint consensus).
+- Growing transactions, validators: the same pack of growing transactions (10, 100, 1000, etc.) with the growing of number of valiators (4, 6, 8, 10, etc.)
+
+Different hardware configurations (cloud) set up for same pack of the scenario above.
```
aguillon, line 35: Related to previous question on hardware: are you testing specific configurations offered by typical cloud providers, or are you targeting independent hardware variables (increasing RAM size, increasing CPU power…)?

notes/benchmarks/benchmarks_kpis.md
lykimq: - can you list all the factors you expect to test?

I think it is what I explain below:

```
For transactions, there are two kinds of transactions we are taking into account:
- Normal transaction (A sends 10 xtz to B)
- Smart contract call (A calls a smart contract C)
```

About the hardware, currently I do not have any specific configuration that we want to test. I will ask the infra team for help about it and will update the doc.

notes/benchmarks/benchmarks_kpis.md
lykimq: The second one: For the variables: increase RAM size, CPU power. (if the cloud providers have)

notes/benchmarks/benchmarks_kpis.md
```diff
@@ -0,0 +1,71 @@
...
+
+### State 2
+- Profile Deku-node
+ 
+By profiling the functions responsible for running a Deku node we can get a closer look at the efficiency of the node itself, which will be running all the time, which means that any small efficiency increases will have much larger payoffs.
+
+- Hashing the state root hash of block
+
+Focus on answering a question: how many GB a state root hash of a block can have?
```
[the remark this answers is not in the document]
lykimq: What I mean is that: if we have 10k operations then at the end after hashing how many GB we archive? and if possible what is the max of that GB? does it make sense?

notes/benchmarks/benchmarks_kpis.md
lykimq: Can you explain more? What do you mean by development only?

lykimq: What I mean by locally is what we are running the benchmarks on for instance, my laptop. And we will run that same tests on cloud.

notes/benchmarks/benchmarks_kpis.md
ulrikstrid: I think we want to be able to run benchmarks in 3 different scenarios:

1.  Locally to easily see some regressions when working on performance sensitive code
2. In CI (GitHub Actions) which should be more stable to see that we don't introduce big regressions
3. In a testnet to see that we don't slow down over time. This is most likely some cronjob that creates different transactions and then we can look at graphs from  prometheus data

1 and 2 should be written with the help of some benchmarking library
3 is a e2e test that we can run periodically

lykimq: I added your comments into the doc. Thanks.

d4hines: @ulrikstrid, my thinking is that we do not know where to write benchmarks for 1 a priori. I think we should run some kind of e2e benchmarks in CI constantly, at least until we better understand the performance characteristics of the system. If every commit is too much, we can do nightly.

notes/benchmarks/benchmarks_kpis.md
```diff
@@ -0,0 +1,71 @@
...
+
+Define KPIs (Key Performance Indicators) to benchmark various parts of Deku (consensus, hashing...)
+
+## Motivation
+
+We want to measure Deku's ability to handle an important number of transactions. This number depends on several factors (number of validators, hardware, network...). We want to establish a protocol to make this measurements and find potential bottlenecks in the project.
+
+## Production
+
+We define the KPIs into 3 states, and by priority order.
```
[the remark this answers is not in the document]
lykimq: Thanks, typo: stage

notes/benchmarks/benchmarks_kpis.md
aguillon: I was wondering if you were aiming for the same workloads on your laptop (with several instances of deku running!). It does not sound feasible, nor necessary, to me.

lykimq: Yes, not the same workloads of testing the hardware on my laptop.

aguillon (approved):
Ok

PR #584, opened 2022-05-06
Infra: add bench libs esy

lykimq:
Add benchmark libs: core_bench, memtrace, landmarks, benchmark in Esy. 
Change the folder name: core to deku_core because of the same name of the library: Core.


d4hines (approved):
This change looks good. Thanks.

esy.json
```diff
@@ -221,6 +225,80 @@
...
       }
-    }
+    },
+    "@opam/core": "InfiniteSwerve/core:core.opam#9c8f9f90ac74fe33aae13c7c040951a5ab78b064",
+    "@opam/core_bench": "janestreet/core_bench:core_bench.opam#48b0271934d742da433b109b41ccba53b5de4682",
+    "@opam/bin_prot": "EduardoRFS/bin_prot:bin_prot.opam#4b944ab23b0e71e995b2abb7032ff9dd1b3443f3",
+    "@opam/ppx_sexp_message": "janestreet/ppx_sexp_message:ppx_sexp_message.opam#3b8dc0d319a5e1981790fee786646d15c8a11371",
+    "@opam/ppx_custom_printf": "janestreet/ppx_custom_printf:ppx_custom_printf.opam#b809ddfa0b8b97980becd58a2a5929266719cb38",
+     "@opam/ppx_bin_prot": {
+      "version": "opam:v0.14.0",
```
ulrikstrid, line 235: I'm pretty sure we need 0.15 for ocaml 5.0 compatibility, are you sure this works?

ulrikstrid (changes requested):
I think we need 0.15 of janestreet packages

esy.json
lykimq: I tried to change it to `0.15` and this is the error of it. It passed compile with `0.14`

```
error: No solution found:

No package matching:

  sidechain -> @opam/core_bench -> @opam/core -> @opam/ppx_jane -> @opam/ppx_bin_prot@>=opam:v0.14 && @opam/ppx_bin_prot@<opam:v0.15
  
  Versions available:
  
    @opam/ppx_bin_prot@opam:v0.15.0

  
esy: exiting due to errors above
```

aguillon: @ulrikstrid What makes you say this?

aguillon:
I don't understand the issue with the CI on Mac OS, how come the build options are different from those on Linux?

esy.json
ulrikstrid: @aguillon I thought janestreet packages 0.14 is not compatible with OCaml 5.00. If this is not true then I guess that's good :ok_hand: But it might be worth noting that we're on 0.15 in nix so there might be conflicts between setups if we stay on 0.14 in esy.

lykimq:
I am closing this PR because it takes so long for Esy to resolve it.

PR #585, opened 2022-05-06
dev-environment: run any number of nodes in tilt

ulrikstrid:
## Problem

This PR addresses multiple issues:
* Our docker image is just named `deku` when we generate it
* In the docker-case for `sandbox.sh` we didn't run the commands correctly
* Our tilt setup still tried building docker images the old way
* `docker-compose.override.yml` is apparently a special name that will be applied so we're always running (old) deku nodes in compose
* We can only run 3 nodes

## Solution

With this PR you can run `tilt up -- --nodes 10` to run a network with 10 nodes


d4hines (approved):
LGTM. Thanks for this Ulrik

PR #589, opened 2022-05-09
Nix: add missing folder direnv

lykimq:
When using `direnv` in Nix, it is missing the folder `.direnv` to execute the command.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,3 +1,5 @@
+mkdir -p .direnv
+
 use_flake() {
   watch_file flake.nix
   watch_file flake.lock


.envrc
```diff
@@ -1,3 +1,5 @@
+mkdir -p .direnv
```
ulrikstrid, line 1: If we move this inside of the `use_flake` function we don't create it when we don't use nix. This might be useful if we create a esy setup in the future.

This is however not a blocker since I don't think there's a good workflow for esy and direnv currently.

What do you think @d4hines?

ulrikstrid (approved):
Added a small comment, but I think this is just a strict improvement

.envrc
d4hines: Yeah, I think this is ok for now.

d4hines (approved):

PR #595, opened 2022-05-10
Tilt configuration for  deku-parametric

Pilou97:
## Problem

The tilt configuration does not start the vm for deku parametric


Tiltfile
```diff
@@ -1,21 +1,25 @@
...
 
-  for i in range(n + 1):
+  for i in range(n):
     deku_node_name = "deku-node-%s" % i
     services.append((deku_node_name, {
     'container_name': deku_node_name,
     'restart': 'always',
     'image': 'ghcr.io/marigold-dev/deku',
-    'expose': [ 4040 ],
+    'expose': [ '444%s' % i ],
```
ulrikstrid, line 20:
```suggestion
    'expose': [ '4040%s' % i ],
    'ports': [ '404%s:4040' % i ],
```

This way we don't need to use `network_mode: host`

Pilou97: Yep I see what you mean, but if I do that deku nodes won't be able to communicate which each others
To do so, I have to use `./sandbox setup docker` and `./sandbox start docker`, with these commands nodes will be able to communicate with each others.
But this commands introduce a new problem : I can't use `deku-cli create-transaction ...` in my terminal because it doesn't know the uri http://deku-node-1:4440 (which is read from the node folder) or I can't use `curl` on the nodes either
If you have an idea I'm very interested, I'm a bit lost about networking with docker

ulrikstrid: This PR solves it another way: https://github.com/marigold-dev/deku/pull/615

PR #597, opened 2022-05-10
Run nix CI on macOS

ulrikstrid:
## Problem

We can't be sure our nix setup works on macOS as we do with esy

## Solution

Run the nix build on both Linux and macOS


d4hines (approved):
Thanks for this!

PR #599, opened 2022-05-11
increase number of domains for async state hashing

d4hines:
## Problem

Four is too small a default for the number of domains to setup.

## Solution

Let's use 8.


lykimq:
Newbie question: What is the purpose of domain? is it like the number of deku node that we allow to run?

EduardoRFS (approved):
little animel girl approving this PR

PR #600, opened 2022-05-11
networking: optimize broadcasting

EduardoRFS:
## Problem

Currently Deku TPS is quite low, this is due to the block producer waiting the other nodes to parse each block, but currently this is done in serial, which makes so that our pipeline is empty most of the time.

Also we're serializing the same message many times for each validator.

## Solution

Sends the messages in parallel and serialize only once.

### DIFF
--- a/src/node/networking.ml
+++ b/src/node/networking.ml
@@ -9,10 +9,11 @@ module type Request_endpoint = sig
   val path : string
 end
 exception Error_status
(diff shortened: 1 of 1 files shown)


d4hines (approved):
zoomers go zoom

d4hines:
About a 50% TPS increase. Nice. 
image

PR #601, opened 2022-05-11
node: optimize pending_operations a map

EduardoRFS:
## Problem

Currently pending_operations is a list, but in almost all the usages of it we first need to search on it to avoid duplicates, this makes so that any linear operation on pending operations is actually O(n^2).

## Solution

Because the order doesn't matter for us right now, I changed from a list to a map which in OCaml is a binary search tree, making so that all the O(n^2) operations are now O(n log n).

## **Warning**

This code is not ideal, but it's not a big regression, so I would ask for a bit of patience.


d4hines (approved):
LGTM
getta ger fst

PR #603, opened 2022-05-11
network: move to it's own library

EduardoRFS:
## Problem

All of our networking lives inside the node library, which also contains bits from the consensus.

## Solution

This makes the networking module it's own library called Network, so that we can work on it without making random cycles.


ulrikstrid:
How does this relate to pollinate?

d4hines (approved):

d4hines:
@ulrikstrid, this will make it easier to integrate Pollinate, as our network layer becomes less complected to the actual application state.

PR #604, opened 2022-05-11
POC: Add json reporter for logs and change some outputs to use Logs

ulrikstrid:
## Problem

We want predictable outputs in our logs

## Solution

Use a custom reporter that will add metadata like timestamp etc and start using the `Logs` library.
Created this repo since it's non-trivial and we can re-use it for the dashboard: https://github.com/marigold-dev/json-logs-reporter

## Related

- #285 @anchpop  feel free to pick anything or nothing from this PR or work from here


src/bin/deku_node.ml
```diff
@@ -150,6 +150,12 @@ let node folder prometheus_port =
 (* TODO: https://github.com/ocaml/ocaml/issues/11090 *)
 let () = Domain.set_name "deku-node"
 
+let () =
+  (* This is needed because Dream will initialize logs lazily *)
+  Dream.initialize_log ~enable:false ();
+  Logs.set_reporter (Json_logs_reporter.reporter Fmt.stdout);
+  Logs.set_level (Some Info)
+
```
aguillon, line 158: Is this the way to prevent Dream from formatting the output as in @anchpop's former PR?

aguillon:
Ok, but this makes it harder to read the output when developing: either add an option to turn it off, or (probably better) we need to think of a way to input `jq` filtering and formatting. I'm pretty bad with `jq`, but something like `| jq -r '"[\(.time)]\t \(.log.level) \(.message)"'` is a start IMO (except that the field `log.level` is hard to use, because of the name…). Any suggestion?

src/bin/deku_node.ml
ulrikstrid: `Dream.initialize_log ~enable:false ();` tells Dream not to add it's own reporter and level.
```
  Logs.set_reporter (Json_logs_reporter.reporter Fmt.stdout);
  Logs.set_level (Some Info)
```
sets our custom reporter and a log level

aguillon: Ok thanks

aguillon:
@ulrikstrid @anchpop something like this for now (see last commit)? Shouldn't Piaf be in level "Debug" btw?

aguillon (approved):

PR #605, opened 2022-05-11
network: parallel encode and decode

EduardoRFS:
## Problem

Currently Deku encode and decodes all the messages on the main thread, this makes so that if users are sending messages fast the node freezes while processing those.

## Solution

This adds per message parallelism, as parsing the message in another thread is a small overhead due to OCaml Multicore, this leads to a considerable performance increase.


nix/overlay.nix
```diff
@@ -5,6 +5,17 @@ in {
     ocamlVersion.overrideScope' (oself: osuper: {
       lwt_react = osuper.lwt_react.overrideAttrs
         (o: { nativeBuildInputs = o.nativeBuildInputs ++ [ oself.cppo ]; });
+      lwt_domain = osuper.buildDunePackage {
```
d4hines, line 8: @ulrikstrid can you help me fix this correctly?

d4hines (approved):
I'm going to approve the change, but the nix setup is not ideal. @ulrikstrid, what's the right way to do this?

nix/overlay.nix
```diff
@@ -5,6 +5,17 @@ in {
...
+        pname = "lwt_domain";
+        src = prev.fetchFromGitHub {
+          owner = "ocsigen";
+          repo = "lwt";
+          rev = "654952b62dccea63f186747be79916dbfdccf48d";
+          sha256 = "sha256-j+ufbnovlcJnKM0DwU3Z/DRLeyJ9+AmhKSrJCpMVW7w=";
+        };
+        version = "dev";
+        propagatedBuildInputs = with osuper; [ domainslib lwt ];
+      };
```
ulrikstrid, line 18:
```suggestion
      lwt_domain = osuper.buildDunePackage {
        pname = "lwt_domain";
        version = "0.2.0-dev";
        inherit (osuper.lwt) src;
        propagatedBuildInputs = with osuper; [ domainslib lwt ];
      };
```

ulrikstrid: This seems to work on my end and is closer to what we would upstream.

ulrikstrid: Opened this PR to the overlays: https://github.com/anmonteiro/nix-overlays/pull/291

src/helpers/parallel.ml
```diff
@@ -0,0 +1,26 @@
+open Lwt_ext.Let_syntax
+
+(* lazy because Unix.fork stops working after domains start *)
+(* TODO: proper number of domains *)
+(* THIS IS GLOBAL STATE OUTSIDE OF src/node/server.ml. CAUTION *)
+let pool = lazy (Lwt_domain.setup_pool 8)
+let pool () = Lazy.force_val pool
```
aguillon, line 7: Shouldn't we avoid this on account that Lazy.force_val is racy? Is it ok just because `setup_pool` is fast?

aguillon (approved):
Thanks!

nix/overlay.nix
ulrikstrid: This is merged in the overlays, if you need the very specific version you added then you can just override the `src` now, otherwise it should get the same version as the lwt version we have

src/helpers/parallel.ml
d4hines: What race do you have in mind? I'm pretty sure `Lazy.t` is just a more efficient wrapper over what we're doing elsewhere, which is `'a option ref`. If Lazy.t is subject to race's, then so is our ref, and then we have a big issue with async state hashing.

d4hines: cc @EduardoRFS

d4hines: Answer from @EduardoRFS: 
- Arthur is right that Lazy.t is not thread safe.
- But here the Lwt_domain pool is only accessed in the main thread, so it's ok (just like the task pool for async state hashing is only accessed from the main thread).
- If we spawned domains from inside domains, this would not be safe, but we don't ,so it's ok.

aguillon: Ah yes, sure. Thanks!

PR #607, opened 2022-05-12
Add hydraJobs

ulrikstrid:
## Problem

We can't run hydra on main

## Solution

Add hydra config.
The hydra jobset looks like this: https://hydra.strid.ninja/jobset/deku/deku

This is purely additive and doesn't touch the current config.


d4hines (approved):
LGTM
hades gives thumbs up

PR #608, opened 2022-05-12
fix nix pollinate

ulrikstrid:
## Problem

We want to use pollinate

## Solution

Add pollinate's overlay from https://github.com/marigold-dev/pollinate/pull/41
When the above PR is merged we can change the flake input to point to main instead of my branch


d4hines (approved):
LGTM

PR #611, opened 2022-05-12
deps: move to nodejs 14 as nodejs 12 is EOL

EduardoRFS:
## Problem

NodeJS 12 is now EOL and nix complaints about that when updating.

## Solution

Just move to NodeJS 14.


ulrikstrid (approved):
Yes please

PR #612, opened 2022-05-12
chore: use dream2nix for a better npm experience

ulrikstrid:
## Problem

The current npm solution for nix doesn't work too well

## Solution

Use dream2nix that seems to work better with newer versions of npm.
This is a simplified/refactored version of #579

Also removed landmarks on `aarch64-darwin` since it's not compatible with the m1

As you can see here it unblocks m1: https://hydra.strid.ninja/jobset/deku/deku-m1

## Related

- #579 This is heavily inspired by the work in this PR so I'm adding co-author

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -10,7 +10,10 @@
   inputs = {
     nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
     flake-utils.url = "github:numtide/flake-utils";
-    nix-npm-buildpackage.url = "github:serokell/nix-npm-buildpackage";
+    nix-filter.url = "github:numtide/nix-filter";
+
+    dream2nix.url = "github:nix-community/dream2nix";
+    dream2nix.inputs.nixpkgs.follows = "nixpkgs";
 
     ocaml-overlays.url = "github:anmonteiro/nix-overlays";
     ocaml-overlays.inputs = {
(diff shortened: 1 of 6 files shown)


nix/deku.nix
```diff
@@ -14,15 +14,14 @@ in ocamlPackages.buildDunePackage rec {
       [ "dune-project" "sidechain.opam" "package.json" "package-lock.json" ];
   };
 
-  configurePhase = ''
-    export PATH=${npmPackages}/node_modules/.bin:$PATH
-    export NODE_PATH=${npmPackages}/node_modules
-
-    ln -s ${npmPackages}/node_modules ./node_modules
-  '';
+  #preBuildPhase = ''
```
d4hines, line 17: Dead code?

nix/npm.nix
```diff
@@ -0,0 +1,38 @@
...
+
+    packageOverrides = {
+      webpack-cli = {
+        remove-webpack-check = {
+          patches = [ ./patches/remove-webpack-check.patch ];
+        };
+      };
+    };
+
+    inject = { acorn-import-assertions."1.8.0" = [[ "acorn" "8.7.1" ]]; };
```
d4hines, line 18: Curious, what is this?

nix/shell.nix
```diff
@@ -1,9 +1,7 @@
-{ pkgs, deku, ligo, npmPackages }:
+{ pkgs, deku, ligo, nodejs ? pkgs.nodejs }:
 pkgs.mkShell {
   shellHook = ''
-    export NODE_PATH=${npmPackages}/node_modules
```
d4hines, line 4: this is a slight loss because now we can't run nodejs locally with our node_modules pre-configured. However, I think it's reasonable, and it's easy to add it back if we need it.

nix/deku.nix
ulrikstrid: Yes, thank you

nix/npm.nix
ulrikstrid: I think it's a dependency of webpack, I tried removing it but then it complained, we should add a comment

nix/shell.nix
ulrikstrid: This is now wrapped into the webpack binary. We can easily revert this if you run it in another way

ulrikstrid:
Fixed comments, formatted and rebased on main

flake.nix
```diff
@@ -10,7 +10,10 @@
   inputs = {
     nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
     flake-utils.url = "github:numtide/flake-utils";
-    nix-npm-buildpackage.url = "github:serokell/nix-npm-buildpackage";
+    nix-filter.url = "github:numtide/nix-filter";
```
ulrikstrid, line 13: After we land this we should use `nix-filter` for the deku source to cut down on the rebuilds when changing unrelated files

d4hines (approved):
LGTM

PR #615, opened 2022-05-13
chore: Split tilt to allow both docker and local workflow

ulrikstrid:
## Problem

`Tilt` currently only runs with nodes in docker, we often want to run them locally

## Solution

Moved the current (docker based) deku nodes configuration into a separate file.
Created a file for the local workflow with the same API but calls `sandbox.sh` with local settings.

Added a `mode` argument that can be used like `tilt up -- --mode=local`, it will default to `docker` when omitted.

When running in tilt we also override the prometheus service with a special configuration that reads the nodes running in docker.
If `USE_NIX` is unset we will not try to build the docker image with `nix`

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -30,28 +30,21 @@ jobs:
         run: nix develop -c dune runtest --force --no-buffer
       - name: "Build deku via flakes"
         run: nix build --verbose .#deku
-  e2e_test:
+  # Includes all our e2e tests
+  tilt_ci:
     runs-on: ubuntu-latest
     needs: build
     steps:
       - uses: actions/checkout@v2
+      
       - uses: cachix/install-nix-action@v16
       - uses: cachix/cachix-action@v10
         with:
           name: deku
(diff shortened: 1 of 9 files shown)


oteku:
LGTM it is an improvement over the current situation.

Docker mode is base on remote image: i guess it is used for demonet
It would be great to have a "local docker mode" where you build docker container from an image of the actual source for intergration testing purpose. Also it would be great to upgrade flextesa to Ithaca (Benjamin's test looks like it works)

For Demonet, it would be better to use an actual Tezos testnet (ithacanet?)

So we should have:
- Tiltfile.demonet (docker)
- Tiltfile.integration (docker)
- Tiltfile.local

=> Feel free to create a follow-up issues if you wanna keep PR small

oteku:
> By demonet, are we talking about the long running net that infra team is running or something else?

Yes

ulrikstrid:
Might want to drop the "raw" e2e test before we merge this since `tilt ci` does the same thing but not as flaky

d4hines (approved):
Tested, works! Stoked to start using this day-to-day
pirate tested and approved

PR #616, opened 2022-05-13
chore: Use nix-filter to filter sources

ulrikstrid:
## Problem

Sometimes we have rebuilds in nix even when we changed a file that we don't care about

## Solution

Since we added  `nix-filter` in a previous PR we can use that since it seems to work very well for this

## Related

- #612


d4hines (approved):
LGTM!

PR #618, opened 2022-05-16
Wasm VM proof of concept

renatoalencar:
## Problem

We need to define an interface to interact and create contracts running on a WASM runtime.

## Solution

Create bindings for the wasm spec implementation for instantiating and running contracts.


src/wasm-vm/contract.ml
```diff
@@ -0,0 +1,21 @@
...
+type t = {
+  module_ : Ast.module_;
+  storage : bytes;
+}
+
+let make ~storage ~code =
+  try
+    let module_ =
+      match Parse.string_to_module code with
+      | { it = Script.Textual module_; _ } ->
```
Zett98, line 12: bind left_out fields to `unused_field = _`

Zett98 (changes requested):

src/wasm-vm/contract.ml
```diff
@@ -0,0 +1,21 @@
...
+        Valid.check_module module_;
+        module_
+      | _ -> assert false in
```
Zett98, line 15: We shouldn't crash the node when user submits binary repr of the module

src/wasm-vm/runtime.ml
```diff
@@ -0,0 +1,54 @@
...
+      (* It is trivial to have an arugment 0 when the argument position is known,
+         * but then we can make sure that we can put the argument wherever we want to. *)
+      (0L, Int64.of_int @@ Bytes.length argument) in
+    memory_blit memory storage t.contract.storage;
+    (Int64.to_int32 argument, Int64.to_int32 storage) in
+  try
+    (* TODO: Encode operations on the return *)
+    match
+      Eval.invoke gas entrypoint [Value.i32 argument; Value.i32 storage]
+    with
```
Zett98, line 43: no nested with please

Zett98 (changes requested):

src/wasm-vm/contract.ml
renatoalencar: If you observer the parser we can't get into the binary and quoted formats from `string_to_module`.

https://github.com/Zett98/spec/blob/deku/text/parser.mly#L1173

I also added a test case if you'd like to check for yourself, or if I'd understanding this incorrectly.

```ocaml

let test_successful_parsing () =
  let code = "(module binary \"\x00asm\x01\x00\x00\x00\x01\x07\x01`\x02\x7F\x7F\x01\x7F\x03\x02\x01\x00\x05\x03\x01\x00\x00\x07\x10\x02\x03add\x00\x00\x06memory\x02\x00\n\t\x01\x07\x00 \x00 \x01j\x0B\x00$\x10sourceMappingURL\x12./release.wasm.map\")"
  in
  ignore (parse code)
```

Zett98 (approved):

PR #620, opened 2022-05-18
change core to core_deku; format codes

lykimq:
The folder `core` has the same name with the library `Core` of JaneStreet. It will conflict when using the `core_bench` for benchmarking.

Format codes by adding space between functions, make it easier to read codes.


lykimq:
> The thing is, some of us have a pretty strong opinion on what line should be blank or not. I don't (but I wish this would be enforced by ocamlformat), but I have to ask for other persons' opinion here.

Yes, I agree on the ocamlformat. If we get opinion to favor it, then I can split this work in ocamlformat and the changing of core -> deku_core into 2 PRs.

PR #624, opened 2022-05-18
Rename core to core_deku

lykimq:
Change core folder to core_deku because it is the same name with `Core` of Janestreet.


aguillon (approved):

d4hines (approved):
LGTM

PR #626, opened 2022-05-18
add global ticket_table 

Zett98:
## Problem

We're going to start moving tickets between implicit and originated accounts  (e.g. putting ticket handles in contract storage, putting them as parameters, dropping from table, etc(as in "End of Thinking Capacity".)

## Solution
Create a global ticket table, with api that in theory shouldn't allow you to do manipulations with ticket that is already droppped or doesnt belong to you.

### DIFF
--- a/src/core/contract_ffi.ml
+++ b/src/core/contract_ffi.ml
@@ -1,0 +1,55 @@
+(* type _ syscall =
+     | FAILWITH         : int -> bytes syscall
+     | EXAMPLE          : int -> (int * string) syscall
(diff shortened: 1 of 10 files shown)


src/core/contract_ffi.ml
```diff
@@ -0,0 +1,55 @@
+(* type _ syscall =
```
renatoalencar, line 1: Why this is commented?

Zett98: This is just ffi to not lose it anywhere

Issue #606, opened 2022-05-12
initial state for state machines in SDK


d4hines:
- [ ] JS SDK
- [ ] Go SDK

d4hines:
This works now 👍

PR #628, opened 2022-05-19
Add state initialization

Pilou97:
## Problem

 - https://github.com/marigold-dev/deku/issues/606
 - add the missing  fa2 example


sdks/deku_js_interop/src/interop.ts
```diff
@@ -69,13 +69,20 @@ const set = (key: string, value: string) => {
 
 /**
  * The main function
+ * @param initial_state the initial state of your vm
  * @param state_transition the function called when there is an new input
  */
 const main = (
+  initial_state: {[key: string]: any},
```
d4hines, line 76: `any` is not a good type for JSON in typescript.

Please add something like this:
```
// TODO: add a better type for JSON values
```

sdks/deku_go_interop/interop.go
```diff
@@ -96,7 +110,7 @@ func Main(initial_state map[string]any, state_transition func(sender string, tx_
 		var end_message []byte
 		if err != nil {
 			end_message = []byte(fmt.Sprintf("[\"Error\", \"%s\"]", err.Error()))
-
+
```
d4hines, line 113: Looks like some whitespace got added here. please remove.

src/external_vm/external_vm.ml
```diff
@@ -20,7 +22,17 @@ let start_vm_ipc ~named_pipe_path =
     Some
       (External_process.open_pipes ~named_pipe_path ~to_yojson:Fun.id
          ~of_yojson:vm_message_of_yojson)
-
+let empty () =
+  match !vm with
+  | None -> String_map.empty (*TODO: failwith ? *)
```
d4hines, line 27: You're right, we should fail here.

src/external_vm/external_vm.ml
```diff
@@ -36,6 +48,7 @@ let apply_vm_operation ~state ~source ~tx_hash operation =
     let state = ref state in
     while not !finished do
       match vm.receive () with
+      | Init _ -> ()
```
d4hines, line 51: We should fail in this case, since the vm has violated the protocol.

d4hines (changes requested):
This looks good - it's the change I expected for this issue. I left a comments to address. Also, please clean up the commit history - fewer commits that encapsulate well-structured changes will make it easier to manage our fork of Deku.

d4hines:
I've merged this into #540. Thanks for this @Pilou97!

PR #630, opened 2022-05-19
allowing controlling verbosity of node from CLI

d4hines:
## Problem

Merging #604 removed some logs from the default sandbox setup that were used to assess whether the node was working or not.

## Solution

Add the Logs Cmdliner terms to our setup, so we can control the verbosity from the command line.

Created in draft mode because I'm having some trouble building the node with this setup.

### DIFF
--- a/src/bin/deku_node.ml
+++ b/src/bin/deku_node.ml
@@ -189,14 +189,16 @@ let node folder prometheus_port =
 (* TODO: https://github.com/ocaml/ocaml/issues/11090 *)
 let () = Domain.set_name "deku-node"
 
-let () =
+let setup_log level () =
   (* This is needed because Dream will initialize logs lazily *)
   Dream.initialize_log ~enable:false ();
-  match Sys.getenv_opt "DEKU_LOGS" |> Option.map String.lowercase_ascii with
-  | Some "json" ->
-    Logs.set_reporter (Json_logs_reporter.reporter Fmt.stdout);
-    Logs.set_level (Some Info)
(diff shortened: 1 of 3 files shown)


src/bin/deku_node.ml
```diff
@@ -189,14 +189,16 @@ let node folder prometheus_port =
...
-    Logs.set_reporter (Json_logs_reporter.reporter Fmt.stdout);
-    Logs.set_level (Some Info)
-  | _ -> Logs.set_reporter (Logs.format_reporter ~app:Fmt.stdout ())
+  (match Sys.getenv_opt "DEKU_LOGS" |> Option.map String.lowercase_ascii with
+  | Some "json" -> Logs.set_reporter (Json_logs_reporter.reporter Fmt.stdout)
+  | _ -> Logs.set_reporter (Logs.format_reporter ~app:Fmt.stdout ()));
+  Logs.set_level level;
+  ()
+
+let setup_log = Term.(const setup_log $ Logs_cli.level ())
```
aguillon, line 201: I don't think we need both environment variables and command line options so this can get back to what it was before, @ulrikstrid?

ulrikstrid:
You can pick different levels for different sources, so we can only set higher level than info for the deku source

src/bin/deku_node.ml
ulrikstrid: I'm happy with only 1 way of setting it

d4hines:
Closing since this is implemented in #285

PR #631, opened 2022-05-19
fix: add snapshot logging back

d4hines:
## Problem
There's no way to tell when we hashed a new snapshot as of #604

## Solution

This is a hacky solution - just put it on the `app` log level. A better solution would be to understand how `Logs` actually works, and use Cmdliner to control it. I've started this in #630.

### DIFF
--- a/src/node/snapshots.ml
+++ b/src/node/snapshots.ml
@@ -52,7 +52,7 @@ let add_snapshot_ref ~block_height t =
       additional_blocks = t.additional_blocks;
     } )
 let set_snapshot_ref ref_ snapshot =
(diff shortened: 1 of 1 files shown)


src/node/snapshots.ml
```diff
@@ -52,7 +52,7 @@ let add_snapshot_ref ~block_height t =
       additional_blocks = t.additional_blocks;
     } )
 let set_snapshot_ref ref_ snapshot =
-  Logs.info (fun m ->
+  Logs.app (fun m ->
```
ulrikstrid, line 55: I think the correct solution is to set the logging level to `Info` instead

PR #632, opened 2022-05-23
fix the version of baking bad

Pilou97:
Problem:
Inside the `docker-compose.yml` we are using the "master" tag for baking-bad images which can lead to a non  reproductible dev environment.
The first time I did a `docker compose up`, docker pulled the `master` tag from baking-bad images which was pointing at that time to version 4.2.0 which is not working.

Solution:
Not using the master tag but a fixed version

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,7 +27,7 @@ version: "3.8"
 services:
   elastic:
     container_name: deku_elastic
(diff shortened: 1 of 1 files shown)


d4hines (approved):
LGTM! @Pilou97 if you remove draft mode I'll merge.

PR #633, opened 2022-05-23
poc: Add JSON output to deku-cli commands

ulrikstrid:
## Depends

- [ ] #580

## Problem

We are currently juggling `awk` etc when scripting to get the output we want which is error prone

## Solution

Add a `--output` flag so that we can choose to get JSON output which is easier to parse using as an example `jq`


src/tezos_interop/tezos_js_bridge.js
```diff
@@ -137,7 +137,9 @@ const read = (callback) => {
       return done();
     },
   });
-  const errorHandler = (err = "close") => failure(err);
+  const errorHandler = (err) => {
+    if (err != null) failure(err);
```
d4hines, line 141: What is the purpose of this change @ulrikstrid ?

d4hines (commented):
This looks great. I just have a question about the changes to the js_bridge.

nix/overlay.nix
```diff
@@ -19,34 +19,16 @@ in {
...
       alcotest = osuper.alcotest.overrideAttrs (o: {
         propagatedBuildInputs =
           prev.lib.lists.remove osuper.uuidm o.propagatedBuildInputs;
       });
-      qcheck = osuper.qcheck.overrideAttrs (_: {
-        src = prev.fetchurl {
-          url = "https://github.com/c-cube/qcheck/archive/v0.18.1.tar.gz";
-          sha256 = "1jzhwrzsf5290rs7hsa1my5yh1x95sh2sz92c4svd8yahzdlny7m";
-        };
-      });
```
aguillon, line 32: I don't usually follow what's going on with Nix scripts but I'm not sure why this was introduced in the first place, and why it has to go now? I believe qcheck is used in Pollinate and @gsebil08 and @rosalogia will probably use it more.

aguillon (changes requested):
Can you just list `jq` as a dependency for the sandbox script in the README please?

nix/overlay.nix
ulrikstrid: Moved this question to #580 where it's introduced

d4hines (approved):
Thanks for this guys.

ulrikstrid:
This needs to be rebased etc, but what do we want to do to silence other things than the json output?

d4hines:
@ulrikstrid, I'm not sure yet. I'm not super comfortable with the change to the tezos bridge - I want to understand why we're calling the error handler with `null` in the first place. I need to review it more thoroughly and test it. I'll look at this PR in more detail early next week.

src/tezos_interop/tezos_js_bridge.js
EduardoRFS: This callback should never be called, as js bridge is permanently lived. I think the only case where this could happen is if you're killing the node, in which case I'm not sure what is the right behavior.

And if this callback is called it should definitely kill the bridge, even if not with a failure.

I would like to avoid this, if the only argument pro it is that it supresses some noise

ulrikstrid: It's not forever lived when using the CLI and there is noise that wrecks the json which is the whole point of this PR @EduardoRFS

d4hines: Actually, we already are printing to stdout. So this PR confuses me a bit. I'm going to minimize it to the smallest change that lets us get rid of `sed` in sandbox.sh.

d4hines:
@ulrikstrid, I modified this PR to be much smaller. Let me know if you don't like it and we'll roll back. If this minimal change doesn't address problems you're having in Grafana etc. we can address in a follow up.

@aguillon, no need to mention jq in readme since the only supported way to run the sandbox is either `nix run .#sandbox` or `nix develop -c ./sandbox.sh` either of which will include everything required.

src/bin/deku_cli.ml
```diff
@@ -610,11 +623,19 @@ let produce_block node_folder =
       let open Network in
       broadcast_to_list (module Block_spec) validator_uris { block } in
     Format.printf "block.hash: %s\n%!" (BLAKE2B.to_string block.hash);
+    let () =
+      let hash_string = BLAKE2B.to_string block.hash in
+      match output with
+      | Human -> Format.printf "block.hash: %s\n%!" hash_string
+      | JSON ->
+        Format.printf "%a"
+          (Yojson.pretty_print ~std:true)
```
ulrikstrid, line 632: Why are we pretty printing if it's not for humans anyway?

ulrikstrid (commented):
If this works with jq I'm happy, we'll never see this output in grafana so it doesn't matter what it looks like as long as we can parse it without awk/sed etc 😅

aguillon (approved):
I don't think I understand the point of this PR but I see no reason of blocking it as such.

PR #634, opened 2022-05-23
fix: lint dummy_ticket.mligo

d4hines:
There are a few minor issues in dummy_ticket.mligo
- a type alias isn't used once
- some missing whitespace


aguillon (approved):

aguillon:
Oh no, I shouldn't have approved while the test was running!

PR #635, opened 2022-05-23
set networking_mode to "host"

d4hines:
The Prometheus server is broken when running in local mode
    because it is configured to connect with the targets running
    on localhost per prometheus.yml. By adding the
    "network_mode: host" option to our docker-compose.yml
    we can let prometheus talk to the targets in local mode as well as in the
    Tilt configuration

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -130,6 +130,7 @@ services:
     container_name: deku_prometheus
     restart: always
     image: prom/prometheus
+    network_mode: host
     ports:
       - "9090:9090"
     expose:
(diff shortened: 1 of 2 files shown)


aguillon:
I'm not qualified, so I'm naming @gsebil08 as a reviewer.

ulrikstrid:
Alternative fix in #615

d4hines:
@gsebil08, yes we do - it's how we do collect benchmarking data.

@ulrikstrid I'll have a look at that PR and close this one if I can.

ulrikstrid:
Would love to have a walkthrough of that PR with anyone that is interested @d4hines

d4hines:
#615 is better. Closing in favor it.

PR #638, opened 2022-05-24
Wasm vm integration

renatoalencar:
## Problem

We need to run WASM contracts on Deku

## Solution

Integrate wasm_vm into Deku core and CLI.

### DIFF
--- a/src/bin/deku_cli.ml
+++ b/src/bin/deku_cli.ml
@@ -139,13 +139,15 @@ let vm_flavor =
     (match string with
     | "Lambda" -> Some `Lambda
     | "Dummy" -> Some `Dummy
+    | "Wasm" -> Some `Wasm
     | _ -> None)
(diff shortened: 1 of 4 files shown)


src/bin/deku_cli.ml
```diff
@@ -276,7 +279,15 @@ let originate_contract node_folder contract_json initial_storage
           let int =
             try Yojson.Safe.Util.to_int initial_storage with
             | _ -> failwith "Invalid storage fro contract" in
-          Contract_vm.Origination_payload.dummy_of_yojson ~storage:int in
+          Contract_vm.Origination_payload.dummy_of_yojson ~storage:int
+        | `Wasm ->
+          let code, storage =
+            match (contract_program, initial_storage) with
+            | `String code, `String storage ->
```
renatoalencar, line 286: I am not really sure about encoding stuff as JSON for WASM. One would have to encode contracts and storage as JSON strings and that makes it harder to originate and invoke contracts.

PR #639, opened 2022-05-25
Build framework for profiling: deku node

lykimq:
Build a profiling framework.
1. Perf
Using Linux perf to get the profiling for the `sandbox.sh start`.
Reading perf.data with:
- Flamegraph
- Hotspot
- Gprof2dot
2. Valgrind using massif
3. Documentation


lykimq:
> 



> Are the `svg`s in `benchmarks/profiling` intended to be checked in?

What do you mean checked in? I push those `svg` just as an example.

InfiniteSwerve (changes requested):
One suggestion, but otherwise things look good to me.

flamegraph.rb
```diff
@@ -0,0 +1,31 @@
...
+# - Copy perf.data generated in Deku to FlameGraph folder* 
+# - Go inside FlameGraph folder and use `perf script` to read the information inside
+# perf.data
+# - Use flamegraph commands and save the result in picture as `perf-flamegraph.svg`
+# - Wait for awhile to let the graph generated, you can use internet browser to see the file
+# `perf-flamegraph.svg`:
+#  + firefox perf-flamegraph.svg # or chrome, etc.
+#
+# Note:
+# FlameGraph need to be downloaded at:
```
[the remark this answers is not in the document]
lykimq: I am not sure if it is a good idea to include the code of flame-graph inside, because it can be problematic to maintain. 
cc: @aguillon what do you think about it?

benchmarks/profiling/profile_deku.sh
```diff
@@ -0,0 +1,121 @@
...
+        && perf record -F 99 -a -g --call-graph dwarf ./sandbox.sh start \
+        && sleep 10
+}
+
+# Read perf.data using FlameGraph
+flame_graph(){
+  cp ~/deku/perf.data ~/FlameGraph/perf.data \
+      && cd ~/FlameGraph && perf script \
+          | ./stackcollapse-perf.pl \
+          | ./flamegraph.pl > ~/deku/benchmarks/profiling/perf-deku-flamegraph.svg
```
aguillon, line 32: What are those paths and scripts? They're not included in the PR.

benchmarks/profiling/readme.md
```diff
@@ -0,0 +1,177 @@
...
+Run this script only when `perf record` is **properly terminated** (a.k.a: `./profile_deku.sh run`).
+
+```
+./profile_deku.sh flame-graph
+```
+
+Wait for awhile to let the graph generated. The result in picture as `perf-deku-flamegraph.svg`. You can use internet browser to see it `firefox perf-deku-flamegraph.svg` or chrome, etc. 
+
+This is the generated flame-graph of deku-node running with 1012 blocks:
+[flame-graph.svg](https://github.com/lykimq/deku_docs/blob/main/benchmarking/graphs/perf-deku-flamegraph.svg)
```
aguillon, line 56: Uuuh
* I thought it was for 10 seconds?
* most functions have no name here; how do we fix that?
* the only functions that have a name… appear to be related to your desktop. Can we filter them out?

I understand that you don't clean the data after it has been produced, however that's something we should do (and the scripts should be in the PR as well).

benchmarks/profiling/readme.md
```diff
@@ -0,0 +1,177 @@
...
+
+
+#### Understand Flamegraph
+It shows: 
+- x-axis: show the stack profile (the sample) population, sorted alphabetically
+- y-axis: stack depth, counting from zero at the bottom.
+- Each function (stack frame) is drawn as a rectangle, with the width relative to the number of samples. The wider a frame is, the more often it was present in the stacks.
+- The top edge shows what is on-CPU.
+- The beneath it is its ancestry.
```
aguillon, line 65: I think I understand: x-axis is the aggregated time spent in each function; as such, we don't know how often we called the function. Might be that it was inefficient, might be that it's just the mainloop (but I reckon that it is already enough info to know which part of the code which should optimize).

benchmarks/profiling/readme.md
```diff
@@ -0,0 +1,177 @@
...
+ sudo apt-get install python3-pip
+ sudo pip install gprof2dot
+```
+
+Wait for awhile to let the graph generated. The result in picture as `perf-deku-callgraph.png`. You can use internet browser to see it `firefox perf-deku-call-graph.png` or chrome, etc.
+
+This is the generated call-graph of deku-node running with 1012 blocks: 
+[call-graph.svg](https://github.com/lykimq/deku_docs/blob/main/benchmarking/graphs/perf-deku-callgraph.svg)
+
+Note that, it may happen that you cannot see this graph on a browser of Github, you can download this file and open it on local.
```
aguillon, line 87: You can remove this (you already said the first two paragraphs, the last line is not necessary IMO).

sandbox.sh
```diff
@@ -121,8 +121,7 @@ trusted_validator_membership_change_json() {
     i=$(echo "$VALIDATOR" | awk -F';' '{ print $1 }')
     ADDRESS=$(echo "$VALIDATOR" | awk -F';' '{ print $4 }')
     if [ "$i" != 0 ]; then
-      printf ",
-"
+      printf ","
```
aguillon, line 124: Why? We should probably use jq to pretty print this part instead of printf, but I'm not sure why you change this.

aguillon (changes requested):
The script is rather clear, thank you for taking the time to write all the documentation.

I haven't tried to run the scripts yet (but apparently I need some scripts that this PR doesn't provide). I think the PR should not be merged until we can all run it on our systems, even if it needs some tinkering beforehand. Don't feel compelled to rebase it yet, this is something we can do when merging.

Could you also please put the perf.data somewhere so I can take a look at the format?

flamegraph.rb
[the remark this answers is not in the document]
aguillon: I think including the code directly is a bad idea (for licensing and versioning reasons). It seems that it is a single file with no dependency, right? You can download it automatically pretty easily if it's missing from your main script.

benchmarks/profiling/profile_deku.sh
```diff
@@ -0,0 +1,121 @@
...
+#
+# **Note:**
+# The rate for perf is 99 Hertz, the generated FlameGraph from a 1000 Hertz profile.
+# Choosing 99 Hertz instead of 100 Hertz, is to avoid accidentally sampling in 
+# lockstep with some periodic activity, which would produce skewed results.
+# You can also increase to higher rates (eg, up to 997 Hertz) for finer resolution.
+# Bear in mind that higher frequencies means higher overhead
+
+run(){
+  cd .. && cd .. \
```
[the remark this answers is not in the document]
lykimq: About the `perf.data` it is quite heavy to included in (for ~1000 block, it is around 3GB) and I can't add it in Git. I can give a light example with fewer nodes, but not sure if the data output is useful.

benchmarks/profiling/profile_deku.sh
```diff
@@ -0,0 +1,121 @@
...
+        && ./sandbox.sh tear-down \
+        && ./sandbox.sh setup \
+        && perf record -F 99 -a -g --call-graph dwarf ./sandbox.sh start \
+        && sleep 10
```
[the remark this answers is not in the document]
lykimq: It is just a random choice.

sandbox.sh
lykimq: I just move the `"` up next to the `printf ",`.

PR #643, opened 2022-05-26
chore: remove x86_64-darwin from CI

d4hines:
No one on the team is using x86_64-darwin, and
it's slowing down our CI. Let's remove it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        system: [ubuntu, macos]
+        system: [ubuntu]
 
     runs-on: ${{ matrix.system }}-latest
 
--- a/flake.nix
+++ b/flake.nix
@@ -49,8 +49,7 @@
   outputs = { self, nixpkgs, flake-utils, nix-filter, dream2nix, ocaml-overlays
     , prometheus-web, tezos, json-logs-reporter }:
     let
-      supportedSystems =
-        [ "x86_64-linux" "x86_64-darwin" "aarch64-linux" "aarch64-darwin" ];
+      supportedSystems = [ "x86_64-linux" "aarch64-linux" "aarch64-darwin" ];
 
       dream2nix-lib = dream2nix.lib2.init {
         systems = supportedSystems;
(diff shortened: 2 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        system: [ubuntu, macos]
+        system: [ubuntu]
```
ulrikstrid, line 15: Can we change this to `ubuntu-latest` while we're at it?

ulrikstrid (approved):
Do we have to make it explicit that we don't officially support `x86_64-darwin` in the readme?

.github/workflows/test.yml
d4hines: Isn't it already running on latest via `${{ matrix.system }}-latest` on the next line?

ulrikstrid: You are correct, missed that :facepalm:

flake.nix
```diff
@@ -49,8 +49,7 @@
   outputs = { self, nixpkgs, flake-utils, nix-filter, dream2nix, ocaml-overlays
     , prometheus-web, tezos, json-logs-reporter }:
     let
-      supportedSystems =
-        [ "x86_64-linux" "x86_64-darwin" "aarch64-linux" "aarch64-darwin" ];
+      supportedSystems = [ "x86_64-linux" "aarch64-linux" "aarch64-darwin" ];
```
ulrikstrid, line 52: I will try to setup my hydra for `aarch64-linux` soon so that we can actually test that workflow somewhere

EduardoRFS:
This is not a good change, while no one is using x86_64-darwin a bunch of people are running arm64-darwin, which was tested in part by x86_64-darwin.

EduardoRFS:
Also the only slow down of it was using an additional github actions instance, as it did not prevent merging.

ulrikstrid:
@EduardoRFS there was a bunch of errors, both with esy (problem compiling OCaml?) and nix (installing the nix action have 403 frequently).
aarch64-darwin is tested by hydra, even though it's not official (yet) it's an actual test of the architecture.

EduardoRFS:
@ulrikstrid If the reason to remove it was because it was failing randomly, that's an even worse excuse to be removing support to it, it was an evidence of something being broken.

So instead of fixing the problem on nix, that someone may hit we decided to just drop the architecture.

PR #644, opened 2022-05-26
feat: block by level RPC

d4hines:
Successor to #536

## Problem

There is no way to inspect the contents of blocks from outside a deku node.

## Solution

Add a `block-by-level` rpc that takes a block level as an int and returns the block with that block height, or `null` if it doesn't know the contents of that block.

This is a workaround to the fact we have no indexer, see #535. But it's pretty essential for using Deku at all, so I think the work-around is justified for now.

### DIFF
--- a/src/bin/deku_node.ml
+++ b/src/bin/deku_node.ml
@@ -68,12 +68,26 @@ let handle_block_by_hash =
       Ok block)
 
 (* POST /block-level *)
-(* Retrieve height of the chain? *)
+(* Retrieve height of the chain *)
 let handle_block_level =
   handle_request
     (module Network.Block_level)
     (fun _update_state _request ->
       Ok { level = Flows.find_block_level (Server.get_state ()) })
+
+(* POST /block-by-level *)
+(* Retrieves the block at the given level if it exists *)
+let handle_block_by_level =
+  handle_request
+    (module Network.Block_by_level_spec)
+    (fun _update_state request ->
+      let state = Server.get_state () in
+      let block =
+        List.find_opt
+          (fun block ->
+            Int64.equal block.Protocol.Block.block_height request.level)
+          state.applied_blocks in
+      Ok block)
 
 (* POST /protocol-snapshot *)
 (* Get the snapshot of the protocol (last block and associated signature) *)
@@ -176,6 +190,7 @@ let node folder prometheus_port =
              handle_received_block_and_signature;
              handle_received_signature;
              handle_block_by_hash;
+             handle_block_by_level;
              handle_protocol_snapshot;
              handle_request_nonce;
              handle_register_uri;
--- a/src/network/network_packet.ml
+++ b/src/network/network_packet.ml
@@ -35,6 +35,14 @@ module Block_by_hash_spec = struct
   let path = "/block-by-hash"
 end
 
+module Block_by_level_spec = struct
+  type request = { level : int64 } [@@deriving yojson]
+
+  type response = Block.t option [@@deriving yojson]
+
+  let path = "/block-by-level"
+end
+
 module Block_level = struct
   type request = unit [@@deriving yojson]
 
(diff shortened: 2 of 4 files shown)


src/bin/deku_node.ml
```diff
@@ -69,12 +69,27 @@ let handle_block_by_hash =
 
 (* POST /block-level *)
 (* Retrieve height of the chain? *)
+
```
gsebil08, line 72: I guess this empty line should be removed.
By the way, would you mind delete the `?` on the previous line?

src/node/state.ml
```diff
@@ -25,6 +25,10 @@ type t = {
   data_folder : string;
   pending_operations : timestamp Operation_map.t;
   block_pool : Block_pool.t;
+  (* TODO: we need a bound on the size of this and put
+     behind an abstract type. We should also change how
+     this works once we have an indexer. See https://github.com/marigold-dev/deku/issues/535 *)
+  applied_blocks : Block.t list;
```
gsebil08, line 31: Naive question, why a list and not a Set?

gsebil08 (commented):
Format + naive question, seems ok

src/node/state.ml
d4hines: Semantically it's a more like set, but in this case switching to `Block.t Set.t` won't buy us any additional performance. The right thing to implement a data structure with an upper bound on size and that is indexed on the dimension we access it from (block level), like a `Block.t (Map.Make (Int64)).t`. But a better thing to do is to implement the indexer per the linked issue.

gsebil08: >Semantically it's a more like set

Yes, that is exactly what I understood!

>in this case switching to Block.t Set.t won't buy us any additional performance

Is there any benchmark about `Set` vs `List` in OCaml?
What about using Array instead of List, I think I read something about Array is 7% more performant that List.

>But a better thing to do is to implement the indexer per the linked issue.

Definitively agreed.

src/bin/deku_node.ml
```diff
@@ -68,13 +68,27 @@ let handle_block_by_hash =
...
+    (module Network.Block_by_level_spec)
+    (fun _update_state request ->
+      let state = Server.get_state () in
+      let block =
+        List.find_opt
+          (fun block ->
+            Int64.equal block.Protocol.Block.block_height request.level)
+          state.applied_blocks in
+      Ok block)
+
```
aguillon, line 91: I'm a bit annoyed by the fact that we use two terms, "level" and "height", to designate the same thing. Can we decide which one we want to use once and for all? Switching to `level` seems acceptable as it is the term used in Tezos, keeping `height` is OK as well because inertia, also the term seems to be more common outside of Tezos.

src/node/state.ml
aguillon: @gsebil08 Array will be more costly when we commit a block, a list makes sense regarding this.

aguillon (changes requested):
I'm not happy with the naive usage of the list here, give me a minute

src/network/network_packet.ml
```diff
@@ -35,6 +35,14 @@ module Block_by_hash_spec = struct
   let path = "/block-by-hash"
 end
 
+module Block_by_level_spec = struct
```
aguillon, line 38: Why "_spec"?

src/bin/deku_node.ml
d4hines: This seems out of scope for this PR. I'm open to a PR that standardizes usage across the code base though.

src/network/network_packet.ml
aguillon: Oh, true ahah. That's also something that we could standardize a bit.

aguillon (approved):
I was wrong, everything is ok.

d4hines:
The CI is flaky on the e2e tests. I tested locally and it passes. Merging.

PR #645, opened 2022-05-26
Fix number of nodes launched by sandbox.sh

d4hines:
## Problem

We have an off-by-1 that causes us to launch n+1 nodes.

## Solution

Fix it.


ulrikstrid (commented):
I think this needs changes in tilt since I was too lazy to fix this when implementing the initial thing

ulrikstrid (approved):

PR #646, opened 2022-05-26
fix: use absolute path for PATH in shell.nix

d4hines:
## Depends
- [ ] #645 

## Problem

Our `PATH` is relative in shell.nix, which breaks things when you're working in deep directories.

## Solution
Use absolute path.


gsebil08 (approved):
Seems great, don't face the trouble, so this is a theoretical approval :)

PR #647, opened 2022-05-26
[d4hines] feat: block rate and transaction rate metrics  d4hines/block-rate-and-tps

d4hines:
Successor to #537 

## Depends

- [ ] #644 

## Problem

We want metrics for block rate and operations per block (from which you can derive operations rate - i.e tps - in prometheus dashboard)

## Solution

Adds these metrics.

### Stacked PR Chain: d4hines
|#647|👉feat: block rate and transaction rate metrics  d4hines/block-rate-and-tps||#691|

### DIFF
--- a/src/bin/deku_node.ml
+++ b/src/bin/deku_node.ml
@@ -19,6 +19,9 @@ let handle_request (type req res)
   let handler request =
     let%await body = Dream.body request in
     let%await request = Parallel.decode E.request_of_yojson body in
+    Metrics.Networking.inc_network_messages_received E.path;
+    Metrics.Networking.measure_network_received_message_size E.path
+      (String.length body);
     match request with
     | Ok request -> (
       let response = handler update_state request in
--- a/src/metrics/blocks.ml
+++ b/src/metrics/blocks.ml
@@ -12,3 +12,10 @@ let blocks_signed_total =
 let inc_block_produced () = Prometheus.Counter.inc_one blocks_produced_total
 
 let inc_block_signed () = Prometheus.Counter.inc_one blocks_signed_total
+
+let operations_processed_total =
+  let help = "The total number of operations processed by the node" in
+  Counter.v ~help ~namespace ~subsystem "operations_processed_total"
+
+let inc_operations_processed ~operation_count =
+  Counter.inc operations_processed_total (Float.of_int operation_count)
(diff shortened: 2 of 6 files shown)


src/metrics/throughput.ml
```diff
@@ -0,0 +1,22 @@
...
+let previous_timestamp = ref None
+
+let collect_block_metrics ~timestamp ~operation_count =
+  Gauge.set operations_per_block (Float.of_int operation_count);
+  match !previous_timestamp with
+  | Some previous_timestamp ->
+    let delta = timestamp -. previous_timestamp in
+    Gauge.set block_rate delta
+  | None ->
+    ();
```
gsebil08, line 21: I am confused, the semicolon make the following instruction the last one, right?
But why are they in this order instead of:
```ocaml
  | None ->
    previous_timestamp := Some timestamp;
    ()
```
🤔

src/node/state.ml
```diff
@@ -84,6 +84,11 @@ let apply_block state block =
       (fun results (hash, receipt) -> BLAKE2B.Map.add hash receipt results)
       state.recent_operation_receipts receipts in
   let applied_blocks = block :: state.applied_blocks in
+  (* TODO: we might want to use a monotonic clock so we don't get weird things like
+     negative block rates. *)
+  let timestamp = Unix.time () in
```
gsebil08, line 89: Is second precise enough?

src/node/state.ml
```diff
@@ -84,6 +84,11 @@ let apply_block state block =
...
+  let operation_count = List.length block.operations in
+  Metrics.Througput.collect_block_metrics ~timestamp ~operation_count;
```
gsebil08, line 91: Any benchmarking about adding this?

gsebil08 (commented):
Seems ok, but still several questions 😇

src/metrics/throughput.ml
d4hines: Closing because this code is removed.

src/node/state.ml
d4hines: We've changed this to just increment a counter every 1 second - there should be extremely little performance overhead.

src/node/state.ml
InfiniteSwerve: We care about TPS (transactions per second), so this will probably be fine.

lykimq (approved):

aguillon (approved):
LGTM

PR #648, opened 2022-05-26
chore: refactor cli terms

d4hines:
## Depends
- [ ] #539 

## Problem

There's some duplicated logic in the CLI - this became apparent as I added a few more commands in deku-p.

Specifically:
1. We want to abstract retrieving the validator URI's out into it's own function.
2. We want to fix how we're defining positional arguments. Currently we define them all over the place, very far removed from where we actually build up the `Term.t` used by Cmdliner.

## Solution
1. Create a `with_validators` function that accepts a continuation that provides the validator uri's.
2. Turn our positional arguments into functions that accept the position and define the position where they're actually used.

Re: 1. Why not create a function just returns a promise for the validators? Because we want to define some exit behavior if we can't find them, and using a continuation allows us to do that all in one place.

### DIFF
--- a/src/bin/deku_cli.ml
+++ b/src/bin/deku_cli.ml
@@ -219,9 +219,7 @@ let info_create_transaction =
   Cmd.info "create-transaction" ~version:"%\226\128\140%VERSION%%" ~doc ~exits
     ~man
 
-let create_transaction node_folder sender_wallet_file received_address amount
-    ticket argument vm_flavor tickets =
-  let open Network in
+let with_validator_uri node_folder f =
   let%await interop_context = interop_context node_folder in
   let%await validator_uris = validator_uris ~interop_context in
   match validator_uris with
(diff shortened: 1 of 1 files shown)


src/bin/deku_cli.ml
```diff
@@ -220,45 +218,46 @@ let create_transaction node_folder sender_wallet_file received_address amount
...
-      Lwt.return (`Ok ()))
+let create_transaction node_folder sender_wallet_file received_address amount
+    ticket argument vm_flavor =
+  let open Network in
+  with_validator_uri node_folder @@ fun (_, validator_uri) ->
+  let%await block_level_response = request_block_level () validator_uri in
+  let block_level = block_level_response.level in
+  let%await wallet = Files.Wallet.read ~file:sender_wallet_file in
+  let operation =
+    match (Address.to_key_hash received_address, argument) with
```
gsebil08, line 231: I guess `create_transaction` should be splitted into two different functions.
We only do things if `received_address` is an Implicit address and `argument` is None.
And if `argument` is `Some argument` and `received_address` is None...
I don't know if I am clear enough, but I think we are trying to handle two different business cases in the same function.

src/bin/deku_cli.ml
```diff
@@ -358,21 +366,13 @@ let originate_contract =
   Term.(
     lwt_ret
       (const originate_contract
-      $ folder_node
+      $ folder_node 0
```
gsebil08, line 369: I don't understand why `0` 😕 ...

d4hines: Because it's a positional argument `Cmdliner.Arg.pos`, which requires us to specify the expected position of the argument. in this case it's the 0th argument.

src/bin/deku_cli.ml
```diff
@@ -220,45 +218,46 @@ let create_transaction node_folder sender_wallet_file received_address amount
...
-          identity.uri in
-      Format.printf "operation.hash: %s\n%!"
-        (BLAKE2B.to_string transaction.hash);
-      Lwt.return (`Ok ()))
+let create_transaction node_folder sender_wallet_file received_address amount
+    ticket argument vm_flavor =
+  let open Network in
+  with_validator_uri node_folder @@ fun (_, validator_uri) ->
+  let%await block_level_response = request_block_level () validator_uri in
+  let block_level = block_level_response.level in
```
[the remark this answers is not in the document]
d4hines: Just following the convention used elsewhere in this module

src/bin/deku_cli.ml
d4hines: Sorry, I gave a very poor PR description. I updated it to make it clear exactly what I intend to refactor in this PR. That said, your proposal is outside my intended scope, but it's probably a good one to do. PR's accepted :smile:

PR #649, opened 2022-05-26
[parametric] feat: add go sdk

d4hines:
## Depends

- [ ] #540 

## Problem

This PR adds a Go SDK for Deku interop. It's a direct translation of the JS SDK.

### DIFF
--- a/sdks/deku_go_interop/interop.go
+++ b/sdks/deku_go_interop/interop.go
@@ -1,0 +1,118 @@
+package deku_go_interop
+
+import (
+	"encoding/binary"
+	"encoding/json"
+	"fmt"
+	"os"
+	"strings"
+)
+
+var chain_to_machine *os.File
+var machine_to_chain *os.File
+
+type chain_message struct {
+	Type    string
+	Payload interface{}
+}
+
+func check(e error) {
+	if e != nil {
+		panic(e)
+	}
+}
+
+func log(message string) {
+	// TODO: implement debug mode logging.
+	if false {
+		colored := fmt.Sprintf("\x1b[%dm%s\x1b[0m", 31, message)
+		fmt.Fprintln(os.Stderr, colored)
+	}
+}
+
+func read() []byte {
+	// read the number of bytes being sent
+	b := make([]byte, 8)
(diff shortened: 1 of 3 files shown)


sdks/deku_go_interop/interop.go
```diff
@@ -0,0 +1,118 @@
...
+	// TODO: implement debug mode logging.
+	if false {
+		colored := fmt.Sprintf("\x1b[%dm%s\x1b[0m", 31, message)
+		fmt.Fprintln(os.Stderr, colored)
+	}
+}
+
+func read() []byte {
+	// read the number of bytes being sent
+	b := make([]byte, 8)
```
gsebil08, line 35: I am not fond of one-letter variables...
Is it possible to extract `8` to a constante and use it here and in the `write`?

sdks/deku_go_interop/interop.go
```diff
@@ -0,0 +1,118 @@
...
+	_, err := chain_to_machine.Read(b)
+	check(err)
+	n := binary.LittleEndian.Uint16(b)
+	value := make([]byte, n)
+	bytes_read, err := chain_to_machine.Read(value)
```
gsebil08, line 40: Shouldn't you also do `check(err)` right after?

sdks/deku_go_interop/interop.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func Main(initial_state map[string]interface{}, state_transition func(sender string, tx_hash string, input []byte) (err error)) {
+	log("Opening read")
+	fifo_path := os.Args[1]
+	log(fmt.Sprintf("fifo path: %s", fifo_path))
+	machine_to_chain, _ = os.OpenFile(fifo_path+"_read", os.O_WRONLY, 0666)
+	log("Opening write")
+	chain_to_machine, _ = os.OpenFile(fifo_path+"_write", os.O_RDONLY, 0666)
+	log("done")
```
gsebil08, line 84: How about:
```go
log("Successfully opened read and write")
```

gsebil08 (approved):

PR #650, opened 2022-05-26
[parametric] Parametric go example

d4hines:
## Depends

- [ ] #649 

## Problem

We need some example Go code. 

## Solution

Adds a counter example that mirrors the JS example.


examples/go-counter/main.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/marigold-dev/deku/sdks/deku_go_interop"
+)
+
+type message struct {
```
gsebil08, line 12: What's the point of having a `struct` with only one field? Isn't it ok to use
```go
type message string
```
?

examples/go-counter/main.go
```diff
@@ -0,0 +1,64 @@
...
+				tx_hash,
+				string(input)))
+		err := json.Unmarshal(input, &message)
+		check(err)
+		counter_bytes := deku_go_interop.Get("counter")
+		var counter *int
+		json.Unmarshal(counter_bytes, &counter)
+		var new_counter int
+		switch message.Action {
+		case "Increment":
```
gsebil08, line 42: Why pattern-matching on a string? Isn't it possible to use a sum-type? _(I don't know go, this is a naive question)_

gsebil08 (commented):
Why did I read go?...

examples/go-counter/main.go
d4hines: Go does not have easy sum types.

examples/go-counter/main.go
Pilou97: Because the message is a json like message, and it has the following shape:
```json
{
"Action": "Increment"
}
```

oteku, line 12: Whaever is our opinon on Go  🤫
I think we should use the go-ish way to manage enumeration in our exemple... only for marketing purpose

```go
type Action int

const (
	Decremenet Action = iota+1
	Increment 
)

type Message struct {
	Action Action
}
```

PR #652, opened 2022-05-27
Fix contract originated address

renatoalencar:
## Problem

Contract origination is using the  hash of the `User_operation`, which is the same for the same `sender * code * storage` triple. The would make possible to overwrite a contract's storage with it's initial value, also making not possible for an implicit account to have multiple contracts with same initial state.

## Solution

Use the protocol operation hash instead, which depends on stuff like block height and nonce. Also adds an endpoint for making possible to test it.

### DIFF
--- a/src/core_deku/state.ml
+++ b/src/core_deku/state.ml
@@ -35,9 +35,9 @@ let apply_tezos_operation t tezos_operation =
   let { tezos_operation_hash = _; internal_operations } = payload in
   List.fold_left apply_internal_operation t internal_operations
 
-let apply_user_operation t user_operation =
+let apply_user_operation t operation_hash user_operation =
   let open User_operation in
-  let { source; initial_operation; hash } = user_operation in
+  let { source; initial_operation; hash = _ } = user_operation in
   let { ledger; contract_storage } = t in
   match initial_operation with
   | Transaction { destination; amount; ticket } ->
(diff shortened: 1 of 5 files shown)


src/bin/deku_node.ml
```diff
@@ -158,6 +158,23 @@ let handle_ticket_balance =
       let amount = Flows.request_ticket_balance state ~ticket ~address in
       Ok { amount })
 
+(* POST /get-storage *)
```
d4hines, line 161: This additional endpoint seems unrelated to the goal of the PR. I also don't think we want to expose Core_deku state via the Deku node since this conflicts with the architecture design of Deku p (potentially). See #535.

d4hines (changes requested):

src/core_deku/state.ml
```diff
@@ -97,9 +97,9 @@ let apply_user_operation t user_operation =
...
 
-let apply_user_operation t user_operation =
-  match apply_user_operation t user_operation with
+let apply_user_operation t hash user_operation =
+  match apply_user_operation t hash user_operation with
   | Ok (t, receipt) -> (t, receipt)
   (* TODO: use this erros for something *)
   | Error (`Origination_error _ | `Invocation_error _) -> (t, None)
-  | Error `Not_enough_funds -> (t, None)
+  | Error (`Insufficient_funds | `Not_enough_funds) -> (t, None)
```
d4hines, line 105: Looks like we added an extra error here? `Insufficient` is synonym to `Not_enough` so we ought not have both.

renatoalencar: Sorry, I was tricked by my local environment mentioning the `this pattern is not exhaustive error`. Fixed.

d4hines (approved):
LGTM

PR #653, opened 2022-05-27
Fix landmarks dependency for a version that supports aarch64

renatoalencar:
## Problem

Project is not compiling on Mac M1 because of a dependency that doesn't support aarch64.

## Solution

Add esy resolution with support for it, I have made a fork and updated it for support for aarch64.

https://github.com/renatoalencar/landmarks/commit/91f75a5ddc38e40bcd8f1dd10b68e88a5c169e0d


d4hines (approved):
Nice, thanks for this!

PR #656, opened 2022-05-31
fix: override jsonrpc to a working commit

d4hines:
## Problem

Our nix setup isn't working because a commit was removed that was referenced from https://github.com/anmonteiro/nix-overlays. 

## Solution

@ulrikstrid created a working patch in his fork.


ulrikstrid:
Fix in the overlays: https://github.com/anmonteiro/nix-overlays/pull/325

ulrikstrid (approved):
This is OK, I'm just about to merge the PR in the upstream overlays so we could use that instead. But this fixes the problem NOW

PR #657, opened 2022-05-31
fix tilt file

Pilou97:
# Problem:
Tilt configuration does not work on deku parametric


d4hines:
Cherry-picked, thanks.

PR #661, opened 2022-06-01
Zett98/wasm ffi

Zett98:
## Problem

We need to have ffi for wasm-vm to be able to manipulate tickets/interact with deku itself
## Solution

Add the ffi and change the wasm-vm interface to fit the ffi


src/bin/deku_cli.ml
```diff
@@ -184,7 +185,7 @@ let info_create_transaction =
   Term.info "create-transaction" ~version:"%\226\128\140%VERSION%%" ~doc ~exits
     ~man
 let create_transaction node_folder sender_wallet_file received_address amount
-    ticket argument vm_flavor =
+    ticket argument vm_flavor tickets =
```
renatoalencar, line 188: Why not merging `ticket` and `tickets` into one single argument and probably add constraints to it later on the code? Having two parameters with almost the same and different purposes makes things a little bit confuse at least.

src/core_deku/contract_context.ml
```diff
@@ -0,0 +1,155 @@
...
+  module S = Helpers.Set.Make (Ticket_handle)
+  let get_or_raise t =
+    Result.map_error
+      ~f:(fun e ->
+        Format.asprintf "%a" Ticket_table.Errors.pp (e :> Ticket_table.Errors.t))
+      t
+    |> Result.ok_or_failwith
+
+  let make ~sender ~table ~self ~source ~tickets ~contracts_table =
+    (module struct
```
renatoalencar, line 44: Why so much mutable state? Why not doing more like Michelson and having stuff as immutable values? It'd be a lot better for testing.

src/core_deku/contract_ffi.ml
```diff
@@ -1,55 +1,172 @@
...
+  | 0 -> load mem x S.self SELF
+  | 1 -> load mem x S.source SOURCE
+  | 2 -> load mem x S.sender SENDER
+  | 3 -> load mem x S.read_ticket READ_TICKET
+  | 4 -> load mem x S.split_ticket SPLIT_TICKET
+  | 5 -> load mem x S.own_ticket OWN_TICKET
+  | 6 -> load mem x S.join_tickets JOIN_TICKETS
+  | 7 -> load mem x S.get_contract_opt GET_CONTRACT_OPT
+  | 8 -> load mem x S.transaction TRANSACTION
+  | _ -> failwith "unimplemented"
```
renatoalencar, line 172: failing would make possible for an external agent to crash the node by running a unimplemented syscall.

renatoalencar (changes requested):
I think that's mostly great, I have just a few questions for now. Let's focus on that first and iterate after that.

src/bin/deku_cli.ml
Zett98: We will be using one ticket to pay for fees, while others are passed in contract argument positions  so we need to track their ownership somehow

src/core_deku/contract_context.ml
Zett98: Its localised per one contract call, so mutable state is not a problem, because it will be used only once per invocation and state should be kept somewhere should be kept somewhere

src/core_deku/contract_ffi.ml
Zett98: We catch all errors inside runtime.ml of wasm_vm library

src/bin/deku_cli.ml
renatoalencar: Right, I think we could make use of more descriptive parameter name.

src/core_deku/contract_context.ml
renatoalencar: All right

src/core_deku/contract_ffi.ml
renatoalencar: So that should be a defined exception and not a failure

Zett98: what's the point of defining an exception if the only thing we can do about it is to abort execution?

src/bin/deku_cli.ml
Zett98: we will change it in June when we will have normal argument parsing for contracts

renatoalencar (approved):

PR #663, opened 2022-06-02
Stop when we find the block

aguillon:
## Depends

- https://github.com/marigold-dev/deku/pull/644

## Problem

Same but with a return type and early stopping. We said we were OK with using Base, right? If not, we can always use an exception.

Feel free to just say "no" and close.

## Testing

` echo $(wget -q http://localhost:4440/block-by-level --post-data '{"level":4115}' -O -)`

### DIFF
--- a/src/node/flows.ml
+++ b/src/node/flows.ml
@@ -417,6 +417,21 @@ let find_block_by_hash state hash =
 
 let find_block_level state = state.State.protocol.block_height
 
+let find_block_by_level state level =
+  let open Base.Continue_or_stop in
+  let open Base.Int64 in
+  let open Protocol.Block in
+  let open Network.Block_by_level_spec in
+  Base.List.fold_until state.State.applied_blocks ~init:Not_found
+    ~f:(fun _accum block ->
+      if level > block.block_height then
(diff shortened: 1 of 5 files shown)


aguillon:
Now that I think of it, you will probably find the return type annoying to parse with bash.

src/node/flows.ml
```diff
@@ -417,6 +417,21 @@ let find_block_by_hash state hash =
 
 let find_block_level state = state.State.protocol.block_height
 
+let find_block_by_level state level =
+  let open Base.Continue_or_stop in
+  let open Base.Int64 in
+  let open Protocol.Block in
+  let open Network.Block_by_level_spec in
+  Base.List.fold_until state.State.applied_blocks ~init:Not_found
```
d4hines, line 425: What's the advantage of `Base.List.fold_until` over something like this:
```
List.find_opt
          (fun block ->
            Int64.equal block.Protocol.Block.block_height request.level)
          state.applied_blocks)
|> Option.to_result ~none:Not_found 
```

aguillon:
Closed after discussion on Slack (not that useful).

PR #664, opened 2022-06-02
chore: use extra-substituters since it's addative

ulrikstrid:
## Problem

We are currently seeing weirdness in our cache

## Solution

Stop overriding the caches, instead we want to add *extra* substituters


d4hines (approved):
LGTM

PR #667, opened 2022-06-02
chore: remove matrix syntax from github action yml

d4hines:
## Problem

We only ever test on `ubuntu-latest`, so we don't need the "matrix" syntax. I'm also curious if this will fix the Github Action problem I'm encountering.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -6,11 +6,6 @@ on:
       - main
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


d4hines:
This didn't have the effect I was hoping for :( Will probably close this PR.

d4hines:
Closing since we found the root issue: we CI in another Marigold repository that was hogging all resources.

PR #658, opened 2022-05-31
[d4hines] feat: batches of operations

d4hines:
## Depends

- [ ] #647 

## Problem

We currently don't have any way for users to send batch operations. This kills our TPS, especially since our networking layer is not optimized currently.

## Solution

Add a `User_operations_gossip` network message (note the plural tense).

### Stacked PR Chain: d4hines
| PR | Title | Status |  Merges Into  |
|:--:|:------|:-------|:-------------:|
|#691|fix: fix prometheus with tilt||-|
|#647|feat: block rate and transaction rate metrics  d4hines/block-rate-and-tps||#691|
|#658|👉chore: refactor cli terms                      d4hines/batch-operations||#647|
|#681|chore: refactor config out of node state||#658|
|#668|make minimum block time configurable||#658|
|#539|Benchmarking: ticket transfers                 d4hines/ticket-transfer-benchmarking||#658|

PR #668, opened 2022-06-02
[d4hines] make minimum block time configurable

d4hines:
## Depends
- [ ] #648 

## Problem

Right now, Deku throttles itself to produce blocks every second. Users may want to configure this throttle.

Update:
As @EduardoRFS suggested in #692, this PR additionally implements a short-circuit to the delay if there are any pending operations and changes the default block delay to 5 seconds.

## Solution
Add an option at the command line to configure the throttle.

### Stacked PR Chain: d4hines
|#668|👉make minimum block time configurable||#658|

### DIFF
--- a/src/bin/deku_node.ml
+++ b/src/bin/deku_node.ml
@@ -178,8 +178,10 @@ let handle_ticket_balance =
           ~address:(Core_deku.Address.of_key_hash address) in
       Ok { amount })
 
-let node folder port prometheus_port =
-  let node = Node_state.get_initial_state ~folder |> Lwt_main.run in
+let node folder port minimum_block_delay prometheus_port =
+  let node =
+    Node_state.get_initial_state ~folder ~minimum_block_delay |> Lwt_main.run
+  in
   Tezos_interop.Consensus.listen_operations node.Node.State.interop_context
     ~on_operation:(fun operation -> Flows.received_tezos_operation operation);
   Node.Server.start ~initial:node;
@@ -249,7 +251,13 @@ let node =
     let doc = "Path to the folder containing the node configuration data." in
     let open Arg in
     required & pos 0 (some string) None & info [] ~doc ~docv in
-
+  let minimum_block_delay =
+    let docv = "minimum_block_delay" in
+    let doc =
+      "Determines the minimum time the node will wait before propagating a \
+       newly produced block." in
+    let open Arg in
+    value & opt float 5. & info ["minimum_block_delay"] ~doc ~docv in
   let json_logs =
     let docv = "Json logs" in
     let doc = "This determines whether logs will be printed in json format." in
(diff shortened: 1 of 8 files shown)


src/node/config.ml
```diff
@@ -0,0 +1,24 @@
...
+  uri : Uri.t;
+}
+[@@deriving yojson]
+
+let make_identity ~secret ~key ~uri =
+  { secret; key; t = Key_hash.of_key key; uri }
+
+type t = {
+  identity : identity;
+  minimum_block_delay : float;
```
oteku, line 17: I think it would be better to have it in millisecondes & expressed with int

src/bin/deku_node.ml
```diff
@@ -240,7 +261,13 @@ let node =
     let doc = "Path to the folder containing the node configuration data." in
     let open Arg in
     required & pos 0 (some string) None & info [] ~doc ~docv in
-
+  let minimum_block_delay =
+    let docv = "minimum_block_delay" in
+    let doc =
+      "Determines the minimum time the node will wait before propagating a \
```
oteku, line 267: Would you add to the documentation the unit , format and minimum value.
In this PR
       "Determines the minimum time in secondes in floating number with a minimal value of 0. The node will wait before propagating a \"
       
       or if change to int 
              "Determines the minimum time in milli-secondes in int number with a minimal value of 0. The node will wait before propagating a \"

src/bin/deku_node.ml
```diff
@@ -138,6 +138,21 @@ let handle_receive_user_operation_gossip =
...
         request.user_operation)
 
+(* POST /user-operations-gossip *)
+(* Propagate a batch of user operations (core_user.t) over gossip network *)
+let handle_receive_user_operations_gossip =
+  handle_request
+    (module Network.User_operations_gossip)
+    (fun update_state request ->
+      let operations = request.user_operations in
+      Format.eprintf "packed: %d\n%!" (List.length operations);
```
aguillon, line 148: Please use `Log.debug`

d4hines: Oh, this is a debug message that should e removed. Good catch.

d4hines:
@EduardoRFS , I did what you suggested in the fixup commits.

src/bin/deku_node.ml
```diff
@@ -260,7 +262,13 @@ let node =
     let doc = "Path to the folder containing the node configuration data." in
     let open Arg in
     required & pos 0 (some string) None & info [] ~doc ~docv in
-
+  let minimum_block_delay =
+    let docv = "minimum_block_delay" in
```
aguillon, line 266:
```suggestion
    let docv = "minimum-block-delay" in
```

aguillon (approved):

src/bin/deku_node.ml
d4hines: resolved in the latest commit.

d4hines:
Our CI is broken, but we have fixes downstream of this change - merging without green CI.

PR #671, opened 2022-06-03
protocol: verify payload hash

EduardoRFS:
## Problem

Currently block.payload_hash is not verified when deserializing, which makes Deku susceptible to a couple attacks.

## Solution

Properly verify it, additionally I also removed the verify function that is not used.

### DIFF
--- a/src/protocol/block.ml
+++ b/src/protocol/block.ml
@@ -109,4 +109,3 @@ end)
 
 let sign ~key t = (sign ~key t).signature
 
-let verify ~signature t = verify ~signature t
--- a/src/protocol/block.ml
+++ b/src/protocol/block.ml
@@ -44,8 +44,10 @@ let hash, verify =
     let hash = BLAKE2B.hash (BLAKE2B.to_raw_string hash) in
     f (hash, block_payload_hash) in
   let hash = apply Fun.id in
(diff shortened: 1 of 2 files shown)


oteku (approved):
LGTM

Zett98 (approved):

aguillon (approved):
I'm ok with these changes, but they make it harder for the consensus or economic layers to witness that an actor proposed a bad block. For Tendermint this does not matter too much, as "proposing a bad block" and "not proposing a block" both amounts to the same decision, but still.

If we were to need it anyway, what would you do? Would you add an explicit step e.g. `Producer_failed`?

EduardoRFS:
@aguillon If someone sends a bullshit block it's the same as spamming and is likely due to a bug, as anyone malicious would never send a block with an invalid hash.

So I fail to see any possible consensus where punishing someone who just put a wrong hash would be a good idea.

The networking layer for sure is allowed to block the node(but this comes after pollinate), but I don't think it's a matter of consensus.

d4hines (approved):
LGTM. I talked with Edu and we had some concerns about finding the "right place" to do these verifications, and we agreed its at the networking level, but I opened an issue #677

PR #673, opened 2022-06-03
protocol: opaque user operations

EduardoRFS:
## Depends

- [x] #672

## Problem

Parsing blocks that contain a lot of user operations takes some considerable time, this leads to the block producer waiting the other validators to parse the user operations before they submit the signatures, making a bubble on our pipeline.

## Solution

As user_operations is not required to sign a block, we make it opaque and only parse it when actually applying a block

## Warning

This code is not great, it essentially suffers a bit from "prop drilling" but this is so that we avoid parsing the user operations more than once. But I will refactor this in the near future, so please bare with me.

### DIFF
--- a/src/protocol/block.ml
+++ b/src/protocol/block.ml
@@ -13,9 +13,29 @@ type t = {
   block_height : int64;
   consensus_operations : Protocol_operation.Consensus.t list;
   tezos_operations : Protocol_operation.Core_tezos.t list;
-  user_operations : Protocol_operation.Core_user.t list;
+  user_operations : string;
 }
 [@@deriving yojson]
+
+type user_operations = Protocol_operation.Core_user.t list [@@deriving yojson]
+
+let serialize_user_operations user_operations =
+  let json = user_operations_to_yojson user_operations in
+  Yojson.Safe.to_string json
+
+let parse_user_operations t =
+  let%ok json =
+    try Ok (Yojson.Safe.from_string t.user_operations) with
+    | _exn -> Error "invalid json" in
+  user_operations_of_yojson json
+
+let parse_user_operations t =
+  (* invalid encode means noop *)
+  match parse_user_operations t with
+  | Ok user_operations -> user_operations
+  | Error _ ->
+    (* TODO: maybe log this *)
+    []
 
 let hash, verify =
   let apply f ~state_root_hash ~withdrawal_handles_hash ~validators_hash
@@ -31,7 +51,7 @@ let hash, verify =
         * int64
         * Protocol_operation.Consensus.t list
         * Protocol_operation.Core_tezos.t list
-        * Protocol_operation.Core_user.t list] in
+        * string] in
     let json =
       to_yojson
         ( state_root_hash,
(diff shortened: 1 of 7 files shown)


src/protocol/block.mli
```diff
@@ -11,7 +11,7 @@ type t = private {
   block_height : int64;
   consensus_operations : Protocol_operation.Consensus.t list;
   tezos_operations : Protocol_operation.Core_tezos.t list;
-  user_operations : Protocol_operation.Core_user.t list;
+  user_operations : string;
```
aguillon, line 14: This seems to be a problem for https://github.com/marigold-dev/deku/pull/644

EduardoRFS:
> do you have an evidence of the performance bubble that you mentioned?

Yes, but I don't think we should be talking about evidences here, go one step back, what happens on each step of the consensus? Block -> Waits 2/3 -> Block, the Waits 2/3 is ALWAYS gonna stop the pipeline and as we do strictly in order block production the bubble MUST be there.

But then we should discuss the magnitude of it, it should be measurable as verifying signatures and hashes is one of the slowest things happening on our consensus, so a 10k user ops block will take some considerable amount of time on it.

> so it doesn't change anything regarding security?

Yes, that's the exact thought process.

> do blocks have to contain only serialized transactions

Nope and this is definitely not gonna hold for deku-p, a block may contain anything, is the will of the block producer to decide on it, because as we mentioned before he can always makes empty blocks if he wants.

aguillon:
More questions:
* suppose the block producer includes a user operation that is not properly signed or that uses an old nonce. As we are not checking the signatures anymore (in theory), this block is accepted by everybody else. Then, the incorrect transactions are treated as noops after the block is accepted; however, if the malicious node is also used as an indexer following Daniel's temporary PR #644, then he can give the proof that every other validators accepted the block, right?

So the moral here is that the indexer should really check for the correctness of the transactions itself and that #644 won't do, I guess.

EduardoRFS:
@aguillon 

> if the malicious node is also used as an indexer following Daniel's temporary PR https://github.com/marigold-dev/deku/pull/644, then he can give the proof that every other validators accepted the block, right?

Yes, he can, because they did, I'm not understanding your point. Any indexer should also check if the operations were applied or not, this is not special of Deku, an operation being added to a block doesn't say anything about the operation.

aguillon:
> @aguillon
>
> Yes, he can, because they did, I'm not understanding your point. Any indexer should also check if the operations were applied or not, this is not special of Deku, an operation being added to a block doesn't say anything about the operation.

I see. Are backtracked or failed operations on Tezos also an example of that?

src/protocol/block.ml
```diff
@@ -106,6 +105,28 @@ let of_yojson json =
 
 let compare a b = BLAKE2B.compare a.hash b.hash
 
+let make ~state_root_hash ~withdrawal_handles_hash ~validators_hash
```
d4hines, line 108: Why did you move `make`? I don't see any new dependency on the functions above it.

d4hines (approved):
This PR introduces some unpleasantness around how user operations are handled. Namely, we're operating at multiple levels of abstraction in `Protocol` now w.r.t to user operations in a way that seems unnecessary, all so the set of pending operations can be correctly maintained. I think the right thing to do would be to make operations equally opaque to the node/consensus layers, and include some sort of receipt that tells us when we can safely remove operations from our pending list. But as Edu argues, this is a bigger task, and meanwhile Deku benefits quite a bit in performance and Deku-p benefits by bringing Deku one step closer to Deku-p's model.

I approve, but no anime girl 4 u. You get muk.
but its a shiny muk

d4hines:
Tested locally on my machine. Skipping tilt ci.

PR #674, opened 2022-06-04
consensus: more standalone consensus

EduardoRFS:
## Depends

- [x] #673

## Problem

Currently our consensus is mixed in part with node and in part with protocol, additionally it's some really hackish code mixed with IO, ideally we should have a pure consensus library that can be easily instrumented with logs.

## Solution

This PR introduces(but doesn't integrate) a consensus library where the consensus is described in a discrete way where every step of the consensus is pure and has an intermediary variant.

**WARNING** this is not a complete refactor, it's more as an intermediary step to decouple a couple things and allow future refactors and testing.

## Review

Both commits are supposed to be noops.


src/consensus/steps.ml
```diff
@@ -0,0 +1,185 @@
...
+  else
+    `Past
+
+let is_signed_block_hash ~hash state =
+  match Block_pool.find_signatures ~hash state.block_pool with
+  | Some signatures -> Signatures.is_signed signatures
+  | None -> false
+
+let is_valid_block state block =
+  let is_all_operations_properly_signed _block = true in
```
aguillon, line 91: I don't think this makes sense now.

aguillon (approved):
Approved because I directly read #676 instead.

d4hines:
@EduardoRFS, do you think it makes sense to pull in the `network` library for the `snapshot` type? I have a similar problem in another PR, and I opted just to use tuples when stitching the two libraries together.

src/consensus/produce_block.ml
```diff
@@ -0,0 +1,71 @@
...
+        | Core_user user_operation ->
+          ( consensus_operations,
+            tezos_operations,
+            user_operation :: user_operations ))
+      ([], [], []) operations in
+  let block =
+    Block.produce ~state:state.protocol ~author:state.identity.t
+      ~next_state_root_hash ~consensus_operations ~tezos_operations
+      ~user_operations in
+  Both (Effect (Broadcast_block { block }), Check_block { block })
```
d4hines, line 71: Why must we check our own blocks?

src/consensus/effect.ml
```diff
@@ -0,0 +1,19 @@
...
+      signature : Signature.t;
+    }
+  | Applied_block           of {
+      block : Block.t;
+      receipts : (BLAKE2B.t * Core_deku.State.receipt) list;
+      trusted_validator_membership_change :
+        Trusted_validators_membership_change.Set.t;
+    }
+
+type t = effect
```
d4hines, line 19: I prefer to avoid type synonyms in general. Why this one here? If it's because of the `include Effect` in `Consensus`, I'd prefer to do something like:
```
type effect = Effect.t
include Effect
```

src/consensus/apply_block.ml
```diff
@@ -0,0 +1,75 @@
...
+            trusted_validator_membership_change
+        | Remove_validator validator ->
+          Trusted_validators_membership_change.Set.remove
+            { address = validator.address; action = Remove }
+            state.trusted_validator_membership_change)
+      state.trusted_validator_membership_change block.consensus_operations in
+
+  let state =
+    {
+      state with
```
Zett98, line 62: please no with

src/consensus/consensus.ml
```diff
@@ -0,0 +1,16 @@
+include State
+include Effect
```
Zett98, line 2: why not swap State and Effect that way you wont need to define type t = state

src/consensus/consensus.mli
```diff
@@ -0,0 +1,77 @@
...
+module Operation_map : Map.S with type key = Protocol.Operation.t
+
+type identity = {
+  secret : Secret.t;
+  key : Key.t;
+  t : Key_hash.t;
+  uri : Uri.t;
+}
+
+type state = private {
```
Zett98, line 13: why not type t?

src/consensus/consensus.mli
```diff
@@ -0,0 +1,77 @@
...
+}
+
+type t = state
+
+val make :
+  identity:identity ->
+  trusted_validator_membership_change:Trusted_validators_membership_change.Set.t ->
+  t
+
+type effect = private
```
Zett98, line 31: why not have a distinct module Effect with type t inside pubic part of the library?

src/consensus/produce_block.ml
Zett98: same question

src/consensus/state.ml
```diff
@@ -0,0 +1,61 @@
...
+module Operation_map = Map.Make (Operation)
+
+type identity = {
+  secret : Secret.t;
+  key : Key.t;
+  t : Key_hash.t;
+  uri : Uri.t;
+}
+
+type state = {
```
Zett98, line 12: why not type t?

src/consensus/steps.ml
```diff
@@ -0,0 +1,185 @@
...
+open Crypto
+open Protocol
+open State
+open Is_signable
+
+type signed =
+  | Signed
+  | Not_signed
+
+type step =
```
Zett98, line 11: why not type t?

src/protocol/block.ml
```diff
@@ -44,15 +71,48 @@ let hash, verify =
...
-    apply (fun (hash, _payload_hash) -> hash = expected_hash) in
+  let verify ~hash:expected_hash ~payload_hash:expected_payload_hash =
+    apply (fun (hash, payload_hash) ->
+        BLAKE2B.equal hash expected_hash
+        && BLAKE2B.equal payload_hash expected_payload_hash) in
   (hash, verify)
 
+let of_yojson json =
+  let%ok block = of_yojson json in
+  let%ok () =
```
Zett98, line 82: let%assert ?

Zett98 (commented):
Can we have a stable rule around using with in record updates on deku?

src/consensus/consensus.ml
```diff
@@ -0,0 +1,16 @@
...
+
+type t = state
+
+let with_block effect block state =
```
d4hines, line 6: I was thrown off by this variable name. I suggest `effect_handler` instead for this one, below, and `dispatch.ml`

src/consensus/dispatch.ml
```diff
@@ -0,0 +1,78 @@
+open Steps
+
+let rec dispatch effect step state =
```
d4hines, line 3: I think this function could be cleaned up by capturing effect in a closure and removing it from the function parameters (because we shouldn't be modifying the effect handler, right?). E.g.:
```ocaml
let dispatch effect step state = 
  let rec dispatch step state = 
  and check_block ~block state = ...
  and append_blcok ~block state = ...
  ...
```

d4hines: I also think an MLI would help. I know, it's just one function but still.

gsebil08 (approved):

src/consensus/apply_block.ml
```diff
@@ -0,0 +1,75 @@
...
+      protocol;
+      snapshots;
+      pending_operations;
+      trusted_validator_membership_change;
```
d4hines, line 66: It looks like we don't update the `recent_operation_receipts` here, which will break withdraw proofs. It's actually omitted from `state.ml`. Is this a mistake or am I missing something?

src/consensus/state.ml
EduardoRFS: because of the name

src/consensus/consensus.ml
EduardoRFS: good one, but please address on the latest PR as it's a nitpick

src/consensus/state.ml
EduardoRFS: so that open will not show `t -> Block.t -> t` everywhere

src/protocol/block.ml
EduardoRFS: this change is not from this PR

src/consensus/produce_block.ml
EduardoRFS: We don't, but there is no major advantage on not doing so. And it's better to be safe than sorry.

EduardoRFS: Also it follows the most external API

src/consensus/consensus.mli
EduardoRFS: https://github.com/marigold-dev/deku/pull/674#discussion_r893895978

src/consensus/steps.ml
EduardoRFS: https://github.com/marigold-dev/deku/pull/674#discussion_r893895978

src/consensus/dispatch.ml
EduardoRFS: Then you add another context and another level of indentation, the flat version is just simpler, you're "cleaning it up" and also making each one of the functions harder to analyse as they're now closures.

src/bin/deku_node.ml
```diff
@@ -1,6 +1,7 @@
 open Helpers
 open Cmdliner
 open Node
+open Consensus
```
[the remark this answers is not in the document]
EduardoRFS: nope

src/consensus/apply_block.ml
EduardoRFS: recent_operation_receipts is not consensus data

src/consensus/consensus.mli
EduardoRFS: To not add another level of indirection, effect is just a property of the consensus, there is no major advantage on adding a module for itself but there is disadvantages, that now to use or handle effects you need to open the Effect module.

d4hines (approved):
I reviewed this code in detail, but I have not tested it. There are a few names I want to change to make things more clear to readers, but besides that it's a huge improvement over what we have. I'm going to approve because the code doesn't run. In subsequent PR's, I'll do stress testing to make sure things still work.

Nice job
anime girl approves

PR #675, opened 2022-06-04
consensus: make the library usable

EduardoRFS:
## Depends

- [ ] #674

## Problem

Following #674 we have a quite nicer consensus library, but it doesn't include all operations required by flows.ml.

## Solution

This PR provides all the tools needed to use the consensus library on Deku.

Mostly handling of operations and validators change was missing. Additionally a lot of small hacks are needed on the current solution so I decided to add even if that makes the consensus library itself a bit worse.

### DIFF
--- a/src/bin/files.mli
+++ b/src/bin/files.mli
@@ -1,7 +1,6 @@
 open Consensus
 open Node
 open State
-open Consensus
 
 exception Invalid_json of string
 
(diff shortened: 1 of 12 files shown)


aguillon (approved):
Approved because I directly read #676 instead.

d4hines (approved):
lgtm. will test the next pr manually.
momo trying to grok sidechains

PR #676, opened 2022-06-04
node: use consensus library

EduardoRFS:
## Depends

- [ ] #676

## Problem

After #675 the consensus library is only lacking the integration by the node library.

## Solution

This PR integrates it on the node and removes most of the old code.


d4hines (approved):
LGTM. Deku's getting stronger guys
deku doing code kata's on project euler

Issue #641, opened 2022-05-26
deku-cli: poll for operation hashes


d4hines:
Users need a way to know when their operation is included in a block. I suggest for now we poll the block-by-level endpoint introduced in #536.

lykimq:
open PR for this issue at: #678

d4hines:
This is implemented in deku-toolkit now.

PR #678, opened 2022-06-07
WIP: user-operation-was-included-in-block RPC

d4hines:
## Problem

There's no way to know if an operation was included in a block. We want to be able to do this for accurate benchmarking.

## Solution

This is a bad solution: we just search all the applied blocks. Keeping the applied blocks is already bad - we should instead use an indexer, as discussed in #535. But this approach is "good enough" till we have a proper indexer.

## Related

- #535 
- #539
- open issue: #641


d4hines:
I marked this as draft until @lykimq has a chance to test this.

lykimq:
@d4hines I messed up with the rebase with main, and it took more time to fix git, so I decide to fork this branch and close it. Sorry about it.

PR #681, opened 2022-06-07
[d4hines] chore: refactor config out of node state

d4hines:
## Problem

There is more configuration data we want to be able to add to the node state beside identity - specifically, I'd like to make the block time configurable in the next PR. It makes sense to keep all this together in a module. 

## Solution

Refactor identity out into a `Config` module.

### Stacked PR Chain: d4hines
|#681|👉chore: refactor config out of node state||#658|

### DIFF
--- a/src/bin/deku_cli.ml
+++ b/src/bin/deku_cli.ml
@@ -630,9 +630,10 @@ let ensure_folder folder =
 let setup_identity node_folder uri =
   let%await () = ensure_folder node_folder in
   let identity =
-    let secret, key = Crypto.Ed25519.generate () in
-    let t = Key_hash.of_key (Ed25519 key) in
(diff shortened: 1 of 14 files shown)


InfiniteSwerve:
Test

d4hines:
> Why? Why not just add it to State.t?

@EduardoRFS  I think it makes sense to separate the data that should be present on start up and then never change from the data that is constantly changing as part of normal operation.

aguillon (approved):

d4hines:
> > It makes sense to keep all this together in a module.
> 
> Why? Why not just add it to State.t?

To keep things that can change at runtime separate from things that are set once at configuration time and then remain constant for the life of the node.

PR #682, opened 2022-06-07
[d4hines] feat: make minimum block time configurable

d4hines:
## Problem

Currently, we rate limit block production to 1 second. This is useful for normal manual testing, but doesn't make sense when you're trying to maximize throughput. We should make this configurable.

## Solution

Accept configuration for minimum block time at the command line

### Stacked PR Chain: d4hines
|#647|feat: block rate and transaction rate metrics  d4hines/block-rate-and-tps||-|
|#658|chore: refactor cli terms                      d4hines/batch-operations||#537|
|#668|make minimum block time configurable           d4hines/configurable-blocktime||#658|
|#682|👉feat: make minimum block time configurable||#681|
|#558|~~Simplify Sandbox.sh~~||-|
|#527|~~Dummy-ticket~~||-|
|#536|~~feat: add block-by-level rpc~~||#527|
|#537|~~Add block rate and operation rate metrics~~||#536|

### DIFF
--- a/src/bin/deku_cli.ml
+++ b/src/bin/deku_cli.ml
@@ -594,7 +594,10 @@ let info_produce_block =
 
 let produce_block node_folder =
   let%await identity = read_identity ~node_folder in
-  let%await state = Node_state.get_initial_state ~folder:node_folder in
+  (* TODO: this should be refactored out of the CLI per https://github.com/marigold-dev/deku/issues/659 *)
+  let%await state =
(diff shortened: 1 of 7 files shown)


EduardoRFS:
I agree with this being configurable, but I would suggest a better patch, a delay of 5s by default unless there is pending_operations, if there is pending_operations(that can be added) you make the block as soon as possible.

PR #686, opened 2022-06-10
Change Ticket table representation

Zett98:
## Problem

Right now we use a simple Map of {deku_key_hash;ticket_id}  -> Amount.t to store information about ticket ownership on deku. 
## Solution
In this pr i change the representation of the ledger,  to:
```ocaml 
type t = Amount.t Map.M(Ticket_id).t Map.M(Address).t
```
This will greatly simplify ticket  ownership transfers when you pass them to and from contracts during execution or origination
### E.g: 
   passing 5 tickets from implicit account into originated contract's storage: 
   
   we only touch the senders submap of tickets when transferring them to the contract/changing the amounts, and this is nice because we don't need to manipulate the whole table at once.
   
   When executing the contract with tickets in it's storage we can just use the submap of said contract from the ticket_table to pass them to ticket_transition table alongside tickets passed in argument position by user, again touching only the submap of user and submap of contract
   
Also i added the Ticket_transition_table to be used by smart-contracts when operating on tickets to ensure their linearity during execution.

### DIFF
--- a/src/core_deku/ticket_table.ml
+++ b/src/core_deku/ticket_table.ml
@@ -12,14 +12,14 @@ module Tickets = struct
   let singleton ~ticket ~amount =
     Ticket_map.of_seq (Seq.return (ticket, amount))
 
-  let merge t ~ticket ~amount =
-    let merger option ~amount =
+  let join t ~ticket ~amount =
+    let join option ~amount =
       let value = Option.fold ~none:amount ~some:Amount.(( + ) amount) option in
       value in
     Ticket_map.update ticket
       (fun x ->
-        let merged = merger ~amount x in
-        Option.some merged)
+        let joined = join ~amount x in
+        Option.some joined)
       t
 
   let get_and_remove t ~ticket ~to_take =
@@ -67,7 +67,7 @@ let balance t ~sender ~ticket =
 
 let update_or_create ~amount ~ticket t =
   Option.fold
-    ~some:(fun x -> Tickets.merge ~amount ~ticket x)
+    ~some:(fun x -> Tickets.join ~amount ~ticket x)
     ~none:(Tickets.singleton ~amount ~ticket)
     t
   |> Option.some
--- a/src/core_deku/ticket_table.ml
+++ b/src/core_deku/ticket_table.ml
@@ -1,0 +1,113 @@
+open Helpers
+
+module Tickets = struct
+  module Ticket_map = Map.Make_with_yojson (Ticket_id)
+
+  type t = Amount.t Ticket_map.t [@@deriving yojson]
+
+  let get_balance t ~ticket = Ticket_map.find_opt ticket t
+
+  let to_seq map = Ticket_map.to_seq map
+
+  let singleton ~ticket ~amount =
+    Ticket_map.of_seq (Seq.return (ticket, amount))
+
+  let merge t ~ticket ~amount =
+    let merger option ~amount =
+      let value = Option.fold ~none:amount ~some:Amount.(( + ) amount) option in
+      value in
+    Ticket_map.update ticket
+      (fun x ->
+        let merged = merger ~amount x in
+        Option.some merged)
+      t
+
+  let get_and_remove t ~ticket ~to_take =
+    match
+      Ticket_map.update ticket
+        (function
+          | Some amount ->
+            let () =
+              let compared = Amount.compare amount to_take in
+              if compared >= 0 then () else failwith "error" in
+            if Amount.equal amount to_take then
+              None
+            else
+              Some Amount.(amount - to_take)
+          | None -> failwith "error")
+        t
+    with
+    | _ as x -> Ok x
+    | exception Failure _ -> Error `Insufficient_funds
+
+  let get_and_remove_many t tickets =
+    let fetched =
+      List.fold_left_ok
+        (fun (lst, m) (ticket, amount) ->
+          let%ok m = get_and_remove m ~ticket ~to_take:amount in
+          let lst = Seq.cons (ticket, amount) lst in
+          Ok (lst, m))
+        (Seq.empty, t) tickets in
+    fetched
+
+  let empty = Ticket_map.empty
+
+  let is_empty = Ticket_map.is_empty
+end
+
+module Address_map = Map.Make_with_yojson (Address)
+
+type t = Tickets.t Address_map.t [@@deriving yojson]
+
+let empty = Address_map.empty
+
+let balance t ~sender ~ticket =
+  let%some map = Address_map.find_opt sender t in
+  Tickets.get_balance ~ticket map
+
+let update_or_create ~amount ~ticket t =
+  Option.fold
+    ~some:(fun x -> Tickets.merge ~amount ~ticket x)
+    ~none:(Tickets.singleton ~amount ~ticket)
+    t
+  |> Option.some
+
+let deposit t ~destination ~ticket ~amount =
+  Address_map.update destination (update_or_create ~amount ~ticket) t
+
+let update t ~sender ~submap =
+  if Tickets.is_empty submap then
+    Address_map.remove sender t
+  else
+    Address_map.add sender submap t
+
+let transfer t ~sender ~destination ~ticket ~amount =
+  let%ok map =
+    Address_map.find_opt sender t |> Option.to_result ~none:`Insufficient_funds
+  in
+  let%ok map = Tickets.get_and_remove ~ticket ~to_take:amount map in
+  let t = update t ~sender ~submap:map in
+  let t = Address_map.update destination (update_or_create ~amount ~ticket) t in
+  Ok t
+
+let withdraw t ~sender ~ticket ~amount =
+  let%ok map =
+    Address_map.find_opt sender t |> Option.to_result ~none:`Insufficient_funds
+  in
+  let%ok map = Tickets.get_and_remove ~ticket ~to_take:amount map in
+  let t = update t ~sender ~submap:map in
+  Ok t
+
+let take_tickets t ~sender ~tickets =
+  let%ok map =
+    Address_map.find_opt sender t |> Option.to_result ~none:`Insufficient_funds
+  in
+  let%ok tickets, map = Tickets.get_and_remove_many map tickets in
+  let t = update t ~sender ~submap:map in
+  Ok (tickets, t)
+
+let take_all_tickets t ~sender =
+  let map =
+    Address_map.find_opt sender t |> Option.value ~default:Tickets.empty in
+  let tickets = Tickets.to_seq map in
+  (tickets, Address_map.remove sender t)
--- a/src/core_deku/ticket_table.mli
+++ b/src/core_deku/ticket_table.mli
@@ -1,0 +1,35 @@
+type t [@@deriving yojson]
+
+val empty : t
+
+val balance : t -> sender:Address.t -> ticket:Ticket_id.t -> Amount.t option
+
+val deposit :
+  t -> destination:Address.t -> ticket:Ticket_id.t -> amount:Amount.t -> t
+
+val transfer :
+  t ->
+  sender:Address.t ->
+  destination:Address.t ->
+  ticket:Ticket_id.t ->
+  amount:Amount.t ->
+  (t, [> `Insufficient_funds]) result
+
+val withdraw :
+  t ->
+  sender:Address.t ->
+  ticket:Ticket_id.t ->
+  amount:Amount.t ->
+  (t, [> `Insufficient_funds]) result
+
+val take_tickets :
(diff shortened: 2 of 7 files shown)


src/core_deku/ticket_table.ml
```diff
@@ -0,0 +1,106 @@
...
+    fetched
+
+  let fold = Map.fold
+
+  let empty = Map.empty
+end
+
+module Map = Map.Make_with_yojson (Address)
+
+type t = Tickets.t Map.t [@@deriving yojson]
```
renatoalencar, line 62: I think `Tickets` could have a better name, a more descriptive one. But I don't have nothing in mind rn, so not gonna block it.

renatoalencar (approved):
LGTM

d4hines:
@Zett98 I don't fully understand the point of this PR and I don't understand the design choices (e.g. `Ledger.update_table` seems to leak an abstraction here). Let's have a call to walk through it together.

src/core_deku/ticket_table.ml
```diff
@@ -0,0 +1,118 @@
...
+
+  let subtract t ~ticket ~to_withdraw =
+    let%ok _, map = get_and_remove t ~ticket ~to_take:to_withdraw in
+    Ok map
+
+  let fetch t tickets =
+    let fetched =
+      List.fold_left_ok
+        (fun (lst, m) (ticket, amount) ->
+          let%ok amount_available, m =
```
renatoalencar, line 47: I couldn't understand if that's intentional, but `amount_available` is the same as `amount`.

Zett98: removed

src/core_deku/ticket_table.ml
```diff
@@ -0,0 +1,118 @@
...
+    fetched
+
+  let fold = Map.fold
+
+  let empty = Map.empty
+
+  let is_empty = Map.is_empty
+end
+
+module Map = Map.Make_with_yojson (Address)
```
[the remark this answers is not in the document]
Zett98: renamed

src/core_deku/ticket_table.ml
Zett98: renamed

src/core_deku/ticket_transition_table.ml
```diff
@@ -0,0 +1,144 @@
+open Helpers
```
d4hines, line 1: This code seems dead. Remove it for now?

src/core_deku/ticket_table.ml
```diff
@@ -0,0 +1,126 @@
+open Helpers
+
+module Tickets = struct
+  module Ticket_map = Map.Make_with_yojson (Ticket_id)
+
+  type t = Amount.t Ticket_map.t [@@deriving yojson]
+
+  let get_balance t ~ticket = Ticket_map.find_opt ticket t
+
+  let of_seq = Ticket_map.of_seq
```
d4hines, line 10: You don't use this alias.

src/core_deku/ticket_table.ml
```diff
@@ -0,0 +1,126 @@
...
+
+  let singleton ~ticket ~amount =
+    Ticket_map.of_seq (Seq.return (ticket, amount))
+
+  let merge t ~ticket ~amount =
```
d4hines, line 15: Isn't this called "fuse"?

src/core_deku/ticket_table.ml
```diff
@@ -0,0 +1,126 @@
...
+      value in
+    Ticket_map.update ticket
+      (fun x ->
+        let merged = merger ~amount x in
+        Option.some merged)
+      t
+
+  let get_and_remove t ~ticket ~to_take =
+    let%ok amount =
+      Ticket_map.find_opt ticket t |> Option.to_result ~none:`Insufficient_funds
```
d4hines, line 27: Not sure how much this matters but you could move this into `Ticket_map.update` with an exception to catch the `Insufficient_funds` case and turn into a result. Then you're only doing one lookup.

d4hines: If it were anywhere else in Deku I wouldn't even mention it, but you may care about that sort of thing for the VM, not sure.

src/core_deku/ticket_table.ml
```diff
@@ -0,0 +1,126 @@
...
+          | Some amount ->
+            if Amount.equal amount to_take then
+              None
+            else
+              Some Amount.(amount - to_take)
+          | None -> assert false)
+        t in
+    Ok map
+
+  let subtract t ~ticket ~to_withdraw =
```
d4hines, line 45: This seems like an unnecessary aliasing of `get_and_remove`

src/core_deku/ticket_table.ml
```diff
@@ -0,0 +1,126 @@
...
+    let%ok map = get_and_remove t ~ticket ~to_take:to_withdraw in
+    Ok map
+
+  let fetch t tickets =
```
d4hines, line 49: Let's try to have more consistent names. We could rename this to `get_and_remove_many`.

src/core_deku/ticket_table.ml
```diff
@@ -0,0 +1,126 @@
...
+    let fetched =
+      List.fold_left_ok
+        (fun (lst, m) (ticket, amount) ->
+          let%ok m = get_and_remove m ~ticket ~to_take:amount in
+          let lst = Seq.cons (ticket, amount) lst in
+          Ok (lst, m))
+        (Seq.empty, t) tickets in
+    fetched
+
+  let fold = Ticket_map.fold
```
d4hines, line 59: This breaks the invariant we want for this table. 

Instead I suggest we move the `tickets` function from the outer module in here. Then the table module enforces all our invariants, and the outer module simply adds higher level functions.

src/core_deku/ticket_table.ml
```diff
@@ -0,0 +1,126 @@
...
+  let%ok map =
+    Table.find_opt sender t |> Option.to_result ~none:`Insufficient_funds in
+  let%ok tickets, map = Tickets.fetch map tickets in
+  let t = remove_if_empty t ~sender ~submap:map in
+  Ok (tickets, t)
+
+let tickets t ~sender =
+  let map = Table.find_opt sender t |> Option.value ~default:Tickets.empty in
+  let tickets =
+    Tickets.fold
```
d4hines, line 123: See comment above about moving this into the inner `Table` module.

src/core_deku/ticket_table.ml
```diff
@@ -0,0 +1,126 @@
...
+    fetched
+
+  let fold = Ticket_map.fold
+
+  let empty = Ticket_map.empty
+
+  let is_empty = Ticket_map.is_empty
+end
+
+module Table = Map.Make_with_yojson (Address)
```
d4hines, line 66: I think it would be better to stick with the convention used everywhere else in Deku:
```suggestion
module Address_map = Map.Make_with_yojson (Address)
```

src/core_deku/ticket_table.ml
```diff
@@ -0,0 +1,126 @@
...
+  Option.fold
+    ~some:(fun x -> Tickets.merge ~amount ~ticket x)
+    ~none:(Tickets.singleton ~amount ~ticket)
+    t
+  |> Option.some
+
+let deposit t ~destination ~ticket ~amount =
+  Table.update destination (update_or_create ~amount ~ticket) t
+
+let remove_if_empty t ~sender ~submap =
```
d4hines, line 92:
```suggestion
let update t ~sender ~submap =
```
Won't block on this though.

src/core_deku/ticket_handle.mli
```diff
@@ -0,0 +1,15 @@
+type t [@@deriving yojson, ord, eq]
```
d4hines, line 1: This file is is dead, along with `Ticket_transition_table`

src/core_deku/ticket_table.ml
```diff
@@ -0,0 +1,126 @@
...
+  Ok t
+
+let take_tickets t ~sender ~tickets =
+  let%ok map =
+    Table.find_opt sender t |> Option.to_result ~none:`Insufficient_funds in
+  let%ok tickets, map = Tickets.fetch map tickets in
+  let t = remove_if_empty t ~sender ~submap:map in
+  Ok (tickets, t)
+
+let tickets t ~sender =
```
d4hines, line 120: This should be called `take_all_ticket` for consistency with `take_tickets`.

src/core_deku/ticket_table.mli
```diff
@@ -0,0 +1,34 @@
...
+  (t, [> `Insufficient_funds]) result
+
+val withdraw :
+  t ->
+  sender:Address.t ->
+  ticket:Ticket_id.t ->
+  amount:Amount.t ->
+  (t, [> `Insufficient_funds]) result
+
+val tickets : t -> sender:Address.t -> (Ticket_id.t * Amount.t) Seq.t * t
```
d4hines, line 25: nitpick that I won't block on: would be nice if mli function order is in same order as implementation.

src/core_deku/ticket_table.ml
```diff
@@ -0,0 +1,126 @@
...
+type t = Tickets.t Table.t [@@deriving yojson]
+
+let empty = Table.empty
+
+let balance t ~sender ~ticket =
+  let%some map = Table.find_opt sender t in
+  Tickets.get_balance ~ticket map
+
+let update_tickets t ~sender ~tickets =
+  if Seq.is_empty tickets then
```
d4hines, line 77: This function seems weird to me. 

Either you're depositing tickets, in which case just use `deposit` directly.

Or you're just garbage collecting `Table.remove`. If that's necessary, why not do it automatically when you withdraw all the tickets? If it's not necessary, why do it at all?

d4hines (changes requested):
It's much better but I still see some weird things in the abstractions provided by `Ticket_table` and its inner `Table` module (which i think should be renamed). Also we should delete some dead code - should be added as part of a separate PR. Atomic commits make reviewing the changes much easier since you can focus on one change at a time.

src/core_deku/ticket_table.ml
Zett98: join

src/core_deku/ticket_table.ml
Zett98: changed

src/core_deku/ticket_table.ml
Zett98: removed

src/core_deku/ticket_table.ml
Zett98: changed

src/core_deku/ticket_table.ml
Zett98: changed

src/core_deku/ticket_handle.mli
Zett98: removed

src/core_deku/ticket_table.ml
Zett98: when you transfer all of your tickets from one deku account to another one, reference to the old one inside the ticket table should be deleted, to not keep old empty entries in the map

d4hines: I don't get it - isn't this addressed by my second bullet point?

But I see the code is removed. Resolving.

src/core_deku/ticket_table.ml
d4hines: Ok, so can we rename to `join` then?

src/core_deku/ticket_table.ml
```diff
@@ -0,0 +1,113 @@
...
+
+module Tickets = struct
+  module Ticket_map = Map.Make_with_yojson (Ticket_id)
+
+  type t = Amount.t Ticket_map.t [@@deriving yojson]
+
+  let get_balance t ~ticket = Ticket_map.find_opt ticket t
+
+  let to_seq map = Ticket_map.to_seq map
+
```
d4hines, line 11:
```suggestion
```
This is dead code that isn't used. It's also not really useful as an alias in this case.

src/core_deku/ticket_table.mli
```diff
@@ -0,0 +1,35 @@
...
+  t ->
+  sender:Address.t ->
+  tickets:(Ticket_id.t * Amount.t) list ->
+  ((Ticket_id.t * Amount.t) Seq.t * t, [> `Insufficient_funds]) result
+
+val take_all_tickets :
+  t -> sender:Address.t -> (Ticket_id.t * Amount.t) Seq.t * t
+
+(* val update_tickets :
+   t -> sender:Address.t -> tickets:(Ticket_id.t * Amount.t) Seq.t -> t *)
```
d4hines, line 35:
```suggestion
```
Dead code.

PR #687, opened 2022-06-10
Remove Esy as a supported package manager

d4hines:
## Problem

2 package managers are hard to support simultaneously.

## Solution

Nix for everything.


.ocamlformat
```diff
@@ -1,5 +1,6 @@
-# In both Esy and Nix we're using a patched version of 0.20.1
-# but ocamlformat's version check won't work for both of them simultaneously.
+# We're using a patched version of 0.20.1 but
+# ocamlformat's version check won't work for both of them simultaneously.
```
ulrikstrid, line 2: Maybe change this to something about nix not having a correct version?

ulrikstrid (approved):
😍

.ocamlformat
d4hines: Addressed in https://github.com/marigold-dev/deku/pull/687/commits/480521342e6cc267833d27a4b0cee8db4809fefb

PR #691, opened 2022-06-13
[d4hines] fix: fix prometheus with tilt

d4hines:
Fixes Prometheus in docker-compose by removing
the 'ports' setting, which is incompatible with
the host network mode.

Also removes fixes the grouping and removes the
'autoinit=false' config our Tiltfile for several
services. These services are integral to Deku
development, so they should start every time.

### Stacked PR Chain: d4hines
|#691|👉fix: fix prometheus with tilt||-|

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -27,16 +27,15 @@ deku_yaml = make_deku_yaml(no_of_deku_nodes)
 # Run docker-compose
 docker_compose(["./docker-compose.yml", deku_yaml])
 
-dc_resource("db", labels=["database"], trigger_mode=TRIGGER_MODE_MANUAL, auto_init=False)
-dc_resource("elastic", labels=["database"], trigger_mode=TRIGGER_MODE_MANUAL, auto_init=False)
-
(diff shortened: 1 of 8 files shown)


aguillon (approved):
I totally understand and approve those changes

Tiltfile
```diff
@@ -27,16 +27,15 @@ deku_yaml = make_deku_yaml(no_of_deku_nodes)
 # Run docker-compose
 docker_compose(["./docker-compose.yml", deku_yaml])
 
-dc_resource("db", labels=["database"], trigger_mode=TRIGGER_MODE_MANUAL, auto_init=False)
-dc_resource("elastic", labels=["database"], trigger_mode=TRIGGER_MODE_MANUAL, auto_init=False)
-
+dc_resource("db", labels=["tezos"])
```
ulrikstrid, line 30: These changes makes `tilt ci` needlessly slow and noisy and there's not a good way to disable them. How important is it to always run these resources?
@d4hines

PR #694, opened 2022-06-15
fix: re-use validators and identity from state

d4hines:
## Problem 
When producing a block, we redundantly fetch the validators again after just
having fetched it to build the node state. This is very slow and unnecessary.

## Solution

Use the validators and identity that are part of the State.t that we build to produce the block.


d4hines:
Probably going to close this PR with a better solution.

d4hines:
This is a small improvement but it's an improvement - let's merge it after all.

aguillon (approved):

PR #697, opened 2022-06-17
Update to jakartanet

ulrikstrid:
## Problem

We want to test deku against jakartanet

## Solution

Update  flextesa and run `jakartabox`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -109,8 +109,8 @@ services:
   flextesa:
     container_name: deku_flextesa
     restart: always
-    image: oxheadalpha/flextesa:20220127
-    command: hangzbox start
+    image: oxheadalpha/flextesa:20220510
+    command: jakartabox start
     environment:
       - block_time=4
     ports:
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,9 +9,9 @@
       "version": "0.0.0",
       "license": "MIT",
       "dependencies": {
-        "@taquito/rpc": "^13.0.0",
-        "@taquito/signer": "^13.0.0",
-        "@taquito/taquito": "^13.0.0",
+        "@taquito/rpc": "^11.2.0",
+        "@taquito/signer": "^11.2.0",
+        "@taquito/taquito": "^11.2.0",
         "webpack": "^5.47.1",
         "webpack-cli": "^4.7.1"
       },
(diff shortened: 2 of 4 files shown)


docker-compose.yml
```diff
@@ -109,10 +109,10 @@ services:
   flextesa:
     container_name: deku_flextesa
     restart: always
-    image: oxheadalpha/flextesa:20220127
-    command: hangzbox start
+    image: oxheadalpha/flextesa:20220510
+    command: jakartabox start
     environment:
-      - block_time=4
+      - block_time=30
```
ulrikstrid, line 115: This slows down the local testing significantly, especially if we pair it with a more realistic `--unsafe_tezos_required_confirmations` in `sandbox.sh`, might be good for the devs to test with this, especially when  doing realistic benchmarks.

ulrikstrid: Moved this change and added the sandbox.sh into a separate commit that we can drop before merging this, just want to showcase a more realistic scenario

aguillon (changes requested):
Could you make it a test case for the CI instead of the default, please? There's no way we wait 1min to run a `./sandbox.sh start`, I'm sorry. :)

d4hines:
I think with some upcoming work on deku-cli we can have our cake and eat it to with more realistic confirmation times and a bootstrapping process that is still fast in development.

ulrikstrid:
Dropped the commit that changed blocktime and sandbox.sh and rebased, this should be good to go now @aguillon

aguillon (approved):
Thank you!

oteku:
@ulrikstrid  Branch doesn't build on macOS M1
```
[1/0/1 copied (2.8/111.2 MiB), 0.7/19.5 MiB DL] fetching source from https://cache.nixos.orgdirenv: ([/opt/homebrew/bin/direnv export zsh]) is taking a while to execute. Use CTRL-C to give up.
error: Package ‘python3.9-pyopenssl-22.0.0’ in /nix/store/a2bf460i24mcns7pfvkw65xm8f30p6rg-nixpkgs-patched/pkgs/development/python-modules/pyopenssl/default.nix:73 is marked as broken, refusing to evaluate.

       a) To temporarily allow broken packages, you can use an environment variable
          for a single invocation of the nix tools.

            $ export NIXPKGS_ALLOW_BROKEN=1

        Note: For `nix shell`, `nix build`, `nix develop` or any other Nix 2.4+
        (Flake) command, `--impure` must be passed in order to read this
        environment variable.

       b) For `nixos-rebuild` you can set
         { nixpkgs.config.allowBroken = true; }
       in configuration.nix to override this.

       c) For `nix-env`, `nix-build`, `nix-shell` or any other Nix command you can add
         { allowBroken = true; }
       to ~/.config/nixpkgs/config.nix.
(use '--show-trace' to show detailed location information)
direnv: export +GHSTACK_TARGET_REPOSITORY
```

ulrikstrid:
@oteku is your git repository clean? Does main build?
Since we're not touching anything other than node dependencies this shouldn't affect a python package

aguillon (approved):

d4hines (approved):
LGTM :+1: 

@oteku Deku is already giving this error in main, so we'll merge this one and fix that as a separate pr.

PR #700, opened 2022-06-20
add contract vm execution context

Zett98:
## Depends
- [x] #686

## Problem

We need to have a proper smart contract context module which we will use for conversions to/from core_deku types, interactions with tickets and other external functions provided to the vm's like e.g hashing
## Solution

Provide the layer


src/core_deku/contracts.ml
```diff
@@ -0,0 +1,40 @@
...
+
+    let of_bytes t =
+      let addr = Address.of_string (Bytes.to_string t) in
+      Option.value_exn addr
+  end
+
+  module Ticket_id = struct
+    include Ticket_id
+
+    let size t = 36 + Bytes.length t.data
```
renatoalencar, line 19: I guess using `Address.size` is better than using the hardcoded value.

src/core_deku/contracts.ml
```diff
@@ -0,0 +1,40 @@
...
+      let byt = Tezos.Address.to_string t.ticketer in
+      let byt = Bytes.of_string byt in
+      let byt2 = t.data in
+      Stdlib.Bytes.cat byt byt2
+  end
+
+  module Amount = struct
+    include Amount
+
+    let size = 36
```
renatoalencar, line 31: Why is size 36 for amount?

src/smart_contracts/ticket_handle.ml
```diff
@@ -0,0 +1,46 @@
...
+
+module Make (CC : Conversions.S) = struct
+  include CC
+  include BLAKE2B_20
+
+  let make owner ticket amount =
+    BLAKE2B_20.hash_v
+      [
+        Address.to_string owner;
+        Tezos.Ticket_id.to_string ticket;
```
renatoalencar, line 31: I think that's better using `Core_deku.Ticket_id` instead of Tezos' directly.

renatoalencar (changes requested):
Looks great, apart from the comments I've left, I think there's a few duplicated interfaces and some make changing some types difficult. Could you simplify that a little, please?

src/core_deku/contracts.ml
Zett98: typo

src/core_deku/contracts.ml
Zett98: changed

src/smart_contracts/ticket_handle.ml
Zett98: Core_deku.Ticket_id is just an alias to Tezos.Ticket_id.t that doesn't expose the to_string method. It will be changed in ticket_minting pr when core_deku ticket becomes more self sufficient and not barely an alias

renatoalencar (approved):
LGTM

PR #701, opened 2022-06-21
chore(ci): Speed up CI by consolidating builds

ulrikstrid:
## Problem

We currently pull the cache multiple times which is kind of slow

## Solution

Since our e2e tests needs the same thing as our standard (native) build we can run them just after.
Docker needs our static builds so we can do the same here.

## Depends

#701 should be merged first since it fixes `tilt ci`


d4hines (approved):
LGTM :+1:

PR #702, opened 2022-06-21
Fix TLS in docker

ulrikstrid:
## Problem

Our docker containers can't talk to other nodes that are using https

## Solution

Add `cacert` to the correct place.
This PR also moves our dependencies to `buildInputs` and just keeps `cacert` in `propagatedBuildInputs` which makes our closure slightly smaller. Let me know if we want to move this to a separate PR.

### DIFF
--- a/nix/deku.nix
+++ b/nix/deku.nix
@@ -37,7 +37,7 @@ in ocamlPackages.buildDunePackage rec {
   nativeBuildInputs = [ nodejs removeReferencesTo ] ++ npmPackages
     ++ (with ocamlPackages; [ utop reason ]);
 
-  propagatedBuildInputs = with ocamlPackages;
+  buildInputs = with ocamlPackages;
     [
       cmdliner
       ppx_deriving
@@ -68,9 +68,9 @@ in ocamlPackages.buildDunePackage rec {
     ]
     # checkInputs are here because when cross compiling dune needs test dependencies
     # but they are not available for the build phase. The issue can be seen by adding strictDeps = true;.
-    ++ checkInputs
-    # some benchmarking libraries are broken om m1, so we make them optional
-    ++ (if system != "aarch64-darwin" then [ landmarks ] else [ ]);
+    ++ checkInputs;
+
+  propagatedBuildInputs = [ cacert ];
 
   checkInputs = with ocamlPackages; [ alcotest qcheck qcheck-alcotest rely ];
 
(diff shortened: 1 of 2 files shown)


nix/deku.nix
```diff
@@ -68,28 +68,28 @@ in ocamlPackages.buildDunePackage rec {
     ]
     # checkInputs are here because when cross compiling dune needs test dependencies
     # but they are not available for the build phase. The issue can be seen by adding strictDeps = true;.
-    ++ checkInputs
-    # some benchmarking libraries are broken om m1, so we make them optional
```
ulrikstrid, line 72: This should have been fixed in upstream so this should be OK to remove

aguillon: Can't test this as I don't have aarch64-darwin.

aguillon (changes requested):
I don't really understand the nix changes so I can't comment on that.

Regarding HTTPS: I'm not sure to understand why someone would want to use HTTPS with Deku. On the other hand, if we do want to support either HTTP or HTTPS, then we need to be able to test it. @ulrikstrid would it be easy?

ulrikstrid:
@aguillon we're using https in the testnet running in kubernetes. I think using https is something we want to support by default? At least for making requests.
Note that we're not running the deku node with https but we have a reverse proxy (actually a ingress) that terminates the tls before sending the request to the node.

nix/deku.nix
```diff
@@ -37,7 +37,7 @@ in ocamlPackages.buildDunePackage rec {
   nativeBuildInputs = [ nodejs removeReferencesTo ] ++ npmPackages
     ++ (with ocamlPackages; [ utop reason ]);
 
-  propagatedBuildInputs = with ocamlPackages;
+  buildInputs = with ocamlPackages;
```
ulrikstrid, line 40: These packages doesn't need to be "propagated" to consumers so we can just have them as `buildInputs` this makes the closure much smaller and we don't have to use weird tricks to make it small.

nix/deku.nix
```diff
@@ -68,28 +68,28 @@ in ocamlPackages.buildDunePackage rec {
...
+  propagatedBuildInputs = [ cacert ];
 
   checkInputs = with ocamlPackages; [ alcotest qcheck qcheck-alcotest rely ];
 
   # Remove every directory which could have links to other store paths.
   # This makes the result much smaller
   isLibrary = false;
   postFixup = ''
-    rm -rf $out/lib $out/nix-support $out/share/doc
+    rm -rf $out/lib $out/share/doc
```
ulrikstrid, line 81: `$out/nix-support` can be removed since we moved things to `buildInputs`

nix/deku.nix
```diff
@@ -68,28 +68,28 @@ in ocamlPackages.buildDunePackage rec {
...
     remove-references-to \
       -t ${ocamlPackages.ocaml} \
-      $out/bin/deku-{node,cli}
+      $out/bin/{asserter,deku-node,deku-cli}
```
ulrikstrid, line 84: We want to remove the references from all our binaries

nix/docker.nix
```diff
@@ -23,6 +23,10 @@ in pkgs.dockerTools.buildImage {
     architecture = "amd64";
     os = "linux";
 
+    Env = [
```
ulrikstrid, line 26: Adds the SSL certificates from the ca-bundle to the environment so that our binaries can pick them up for http requests

ulrikstrid:
@aguillon Added comments in github that describes why the changes are made

d4hines (approved):
LGTM :+1: 

@aguillon, I think it's sufficient to let infra ping us if we break HTTPS. They definitely will notice, as they're going to be deploying test-nets regularly.

aguillon:
Works for me, thanks!

aguillon (approved):

PR #703, opened 2022-06-21
Tilt fixes to make `tilt ci` better

ulrikstrid:
## Problem

We are usually running tilt in local mode while docker is the default
We are running a bunch of resources that we don't need when just running the CI

## Solution

Flip the mode configuration to have `local` as default
Add a flag to start all services which defaults to `false`, use it as `tilt up -- --all-services` to start indexer etc


d4hines (approved):
LGTM!

PR #704, opened 2022-06-21
Allow setting PORT environment

ulrikstrid:
## Problem

When deploying a network we might not always want the advertised address to map 1-1 to what we deploy

## Solution

Allow setting `PORT` as a environment variable and use that first.
Also makes the code safer by defaulting to 4440 instead of failing if the address doesn't include a port


lykimq (approved):

d4hines (approved):
LGTM

Issue #99, opened 2021-07-31
generate OCaml module o call Tezos from Ligo contract


EduardoRFS:

EduardoRFS:
I think this is related to #86

PR #705, opened 2022-06-21
D4hines/node-with-consensus

d4hines:
## Depends

- [ ] #100

## Related

- #99


d4hines:
Closing since we merged #676

PR #706, opened 2022-06-21
Consolidate formatting

ulrikstrid:
## Problem

We want to make sure we have consistent formatting

## Solution

[treefmt](https://github.com/numtide/treefmt) is a tool that allows us to run multiple formatters with just 1 command, it has a cache so that it only runs on changed files.
Also adds `nix fmt` to run the formatter(s).


lykimq (approved):

treefmt.toml
```diff
@@ -0,0 +1,16 @@
...
+options = ["build", "@fmt", "--auto-promote"]
+includes = ["*.ml", "dune"]
+
+[formatter.javascript]
+command = "prettier"
+includes = ["*.js"]
+
+[formatter.sh]
+command = "shellcheck"
+includes = ["*.sh"]
```
d4hines, line 16:
```suggestion

[formatter.sh]
command = "shellcheck"
includes = ["*.sh"]
```
I don't think shellcheck does formatting

d4hines (approved):
This is great! :tada: 

Approved, but look at the suggestion.

PR #707, opened 2022-06-21
Add wasm-vm to deku

Zett98:
## Depends

- [x] #686
- [x] #700
## Problem

We need to add Webassembly vm to Deku
## Solution

Add wasm-vm as a standalone library based on Webassembly spec interpreter written in ocaml([link](https://github.com/WebAssembly/spec)), wire it into deku, and validate it using e2e tests


nix/packages/wasm.nix
```diff
@@ -0,0 +1,27 @@
...
+
+    strictDeps = true;
+
+    # x86_64-unknown-linux-musl-ld: -r and -pie may not be used together
+    hardeningDisable = lib.optional stdenv.hostPlatform.isStatic "pie";
+
+    meta = {
+      description = "An executable and OCaml library to run, read and write Web Assembly (wasm) files and manipulate their AST";
+      license = lib.licenses.asl20;
+      maintainers = [ lib.maintainers.vbgl ];
```
d4hines, line 23: Did you find this in nixpkgs? If so let's do an override. If not, why is vgbl listed as the maintainer?

Zett98: Ulrik helped me do it  there was some issues with ocaml versions in nix repository version of wasm. I tried doing the override but it didn't work out

wallet.json
```diff
@@ -0,0 +1 @@
+{"address": "tz1RPNjHPWuM8ryS5LDttkHdM321t85dSqaf", "priv_key": "edsk36FhrZwFVKpkdmouNmcwkAJ9XgSnE5TFHA7MqnmZ93iczDhQLK"}
```
renatoalencar, line 1: Should this be here?

renatoalencar (approved):
This looks good, but I'll leave two small comments/suggestions. The one about validating the contract before sending it to the node just to make our lives easier.

PR #719, opened 2022-06-22
fix: run treefmt from shell env

d4hines:
## Problem

treefmt fails in CI

## Solution

It's failing because it can't find ocamlformat, etc. So we run it with `nix develop` to ensure all required deps are there.


ulrikstrid (approved):
Looks good :ok_hand:

ulrikstrid:
I'm not sure why it formats the nix files, aren't we  using `nixfmt` already?

aguillon (approved):
> I'm not sure why it formats the nix files, aren't we using `nixfmt` already?

Isn't the goal of treefmt to call the relevant formatters?

ulrikstrid:
Yes, but the files should have been formatted already :thinking:

ulrikstrid:
@d4hines I fixed this by creating a custom formatter that just wraps all the formatters we need with treefmt :sweat_smile: we can revert that  particular change, but using git to do the diff seems more robust for CI in either case.

PR #725, opened 2022-06-24
Drop unused vms from Deku

Zett98:
## Depends
- [x] #700 
- [x] #707

## Problem

Lambda vm and Dummy vm were proof of concept and wont be used again
## Solution

Remove them


renatoalencar (approved):
LGTM

PR #726, opened 2022-06-24
Wasm-vm fix ticket_handle indexes on reloads

Zett98:
## Depends

- [x] #700
- [x] #707
- [x] #724 

## Problem

We don't update ticket handles in serialized storage when saving new contract state, and that leads to invalid contract state on next execution
## Solution

Fix it

### DIFF
--- a/sandbox.sh
+++ b/sandbox.sh
@@ -389,7 +389,7 @@ test_wasm() {
   ARG=$(ticket_handle)
   echo "param: \"$ARG\""
   sleep 5
-  deku-cli create-transaction ./data/0 wallet.json "$CONTRACT_ADDRESS" 100 "Pair \"$DUMMY_TICKET\" 0x" --arg="$ARG" --vm_flavor="Wasm" --tickets="Pair \"$DUMMY_TICKET\" 0x":100
+  deku-cli create-transaction ./data/0 wallet.json "$CONTRACT_ADDRESS" 100 "Pair \"$DUMMY_TICKET\" 0x" --arg="$ARG" --vm_flavor="Wasm" --tickets="Pair \"$DUMMY_TICKET\" 0x":100=0:0
   sleep 5
   asserter_contract "$DATA_DIRECTORY/0" "$CONTRACT_ADDRESS" "Pair \"$DUMMY_TICKET\" 0x" 100
 }
@@ -410,7 +410,7 @@ test_wasm_full() {
 
   sleep 5
 
- deku-cli create-transaction ./data/0 wallet.json "$CONTRACT_ADDRESS" 100 "Pair \"$DUMMY_TICKET\" 0x" --arg="$ARG" --vm_flavor="Wasm" --tickets="Pair \"$DUMMY_TICKET\" 0x":100
+ deku-cli create-transaction ./data/0 wallet.json "$CONTRACT_ADDRESS" 100 "Pair \"$DUMMY_TICKET\" 0x" --arg="$ARG" --vm_flavor="Wasm" --tickets="Pair \"$DUMMY_TICKET\" 0x":100=0:0
 
   sleep 5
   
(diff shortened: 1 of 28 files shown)


sandbox.sh
```diff
@@ -389,7 +389,7 @@ test_wasm() {
   ARG=$(ticket_handle)
   echo "param: \"$ARG\""
   sleep 5
-  deku-cli create-transaction ./data/0 wallet.json "$CONTRACT_ADDRESS" 100 "Pair \"$DUMMY_TICKET\" 0x" --arg="$ARG" --vm_flavor="Wasm" --tickets="Pair \"$DUMMY_TICKET\" 0x":100
+  deku-cli create-transaction ./data/0 wallet.json "$CONTRACT_ADDRESS" 100 "Pair \"$DUMMY_TICKET\" 0x" --arg="$ARG" --vm_flavor="Wasm" --tickets="Pair \"$DUMMY_TICKET\" 0x":100=0:0
```
renatoalencar, line 392: What the extra zero means?

Zett98: handle:index_inside_argument

src/smart_contracts/contract_vm.ml
```diff
@@ -261,7 +310,9 @@ module Make (CTX : Context.CTX) : S with module Context = CTX = struct
       | 7 -> load ~ctx mem x GET_CONTRACT_OPT
       | 8 -> load ~ctx mem x TRANSACTION
       | 9 -> load ~ctx mem x MINT_TICKET
-      | _ -> failwith "unimplemented"
+      | x ->
+        Format.printf "code  %d\n" x;
```
renatoalencar, line 314: Is that intentional? You could use `Log.debug` for that.

src/smart_contracts/contract_vm.ml
```diff
@@ -173,11 +221,13 @@ module Make (CTX : Context.CTX) : S with module Context = CTX = struct
         let _, amount2 = R.read_amount memory addr int64_offset in
         let data = (ticket_handle, amount1, amount2) in
         let store (handle1, handle2) =
+          Format.printf "%d %d\n" (Int32.to_int handle1) (Int32.to_int handle2);
           W.write_tickets ~writer_addr [handle1; handle2] memory in
         store (Table_ops.split_ticket ctx data)
       | OWN_TICKET ->
         let _, ticket_handle = R.read_ticket memory writer_addr int32_offset in
         let store handle1 = W.write_ticket ~writer_addr handle1 memory in
+        Format.printf "handle %d\n" (Int32.to_int ticket_handle);
```
renatoalencar, line 230: I don't know it those printfs are intentional, but I guess using the logging library would be better, probably with `Log.debug` so we'd be able to control when that appears or not. It'd also make it more visible.

Also, if that's intentional it'd be great to have a more descriptive logging message, instead of just numbers so someone looking at the logging messages for debugging would know what's that about.

sandbox.sh
renatoalencar: But what are the index for ?

src/smart_contracts/contract_vm.ml
Zett98: need to remove them

src/smart_contracts/contract_vm.ml
Zett98: need to remove them

sandbox.sh
Zett98: to have a pointer to ticket handle inside the argument bytes, in case we have a different ticket with the same handle already in the contract storage, so to avoid the conflicts we just rename the handles for now, what's why we need ticket -> handle mapping inside contract, and index + handle inside cli arguments. We need a proper argument and initial argument parsing functionality

renatoalencar (approved):
LGTM

Issue #721, opened 2022-06-23
Test cases for consensus algorithm


kienle371999:
# Summary

A consensus algorithm is the heart of any blockchain system. We need to generate test cases to prove that our consensus algorithm is resistant to common attacks.

# Test cases

**Scenario 1**
- We have three nodes: A, B, and C. These are validators in Deku.
- Node A is the proposer. It produces its own block.
- Node A also signs this block and broadcasts its signature.
- Node B and node C receive this block but they don't sign it and broadcast its signature.
- We have **less than 2/3 validators** agreeing on this block so this block isn't included in Deku.

**Scenario 2**
- We have three nodes: A, B, and C. These are validators in Deku.
- Node A is the proposer. It produces its own block.
- Node A also signs this block and broadcasts its signature. 
- Node B receives this block, signs it, and broadcasts its signature.
- Node C receives this block, signs it, and broadcasts its signature.
- Meanwhile, these validators are suddenly interrupted. 
- Therefore, these signatures aren't properly handled by **validators** so this block isn't included in Deku.

d4hines:
We have these already.

PR #727, opened 2022-06-25
Generate test cases for consensus algorithm

kienle371999:
## Problem

- [ ] #721

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,9 @@ result
 # IDEs
 .vscode
 
+# Vscode history
+.history
+
 # nix-direnv cache
 .direnv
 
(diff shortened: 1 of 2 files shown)


aguillon:
Hi,

I don't think we need those tests in `sandbox.sh`, as I believe they are already special cases of the "smoke test" (see `./sandbox.sh smoke-test`). In `sandbox.sh` we have mostly end-to-end tests, that are not always very reliable but help us check quickly if deku is broken or not. On the other hand, they are rather slow and costly to run.

What we would need, however, is OCaml tests covering specifically the implementation of the consensus algorithm itself: such tests are much quicker to run, easier to maintain and can be useful when refactoring the consensus algorithm.

I'm closing the PR, let's discuss this together with @oteku on Slack.

PR #730, opened 2022-06-27
D4hines/better-network-handling

d4hines:
## Problem

I noticed while working on some other PR's that our network stack doesn't handle errors well - the client doesn't get any notifcation of what went wrong.

This PR sends the error to the client. It also
changes the block-by-level to return a non-200 status code when the requested block is missing, allowing for more interesting scripting possibilities in future PR's.
## Related

- #99


src/bin/deku_node.ml
```diff
@@ -25,8 +25,12 @@ let handle_request (type req res)
       | Ok response ->
         let%await response = Parallel.encode E.response_to_yojson response in
         Dream.json response
-      | Error err -> raise (Failure (Flows.string_of_error err)))
-    | Error err -> raise (Failure err) in
+      | Error err ->
+        Dream.respond ~status:`Internal_Server_Error (Flows.string_of_error err)
```
ulrikstrid, line 29: Do we expect that we log these errors somewhere else? Otherwise it would be good if we could log them here as well.

PR #731, opened 2022-06-27
D4hines/ci-hacking

d4hines:
## Problem

Our CI isn't very reliable - there are a lot of false positives. It's also slow.

The smoke-test was always a temporary measure. This PR replaces it with something slightly more robust.

## Solution

First commit parallelizes a step in the deposit withdraw test that should speed up the CI by a bit.

Second commit adds a new binary that checks that the deku cluster is making progress by posting blocks to the main chain. The nix setup is a bit brittle here because it depends implicitly on `jq` and `curl` - it should ultimately be part of the sandbox rewrite in OCaml that PL is working on. But we can address that further down the line - as long as it's run from `nix develop`, it should have the required dependencies.

The last PR removes the smoke-test and replaces it with this liveness check. The liveness check is faster and should be more reliable.

In subsequent PR's', we'll work on making the deposit-withdraw test more reliable, but this should already help our CI situation.


tests/e2e/check_liveness.ml
```diff
@@ -0,0 +1,57 @@
...
+let curl = process "curl"
+
+let jq = process "jq"
+
+let current_main_chain_height ~rpc_node ~consensus_contract_address =
+  let url =
+    Format.sprintf "%s/chains/main/blocks/head/context/contracts/%s/storage"
+      rpc_node consensus_contract_address in
+  curl ["--silent"; url]
+  (* |. jq [".args[0].args.[0].args[2].bytes.args[0].args[0].args[0].args[1].int"] *)
```
AngryStitch, line 13: I'm not sure this should be kept here?

tests/e2e/check_liveness.ml
```diff
@@ -0,0 +1,57 @@
...
+  |. jq [".args[0].args[0].args[0].args[1].int"; "-r"]
```
ulrikstrid, line 14: We could probably use `Yojson` for a more readable parse here since we're piping the output to OCaml in the next step anyway

ulrikstrid (commented):
Might make sense to make it explicit that this is checking "liveness" of the chain and not the node(s).

d4hines:
> Might make sense to make it explicit that this is checking "liveness" of the chain and not the node(s).

Done in the latest fixup commit

PR #732, opened 2022-06-28
Moved commands from deku-cli to deku-node

Pilou97:
# Problem

The `deku-cli` is doing a lot of things, some features doesn't make sense from a user point of view.

# Solution

Moved some commands from the `deku-cli` to the `deku-node`.
Moved commands:
 - start
 - produce-block
 - sign-block
 - setup-identity
 - setup-tezos
 - add-trusted-validator
 - remove-trusted-validator
 - self (there will be two versions of self, one in `deku-node` and another one in `deku-cli`


aguillon (approved):

d4hines (approved):
LGTM :+1:

PR #733, opened 2022-06-28
Use deku-cli without data folder

Pilou97:
# Problem

The `deku-cli` can only be used if you have a running Deku cluster on your machine. 

# Solution
Remove the access to the data folder of each node.

# Depends

[PR#732](https://github.com/marigold-dev/deku/pull/732)

### DIFF
--- a/src/bin/deku_cli.ml
+++ b/src/bin/deku_cli.ml
@@ -221,7 +221,7 @@ let info_originate_contract =
   Cmd.info "originate-contract" ~version:"%\226\128\140%VERSION%%" ~doc ~exits
     ~man
 
-let originate_contract node_uri contract_json initial_storage sender_wallet_file
+let originate_contract node_uri sender_wallet_file contract_json initial_storage
     (vm_flavor : [`Wasm]) tickets =
   let open Network in
   let%await block_level_response = request_block_level () node_uri in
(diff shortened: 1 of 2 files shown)


src/bin/deku_cli.ml
```diff
@@ -177,61 +159,42 @@ let info_create_transaction =
...
+    | Some addr, None ->
+      Core_deku.User_operation.make ~source:wallet.address
+        (Transaction { destination = addr; amount; ticket })
+    | Some _, Some _ -> failwith "can't pass an argument to implicit account"
+    | None, None -> failwith "Invalid transaction"
+    | None, Some arg ->
+      let payload =
+        match vm_flavor with
+        | `Lambda -> Contract_vm.Invocation_payload.lambda_of_yojson ~arg
+        | `Dummy -> Contract_vm.Invocation_payload.dummy_of_yojson ~arg in
```
aguillon, line 179: This is not relevant anymore.

aguillon (changes requested):
Small changes are needed because we merged the PR removing old VMs; you can delete the `--vm_flavor` parameter.

Except for that, I guess we could improve the UX a bit, as raw exceptions (e.g. when you specify a bad deku node address with get-balance or create-transaction) are a bit harsh. We'll also definitely run into users that don't understand why create-transaction "succeeded" but get-balance shows 0.

aguillon (approved):
ty Daniel

d4hines (approved):
LGTM
Astro boy running toward a good UX

PR #734, opened 2022-06-28
Restrict critical files permissions for Deku nodes

renatoalencar:
## Problem

Any unprivileged user can read identity files and other sensitive information, which could be targeted as privileged escalation attack or could leak the node private's keys.

## Solution

Change umask to restrict the access to those files.


d4hines (approved):
This is great, thanks! Don't know what's up with CI - i suspect it's a false positive.

PR #736, opened 2022-06-28
Improve logging of operation errors

renatoalencar:
## Problem

I had issues with operations that failed silently and also looking for errors on the logs, which can become difficult sometimes because of the amount of information displayed.

## Solution

* Use environment variable to configure color on the output
  of deku-node
* Log error messages from operation errors, like insufficient
  funds and smart contract errors.
* Add simple pretty printing to BLAKE2B module.


d4hines (approved):
LGTM!

PR #742, opened 2022-07-04
rename core_deku to deku_core

Zett98:
## Depends

- [ ] #740
- [ ] #741

## Problem

Core_deku sounds wonky and should be renamed to Deku_core as in it's a core of Deku.
Plus now that Core by jane street is used we have two completions for <Core> and this solves the need to cycle suggestions 
## Solution

Rename it


aguillon:
jim_halpert_face.png

aguillon:
Will merge after we're done merging Daniel stuff (probably on Friday)

PR #745, opened 2022-07-06
PPX for cmdliner

Pilou97:
Add a ppx to add a simpler way to add commands to Cmdliner


d4hines:
@Zett98, was under the impression that ppx_deriving_cmdliner didn't support everything we needed, but I think I was wrong. We're gonna try it and see how far we get. We can resurrect this PR if we find insurmountable limitations in the other ppx.

PR #746, opened 2022-07-06
Bin_prot usage for blocks, signatures and their components

aguillon:
## Problem

In order to increase TPS, [Pollinate](https://github.com/marigold-dev/pollinate/) proposes a new networking layer, which is currently being integrated to Deku. While Pollinate is (allegedly) format-agnostic, the first, easy step towards the TPS goal is to stop serializing stuff to JSON.

## Solution

We propose to use [Bin_prot](https://github.com/janestreet/bin_prot) to serialize the types used in Deku consensus, as they ought to be the most exchanged and the biggest messages in Deku.

We tested a first implementation of the consensus using these serialization functions and observe that nodes are able to encode and decode messages; however, this is not included in the current PR because we have one or two more technicalities to fix.

## Related

* Why Bin_prot instead of X?

I don't know, from what I understood it was a given.

* How does Bin_prot perform?

According to [preliminary tests](https://github.com/marigold-dev/deku/pull/688) from @InfiniteSwerve, better than Yojson.


src/crypto/key.ml
```diff
@@ -57,3 +57,13 @@ let encoding =
 
 let to_yojson, of_yojson =
   Yojson_ext.with_yojson_string "key" to_string of_string
+
+include Helpers.Bin_prot_ext.String_like (struct
+  type nonrec t = t
+
+  let name = "Key"
```
lykimq, line 64: Does `Key` with capital matter? because the `key_hash` in `key_hash.ml` where name is not capitalize?

lykimq (approved):
LGTM! :) Thank you

src/crypto/key.ml
aguillon, line 64:
```suggestion
  let name = "key"
```

PR #751, opened 2022-07-07
D4hines/block by level timestamp

d4hines:
## Problem

Without an indexer (#535) we need a way to extract timestamp data. This is useful for automated benchmarking.

## Solution

Add it to our `block-by-level` endpoint.


InfiniteSwerve (approved):
LGTM

d4hines:
CI fails but I ran test locally and it passed. It's a nondeterministic failure unrelated to this change. Merging with broken CI. We have a fix upstream of this commit.

PR #753, opened 2022-07-07
D4hines/log refactor

d4hines:
## Problem

In #736 we added an option to control logging color. But Logs has a nicer way to do this.

## Solution

Use the idiomatic Logs way.


InfiniteSwerve (approved):
LGTM

PR #755, opened 2022-07-08
protocol: hold only hash of operation on protocol

d4hines:
## Depends

- [ ] #100

## Related

- #99


EduardoRFS:
This is a major performance win BTW

aguillon (approved):
nice

PR #756, opened 2022-07-08
consensus: produce block before applying

d4hines:
## Depends

- [x] #755

## Problem

Right now we block consensus on applying the current block - the block producer won't
produce the next none until he finishes applying the current one. Because the user data
is not parsed until application, this can take a long time.

## Solution

Split block application into the header (consensus inforamation) and data (which is for
the state machine/vm only). Trigger block production as soon as the header is applied.


EduardoRFS:
I'm not very confident on this change, especially as it implies on #757. #758 is nice tho

PR #758, opened 2022-07-08
consensus: handle out of order block arrival

d4hines:
## Depends

- [ ] #757

## Problem

Right now, a a node does not check to see if he has the next block
after applying the current one. This causes him to needlessly wait if
blocks arrive out of order for any reason.

## Solution

Check for the next block after applying the current one.


d4hines:
Closing since we've rewritten the consensus

PR #761, opened 2022-07-11
Feat/pollinate integration validators

AngryStitch:


src/bin/deku_node.ml
```diff
@@ -591,18 +610,33 @@ let info_add_trusted_validator =
...
+    |> validators_payload_to_yojson
     |> Yojson.Safe.to_string in
+
   let payload_hash = BLAKE2B.hash payload_json_str in
   let signature = Signature.sign ~key:identity.secret payload_hash in
-  let%await () =
-    Network.request_trusted_validator_membership { signature; payload }
-      identity.uri in
+  let%await interop_context = interop_context node_folder in
+  let%await validator_uris = validator_uris ~interop_context in
```
aguillon, line 624: Fetching validator_uris through Tezos is actually pretty slow when we wait for several confirmations, so you should not add this operation when it is not needed. Here you're modifying code that wasn't doing that (I believe it actually was removed fairly recently), because it's not needed.

src/bin/deku_node.ml
```diff
@@ -591,18 +610,33 @@ let info_add_trusted_validator =
...
+  match validator_uris with
+  | Error err -> Lwt.return (`Error (false, err))
+  | Ok validator_uris ->
+    let open Network.Pollinate_utils in
+    let validator_uris = List.map snd validator_uris |> List.somes in
+    let recipients = List.map uri_to_pollinate validator_uris in
```
aguillon, line 630: To continue previous comment: you got this functionality wrong. The goal of this command line action is not to add a validator to a chain, it's to tell a given _node_ that it should accept the addition or removal, through a `Consensus_operation`, of a validator. You tell just one node: the one for which you have the node folder.

Hence, no list of URIs to send messages to.

src/bin/deku_node.ml
```diff
@@ -591,18 +610,33 @@ let info_add_trusted_validator =
...
+    let%await pollinate_node = state.pollinate_node in
+    let operation = create_operation ~name:Validators_change ValidatorsOperation in 
+    let%await () =
+      let open Network in 
+      send_over_pollinate
+        (module Validators_change)
+        pollinate_node { signature; payload } ~operation recipients in
```
aguillon, line 637: and `recipients` should definitely be only one node.

src/consensus/consensus.ml
```diff
@@ -26,8 +26,8 @@ let with_timeout effect state = Dispatch.dispatch effect Can_produce_block state
 let with_operation effect operation state =
   Dispatch.dispatch effect (Check_operation { operation }) state
 
-let with_trusted_validators_membership_change effect ~payload ~signature state =
-  Dispatch.dispatch effect (Check_validator_change { payload; signature }) state
+let with_trusted_validators_membership_change effect ~payload ~signature ~payload_hash state =
+  Dispatch.dispatch effect (Check_validator_change { payload; signature; payload_hash }) state
```
aguillon, line 30: Just to be sure: are you adding this because you don't know how to decode in consensus/steps.ml:check_validator_change as you don't know if you got there with JSON or bin_prot? It would make sense to me.

PR #763, opened 2022-07-12
parallel: decode user operations

EduardoRFS:
## Problem

Decoding user operations in Deku means parsing them from JSON, hashing them to ensure the hash matches and then verifying the signatures, this is a very slow operation and our current bottleneck.

The Protocol.apply_block on main allows only to up to 10k TPS.

## Solution

PARALLELISM. By decoding the user operations in parallel this allows us to go way faster, this PR on my machine increases the TPS limit from 10k TPS to up to 120K TPS.


src/helpers/parallel.ml
```diff
@@ -21,6 +28,25 @@ let decode of_yojson data =
   with
   | exn -> Error exn
 
+let protocol_decode_list of_yojson data =
```
Zett98, line 31: why not do this with `Task.parallel_for?

src/helpers/dune
```diff
@@ -13,6 +13,7 @@
   fmt.cli
   logs.cli
   fmt.tty
-  json-logs-reporter)
+  json-logs-reporter
+  base)
```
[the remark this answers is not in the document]
EduardoRFS: this is not core_deku

src/helpers/parallel.ml
```diff
@@ -1,12 +1,19 @@
 open Lwt_ext.Let_syntax
+open Domainslib
 
 (* lazy because Unix.fork stops working after domains start *)
 (* TODO: proper number of domains *)
 (* THIS IS GLOBAL STATE OUTSIDE OF src/node/server.ml. CAUTION *)
-let pool = lazy (Lwt_domain.setup_pool 8)
+let domains = 32
```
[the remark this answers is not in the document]
EduardoRFS: This is another patch feel free, 32 is a better number to benchmark

src/helpers/parallel.ml
EduardoRFS: I don't want to deal with numbers or mutation and parallel_for requires both, we actually had problems regarding that, parallel_for is like a for loop

Zett98: We use libraries that leverage mutability all the time and still going, besides it's not like the code for parallel_map will be modified all the time.
code for parallel_map is taken from domainslib pr: 
```ocaml
let parallel_map arr ~pool ~f=
  let open Task in
  let len = Array.length arr in
  let res = Array.make len (f arr.(0)) in
  Task.parallel_for ~chunk_size ~start:1 ~finish:(len - 1)
  ~body:(fun i ->
    res.(i) <- (f arr.(i))) pool;
  res
 ```
 and then the code becomes(assuming user_operations are represented as `string array` instead of `string list`: 
 ```ocaml
 let protocol_decode_batch data ~of_yojson  = 
   parallel_map data 
   ~pool:(domainslib_pool ()) 
   ~f:(fun elem ->
               match decode of_yojson data with 
               | Ok (Ok data) -> Ok data
               | Ok (Error error) -> Error (`Of_yojson error)
               | Error exn -> Error (`Of_exception exn))
  |> Array.to_list
 ```

PR #765, opened 2022-07-13
add receipts for contract-calls/originations

Zett98:
## Problem

We need to know the results of contract origination/calls for indexing.
## Solution

Add receipts for contract origination/ contract calls

### DIFF
--- a/src/core_deku/state.ml
+++ b/src/core_deku/state.ml
@@ -9,7 +9,32 @@ type t = {
 }
 [@@deriving yojson]
 
-type receipt = Receipt_tezos_withdraw of Ledger.Withdrawal_handle.t
+type contract_invocation_changes = {
+  remaining_tickets : (Ticket_id.t * Amount.t) list;
+  additional_operations : additional_operation list;
+  new_storage : bytes;
+}
+
+and additional_operation = {
+  user_operation : User_operation.t;
+  operation_receipt : receipt option;
+}
+
+and contract_origination_changes = {
+  originated_contract_address : Contract_address.t;
+  initial_tickets : (Ticket_id.t * Amount.t) list;
+}
+
+and receipt =
+  | Receipt_tezos_withdraw       of Ledger.Withdrawal_handle.t
+  | Receipt_contract_origination of {
+      sender : Address.t;
+      outcome : [`Success of contract_origination_changes | `Failure];
+    }
+  | Receipt_contract_invocation  of {
+      sender : Address.t;
+      outcome : [`Success of contract_invocation_changes | `Failure];
+    }
 [@@deriving yojson]
 
 let empty = { ledger = Ledger.empty; contract_storage = Contract_storage.empty }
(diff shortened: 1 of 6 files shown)


src/core_deku/state.ml
```diff
@@ -9,9 +9,39 @@ type t = {
...
+  initial_tickets : (Ticket_id.t * Amount.t) list;
+}
+
+and receipt =
+  | Receipt_tezos_withdraw       of Ledger.Withdrawal_handle.t
+  | Receipt_contract_origination of {
+      sender : Address.t;
+      outcome : [`Success of contract_origination_changes | `Failure];
+    }
+  | Receipt_contract_call        of {
```
renatoalencar, line 34: This should probably be named `Receipt_contract_invocation` in order to keep naming convention. As well as `contract_invocation_changes`.

src/core_deku/state.ml
```diff
@@ -9,9 +9,39 @@ type t = {
...
+      sender : Address.t;
+      outcome : [`Success of contract_call_changes | `Failure];
+    }
 [@@deriving yojson]
 
+type step = {
```
renatoalencar, line 40: Does step (the name) really makes sense in this context?

src/core_deku/state.ml
```diff
@@ -101,113 +135,176 @@ let rec apply_user_operation ?sender t operation_hash user_operation =
...
-                  acc op)
-              ~init:({ ledger; contract_storage }, None)
-              operations in
-          Ok (receipt, t)) in
-    Ok (t, receipt)
-
-and apply_contract_ops ~source ~sender (t, _) x =
+        let%ok tickets_mapping, tickets, operations =
+          ctx#finalize user_op_list
+          |> Result.map_error ~f:(fun _ ->
```
renatoalencar, line 210: What is the type of `_`? Could you change that to `()` or `_<name>` in order to make it more explicit?

src/core_deku/state.ml
```diff
@@ -101,113 +135,176 @@ let rec apply_user_operation ?sender t operation_hash user_operation =
...
   let report_error error =
     let message =
       match error with
-      | `Origination_error msg -> "Origination error: " ^ msg
-      | `Invocation_error msg -> "Invocation error: " ^ msg
-      | `Insufficient_funds -> "Insufficient funds" in
+      | `Origination_error (msg, _) -> "Origination error: " ^ msg
+      | `Invocation_error (msg, _) -> "Invocation error: " ^ msg
+      | `Insufficient_funds -> "Insufficient funds"
+      | `Invalid_payload _ -> "Invalid contract payload"
```
renatoalencar, line 297: Please, don't ignore error messages and log it back.

renatoalencar (changes requested):

src/core_deku/state.ml
Zett98: This was the receipt, but this code is overwritten now, so I don't get the comment

src/core_deku/state.ml
Zett98: Wdym?

renatoalencar (approved):
LGTM

PR #766, opened 2022-07-14
Add proxy contract for deku

kienle371999:
## Problem

The proxy contract is the bridge between the Tezos devs and the internal contracts in Deku. Specifically, it is a simple contract with map storage where the key is the validator or bridge address and the value is the corresponding contract. There are some problems related to creating this contract

- How to safely deploy and update the map storage
- How to detect the latest contract and update the value in the proxy contract

## Solution
- For safely deploying and updating, I locally generate the validator and bridge account, sign the updated contract, and include this signature in the payload. In the internal contract, I use **Crypto.check** in the `ligo codebase` to authenticate the correctness of this newly-submitted signature. 
- For detection of the latest contract, I include the scripts for deploying and updating in **create_new_deku_environment** function in `sandbox.sh`

### DIFF
--- a/proxy-contract/main.mligo
+++ b/proxy-contract/main.mligo
@@ -1,0 +1,13 @@
+#import "errors.mligo" "Errors"
+#import "parameter.mligo" "Parameter"
+#import "storage.mligo" "Storage"
+
+type address_update = Parameter.Types.t
+type storage = Storage.Types.t 
+
+let main (address_update, storage : address_update * storage) = 
+  let { key; address; signature } = address_update in 
+  let key_hash = Crypto.hash_key key in 
+  let () = Storage.Utils.check_signature key address signature in 
+  let storage = Big_map.add key_hash address storage in 
+  (([] : operation list), storage)
--- a/proxy-contract/errors.mligo
+++ b/proxy-contract/errors.mligo
@@ -1,0 +1,1 @@
+let invalid_signature : string = "invalid signature"
--- a/proxy-contract/parameter.mligo
+++ b/proxy-contract/parameter.mligo
@@ -1,0 +1,8 @@
+module Types = struct
+  (* The updated contract info for the proxy contract *)
+  type t = {
+    key : key; (* The public key of the validator or bridge address *)
+    address : address;
+    signature : signature;
+  }
+end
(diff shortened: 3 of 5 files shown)


ulrikstrid:
sandbox.sh is only used for local development and is also deprecated (removed?) in favour of  the ocaml sandbox.
The best way to allow us to use it when doing the deployment is to expose it via nix currently.

proxy-contract/main.mligo
```diff
@@ -0,0 +1,13 @@
+#import "errors.mligo" "Errors"
+#import "parameter.mligo" "Parameter"
+#import "storage.mligo" "Storage"
+
+type address_update = Parameter.Types.t
+type storage = Storage.Types.t 
+
+let main (address_update, storage : address_update * storage) = 
+  let { key; address; signature } = address_update in
```
oteku, line 9: This is not working.
There is some security issues:
- Anyone can submit an operation signed by the originator of the contract and erase with deprecated values. When you want to now the sign an operation access to `Tezos.get_sender ()` do not use extra parameters
- Any contract can add its key to the storage.

All you need to do is:
- add a `manager_address` in the storage, and originate with the address that originate the contract
- check `Tezos.get_sender () = manager_address` => update contract addresses in the storage is an OPS operation we want to do whenever a new contract is originated (so probably in CI)

proxy-contract/main.mligo
```diff
@@ -0,0 +1,13 @@
...
+#import "storage.mligo" "Storage"
+
+type address_update = Parameter.Types.t
+type storage = Storage.Types.t 
+
+let main (address_update, storage : address_update * storage) = 
+  let { key; address; signature } = address_update in 
+  let key_hash = Crypto.hash_key key in 
+  let () = Storage.Utils.check_signature key address signature in 
+  let storage = Big_map.add key_hash address storage in
```
oteku, line 12: We just need to manage 2 keys, big_map is overkill

proxy-contract/main.mligo
```diff
@@ -0,0 +1,13 @@
+#import "errors.mligo" "Errors"
+#import "parameter.mligo" "Parameter"
+#import "storage.mligo" "Storage"
+
+type address_update = Parameter.Types.t
+type storage = Storage.Types.t 
+
```
oteku, line 7: We need to provide onchain views:

```
[@view] let get_bridge_address ((), store : unit * storage) : address = ...
[@view] let get_discovery_address ((), store : unit * storage) : address = ...
```

oteku (changes requested):

PR #768, opened 2022-07-20
Alphanet changes

d4hines:
I'd like to reduce the distance between main and alphanet. To that end, this PR packages together several changes from `d4hines/devleop`. Each commit is atomic - please review them 1-by-1.

When we merge this, let's do a rebase, not a squash.


src/network/network_client.ml
```diff
@@ -51,4 +51,8 @@ let broadcast_to_list (type req res)
   uris
   (* TODO: limit concurrency here *)
   |> Lwt_list.iter_p (fun uri ->
-         Lwt.catch (fun () -> raw_post E.path data uri) (fun _exn -> await ()))
+         Lwt.catch
+           (fun () -> raw_post E.path data uri)
+           (fun exn ->
+             Log.error "Broadcast exception: %a\n%!" Fmt.exn exn;
```
aguillon, line 57: We may want to add more information here? e.g., but not necessarily as such:
```ocaml
let endpoint = E.path in
let uri_s = Uri.to_string uri in
Log.error "Broadcast exception: %a when sending %s to %s\n%!" Fmt.exn exn endpoint uri_s;
```

src/bin/deku_node.ml
```diff
@@ -15,6 +15,82 @@ let exits =
 let man = [`S Manpage.s_bugs; `P "Email bug reports to <contact@marigold.dev>."]
 
 (* TODO: several functions copied from deku-cli. Refactor. *)
+type json_option =
+  | No
+  | To_stdout
+  | To_stderr
```
aguillon, line 21: The problem with these options is that they include messages from the bridge as strings, whereas it should probably be included as JSON too. As I said in another PR, I don't exactly know how the JSON logs are subsequently used so it's hard to know if it's relevant or not.

Moreover, and keeping the previous comment in mind, I believe the node identity should be included in the JSON output. Although each node most likely runs in its own space and has its own log file/stdout/journal/whatever, I think it would make it easier to analyze the logs.

aguillon (approved):
Very minor remarks, looks good to me.

src/bin/deku_node.ml
d4hines: These are excellent points that I don't want to address here :sweat_smile:

src/network/network_client.ml
d4hines: Done.

src/files/node_state.ml
```diff
@@ -3,6 +3,28 @@ open Protocol
 open Node
 open Consensus
 
+(* FIXME: refactor below function to use this one *)
```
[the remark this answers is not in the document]
d4hines: Changed to `TODO:`

PR #769, opened 2022-07-20
Switch to OCaml Sandbox

d4hines:
## Depends

- [ ] #768 

## Problem

This PR refactors and expands our OCaml sandbox. It squashes a bunch of changes from `d4hines/develop` into a single commit. Yes, this is bad practice, and I'm sorry.

The reason we need this is that the `sandbox.sh` script is getting way to big and complex. With this PR we can use a more sane language than Bash to manage things.

We've already been using this code to develop and deploy alphanet.


d4hines:
I've tested `tilt ci` many times locally and it works most of the time. I'm still seeing a very occasional failure and I don't understand why but I think it's still related to deposit withdraw test. I think we can merge this without that fix.

PR #771, opened 2022-07-22
fix: wrap operation application in try/with

d4hines:
## Problem

We don't handle exceptions in the application of protocol operations. 

## Solution
Use `try...with`


EduardoRFS (approved):
little anime girl approving a cute PR

PR #772, opened 2022-07-22
WIP Cookie game

gsebil08:
Add real example and tutorial about "how to make and use dapp"


examples/cookie-game/src/index.ts
```diff
@@ -0,0 +1,336 @@
...
+        sourceValue.cookieBaker.numberOfFreeIdleverse,
+        sourceValue.cookieBaker.numberOfFreeCordex,
+    );
+
+
+    switch (operation) {
+        case actions.incrementCookie: {
+            const updatedCookieBaker = addCookie(cookieBaker);
+            //update state
+            sourceValue.cookieBaker = updatedCookieBaker;
```
gsebil08, line 81: I guess we can do this, it will be better and easier to read

```suggestion
            sourceValue.cookieBaker = addCookie(cookieBaker);
```

examples/cookie-game/src/index.ts
```diff
@@ -0,0 +1,336 @@
...
+
+const transition = (tx: transaction) => {
+    // source -> tz1 address
+    // op_hash / tx_hash => BLAKE2B => resolved as string
+    // operation => any
+    const source = tx.source;
+    const operation = tx.operation;
+    console.log("Getting source");
+    const sourceValue = JSON.parse(get(source));
+    const cookieBaker: cookieBakerType = createCookieBaker(
```
gsebil08, line 32: What about reworking the `createCookieBaker` function to avoid these 30+ lines?
Would be great to do something like:

```typescript
const cookieBaker: cookieBakerType = JSON.parse(get(source));
```
and hide all this complexity inside `createCookieBaker`

examples/cookie-game/src/index.ts
```diff
@@ -0,0 +1,336 @@
...
+import {
+    createCookieBaker, addCookie, addCursor,
+    addGrandma, addFarm, addMine, addFactory, addBank, addTemple,
+    addWizard, addShipment, addAlchemy, addPortal, addTimeMachine,
+    addAntimatter, addPrism, addChanceMaker, addFractal, addJavascript,
+    addIdleverse, addCordex
+} from "./state"
+import { actions } from "./actions"
+
+const printMessageWithSource = (message: string, source: transaction) => {
```
gsebil08, line 13: I guess we can get rid of this one, since it is only used inside `saveState`

examples/cookie-game/src/index.ts
```diff
@@ -0,0 +1,336 @@
...
+            break;
+        }
+    }
+}
+
+main(
+    //tz address must be replaced by a correct one obtained with 
+    //deku-cli create-wallet
+    {
+        "tz1VULT8pu1NoWs7YPFWuvXSg3JSdGq55TXc":
```
gsebil08, line 264: We need to rework this default state. It can't use a default address. I guess the only correct choice is an empty one.
Then, we need to manage the address here. Maybe in the payload? Like `{"address":"tzxxxxxx", "action":"cookie"}`
What do you think?

examples/cookie-game/src/state_types.ts
```diff
@@ -0,0 +1,132 @@
...
+export const init_timeMachineCost: number = 14e12;
+export const init_antimatterCost: number = 17e13;
+export const init_prismCost: number = 2.1e14;
+export const init_chanceMakerCost: number = 26e15;
+export const init_fractalCost: number = 31e16;
+export const init_javascriptCost: number = 71e18;
+export const init_idleverseCost: number = 12e21;
+export const init_cordexCost: number = 1.9e23;
+
+export type cookieBakerType = {
```
gsebil08, line 48: `type cookieBakerType` is redundant. We should go for `type cookieBaker`

examples/cookie-game/src/state_types.ts
```diff
@@ -0,0 +1,132 @@
...
+    numberOfCookie: number;
```
gsebil08, line 49: I think we should use shorter names, like `cookies`, `cursors`, etc.
Writing `numberOfCookies` is kinda really long for nothing interesting...

examples/cookie-game/src/state.ts
```diff
@@ -0,0 +1,692 @@
...
+    cookieBaker.prismCps = prismCps;
+    cookieBaker.chanceMakerCps = chanceMakerCps;
+    cookieBaker.fractalCps = fractalCps;
+    cookieBaker.javaScriptCps = javaScriptCps;
+    cookieBaker.idleverseCps = idleverseCps;
+    cookieBaker.cordexCps = cordexCps;
+    return cookieBaker;
+}
+
+export const calculateCost = (action: actions, cookieBaker: st.cookieBakerType): number => {
```
gsebil08, line 227: Should this function be there? Or something else? No strong opinion, what do you think?

examples/cookie-game/src/state_types.ts
```diff
@@ -0,0 +1,132 @@
+import { actions } from "./actions"
```
gsebil08, line 1: This file should be renamed to `types`. Maybe inside a `state` package, with `state.ts`. What do you think?

gsebil08 (commented):
Thanks for adding all these levels!

examples/cookie-game/src/state.ts
lykimq: I think for now, this function can be in here, but maybe later if we have more calculation function then we can move it to `utils` or something like that.

d4hines:
Already merged

PR #774, opened 2022-07-22
Add logging to ledger

d4hines:
## Problem
We'd like some better logs for ledger transactions to support the alphanet.

## Solution
Add logs.


d4hines:
@EduardoRFS makes a good point that we should avoid this kind of ad-hoc logging in favor of a more principled and complete way to extract information from the chain. We have this with #759

Issue #773, opened 2022-07-22
Change type of vault_deposit.address from address to key_hash


d4hines:
Right now it's possible to deposit a ticket to a KT1 address on Deku side. But this isn't supported by Deku itself. We should change the type in consensus.mligo from address to key_hash. We should change the types in Deku side from a Tezos_address to a Key_hash as well.

d4hines:
This will make this failwith with go away: https://github.com/marigold-dev/deku/blob/main/src/core_deku/state.ml#L39

d4hines:
Done.

PR #777, opened 2022-07-25
Use key_hash instead of address in contracts + deku

aguillon:
## Problems

As mentioned in https://github.com/marigold-dev/deku/issues/773 the types used in the current contracts (consensus + dummy_ticket) allow to deposit tickets to an originated account ("KT1 address"). Moreover, in various files, types allow all Tezos addresses as targets to a deposit, leading to the use of `failwith`.

```bash
bad_address="KT1NHBasfrhX9d2SgGVARd3Y1C5nkhfoEks3"
tezos-client transfer 0 from alice to $dummy_ticket --entrypoint mint_to_deku --arg "(Pair (Pair \"$deku_consensus\" \"$bad_address\") (Pair 100 0x))" --burn-cap 0.05
# works on main :(
```

## Solution

1. Change the two aforementioned contracts to prevent this
2. Modify `tezos_interop` to use a Key_hash.t instead of an Address.t and remove everything that's now redundant

## Tests

I tested deposit and withdrawal manually. I _did not_ test TzPortal against this; this will likely require updating TzPortal contracts as well (because of types).

## Discussion

This does not support depositing tickets to a Deku-originated contract ("DK1 address"). On the other hand, these are not supported by tezos-client either.

### DIFF
--- a/dummy_ticket.mligo
+++ b/dummy_ticket.mligo
@@ -6,7 +6,7 @@
 
 type blake2b = bytes
 type vault_ticket = bytes ticket
-type vault_deposit = { ticket : vault_ticket; address : address }
+type vault_deposit = { ticket : vault_ticket; address : key_hash }
 
 type vault_handle_id = nat
 type vault_handle_structure = {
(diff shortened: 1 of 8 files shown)


d4hines (approved):
LGTM!
Sora says "nice keys, bro"

PR #779, opened 2022-07-27
fix: better exit code

Pilou97:
# Problem

When the node crashes, it always returns 0 code.

# Solution
Returns a non zero code in case of an exception


d4hines (approved):
LGTM!
jigglymight says yes

PR #780, opened 2022-07-28
test: repr of an operation

Pilou97:
# Problem

A bug in the module Repr of Protocol.Operation:

```ocaml
let operation = transaction |> Operation.yojson_of_t |> Operation.t_of_yojson (*where transaction is of type Operation.t)*
```
Throws an exception because the signature is not valid


d4hines:
This has already been merged.

PR #781, opened 2022-07-28
setup ocaml 5

d4hines:
lmk if ocaml-lsp doesn't work. I think it's enough to replace the source of jsonrpc, but if not we can an entry to the overlay.


d4hines:
Closing because it was cherry-picked.

Issue #770, opened 2022-07-20
Deku V1 Mega-issue


d4hines:
This is an issue to track the stuff we're working on over the next 4 weeks as part of the "deku v1" effort. It's not exhaustive, but mainly to make write down the stuff we're likely to forget.j

- [ ] tests to catch when the genesis hash changes. 
- [ ] benchmark all the things
- [ ] A nice tilt setup

d4hines:
- [ ] Check all usages of `deriving`

d4hines:
- [ ] How to signal to users the results of operations, especially failures (transfers that fail, contract calls that fail, etc. etc.)

d4hines:
- [ ] our debugger is only capturing applied blocks, which is missing most of the trickier failures in Alphanet.
- [ ] Laurent says Deku  nodes are  exiting with status 0 (PL has an easy fix)
- [ ] Our log levels aren't right and aren't consistent. We need standardize them. Soon infra is going to send us slack nottications for every error and this going to get annoying real fast.
- [ ] Lwt.async will crash our program if there's an exception, and there are currently lots of opportunities for this in Deku v0

d4hines:
- [ ] What do we do if State.bin is corrupted?
- [ ] Why is state.bin getting corrupted so frequently? (PL checked and `Lwt.write` is named to be atomic in the documentation).

d4hines:
- [ ] write unit tests for protocol and consensus.

InfiniteSwerve:
- [ ] Check all usages of `signature.t` and possibly replace them with `verified_signature.t`

d4hines:
Testing

## Protocol

- Many operations
- Duplicated operation in same block
- Duplicated operation in different blocks
- Late duplicated operation(after includable window)
- Balances are actually correct
- Invalid string operation
- Invalid signed operation
- Valid key, signature to different source operation
- Ensure receipt is only in place if operation was applied

## Consensus

- Old unsigned blocks are ignored
- Old signeds blocks are ignored
- Old blocks are never valid
- Ensure previous should point to previous block
- Valid blocks with not expected author are not signed
- Valid blocks with expected author are signed
- Invalid blocks with not expected author are not signed
- Invalid blocks with expected author are not signed
- On timeout next producer actually produces blocks
- Verify if the validators are rotating properly when not producing(when timeout happens)
- Ensure 2/3+1 signatures is needed to verifier to accept a block
- Ensure Block is always signed by author on
- Test Validators.skip
[...]

PR #783, opened 2022-08-02
Test v1/protocol

Pilou97:
Add tests of the protocol part from 
https://github.com/marigold-dev/deku/issues/770#issuecomment-1201465297

There is one test per scenario:

- [x] Many operations
- [x] Duplicated operation in same block
- [x] Duplicated operation in different blocks
- [x] Late duplicated operation(after includable window) ⚠️   Doesn't works in v1 but it's fixed in this branch 😅 
- [x] Balances are actually correct
- [x] Invalid string operation
- [x] Invalid signed operation
- [x] Valid key, signature to different source operation
- [x] Ensure receipt is only in place if operation was applied

### DIFF
--- a/src/protocol/Included_operation_set.ml
+++ b/src/protocol/Included_operation_set.ml
@@ -1,4 +1,3 @@
-open Deku_stdlib
 open Deku_concepts
 open Operation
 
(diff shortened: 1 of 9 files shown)


d4hines (approved):
LGTM!

d4hines:
This has been merged into `v1-alphanet` already.

PR #790, opened 2022-08-04
[v1] Add support for bootstrapper

d4hines:
Ports PL's bootstrapper work from main into v1.


d4hines:
Cherry-picked already

PR #791, opened 2022-08-05
Test v1/consensus

aguillon:
Add tests for https://github.com/marigold-dev/deku/issues/770#issuecomment-1201465297


aguillon:
Need to be reworked with new consensus.

PR #794, opened 2022-08-15
Try adding nix cache

ulrikstrid:
## Depends

- [ ] #100

## Related

- #99


ulrikstrid:
Doesn't really seem to work

PR #797, opened 2022-08-23
api: api specification proposition

Pilou97:
API specification for a deku-node


api.json
```diff
@@ -0,0 +1,653 @@
...
+                ],
+                "responses": {
+                    "200": {
+                        "description": "Consensus operation accepted."
+                    }
+                }
+            }
+        },
+        "/consensus/operations": {
+            "post": {
```
oteku, line 284: - could you define `requestBody` for post operation?
- Is this also the endpoint used for customVM (replacement of user-operation-gossip)?

Pilou97: yes, the  `/consensus/operations` is the endpoint to send an operation to the node from a user. Then the node will broadcast it to other nodes

api.json
```diff
@@ -0,0 +1,732 @@
...
+                    "consensus"
+                ],
+                "responses": {
+                    "200": {
+                        "description": "Consensus operation accepted."
+                    }
+                }
+            }
+        },
+        "/consensus/operations": {
```
EduardoRFS, line 283: Minor, this is not a consensus operation, so probably only `/operations`

Major, this needs to be done in two steps, because they need to sign a message for us, as discussed with @d4hines the idea is to have an endpoint like "/make-operation" which will return you some bytes that you should sign and then submit the bytes + signature to "/operations".

Look at how Tezos does.

aguillon (changes requested):
I don't think it's possible to force `/states/wasm`, `/states/unix` and `/states/ledger` like that. Can we find a generic endpoint that gets specialized depending on the chain parameter?

Also, `/consensus/` is a bit vague, are you going for `/consensus/block`, `/consensus/signature` and `/consensus/bootstrap`?

api.json
```diff
@@ -0,0 +1,732 @@
...
+                ],
+                "responses": {
+                    "200": {
+                        "description": "blocks",
+                        "headers": {
+                            "Transfer-Encoding": {
+                                "schema": {
+                                    "type": "string",
+                                    "description": "Used only to stream blocks.",
+                                    "example": "chunked"
```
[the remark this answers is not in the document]
Pilou97: This endpoint aims to stream blocks,
Users will be able to connect to this http stream and receive every block applied by the chain. So that users will have the ability to listen/subscribe to the applied blocks.
When a user open an http stream, the first streamed block is the block at the current level of the chain, (the next block is then at level + 1, etc...)
But I still need to know if an http stream will impact the performance of the chain.

api.json
```diff
@@ -0,0 +1,802 @@
...
+        },
+        "/chain/level": {
+            "get": {
+                "description": "Returns the current level of the chain.",
+                "tags": [
+                    "chain"
+                ],
+                "responses": {
+                    "200": {
+                        "description": "Returns the current level of the chain as a string representing an int64.",
```
gsebil08, line 37:
```suggestion
                        "description": "Get the string representation of the int64 of the current chain level.",
```

api.json
```diff
@@ -0,0 +1,802 @@
...
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        },
+        "/chain/blocks": {
+            "get": {
+                "description": "Streams blocks starting at the current level of the chain",
```
gsebil08, line 61: Since this endpoint only streams from the actual moment you call it, will it be possible to retrieve the previous blocks?
Maybe adding a parameter like `starting_block` could be great? Which will make this endpoint to stream from a given point.
This is a naive question, I don't know if it is relevant to do so.

api.json
```diff
@@ -0,0 +1,802 @@
...
+                    "chain"
+                ],
+                "responses": {
+                    "200": {
+                        "description": "blocks",
+                        "headers": {
+                            "Transfer-Encoding": {
+                                "schema": {
+                                    "type": "string",
+                                    "description": "Used only to stream blocks.",
```
gsebil08, line 72:
```suggestion
                                    "description": "Used to stream blocks.",
```

api.json
```diff
@@ -0,0 +1,802 @@
...
+                }
+            }
+        },
+        "/operations": {
+            "post": {
+                "description": "Sends an operation.",
+                "tags": [
+                    "operations"
+                ],
+                "requestBody": {
```
gsebil08, line 289: I am wondering about this payload. It is really annoying (to don't say difficult) to build from client side.
Why can't we have one `PUT/POST` endpoint per kind of transaction?
Something like:
- `POST /operations/transaction`
- `POST /operations/vm_transaction`
- `POST /operations/tezos_withdraw`
?

An other naive question: why do we as for block-level? Needing this information, makes the client call the `/block/level` endpoint for each time he/she wants to post an operation. Hence, for each operation, we need 2 HTTP calls 😕 

Same question for the `hash`s, Deku could calculate them...

And last question: Where is the `signature` field? 🤔

api.json
```diff
@@ -0,0 +1,802 @@
...
+                                "schema": {
+                                    "type": "object"
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        },
+        "/helpers/hash-operation" : {
```
gsebil08, line 476: I think this is great to provide them to help our clients. But I think it is way more great, to hash the operation for the client under the hood.
I feel like the only use cases for this one, will be:
- oh, I forgot what was the hash of this operation (almost impossible, because I would need to remember the details of each operation I submitted)
- I don't know how to hash, please tell me what would be the hash of this operation I want to submit.
If the second use case looks more realistic, it could lead to have some clients performing 3 HTTP request to submit an operation:
- retrieve block level
- get correct hash
- submit operation
We should be careful on that, because it could add a lot of load on the nodes.

api.json
Pilou97: It's complicated to answer this question, should the search of old blocks be part of the chain or an indexer ? 
This endpoint is only used to observe when an operation is applied:
 - you send an operation
 - listen to block
 - until you see your operation in a block
 - you know the status of your operation (accepted or not)
This is not the best way to know the status of your operation, and maybe in the future we will add an endpoint to know the status of your operation

api.json
Pilou97: Yes, it's mainly for the second use case: a client doesn't know how to hash the operation/packet (because there isn't any hashing library in its favorite language for instance).
It's just a helper, you don't have to use it, for instance in the deku-toolkit I am writing in typescript does not relay on this endpoint, it hashes the payload in the client side

api.json
Pilou97: We need the block level for security reason, it helps to counter replay attack
I've forgotten the signature, but you are totally right, it will be part of the payload

api.json
ulrikstrid: What if no indexer is running when the chain starts? Are those blocks lost?
I think we want a notion of a "archive node" that just hold all the blocks and allows you to query them.

PR #798, opened 2022-08-30
Install latest ligo via the static binary

ulrikstrid:
## Problem

We want the latest LIGO version

## Solution

Just download the static binary and put it in the store


aguillon:
Thanks! I think that will do for now.

PR #800, opened 2022-08-31
V1 tezos interop

aguillon:
Changes chain to commit on Tezos. As indicated in chain.ml, this is level-based and not time-based, which is not ideal. Also the consensus should watch for those updates. Also, better tests.

Manual tests: start the chain and see that the contract storage gets updated. This should not break deposits either.


aguillon:
Fixed according to @Pilou97's offline comments

PR #802, opened 2022-09-13
V1 withdraws + withdraw proofs

aguillon:
Clean version of the withdraw proofs. It needs some fixes that @d4hines has but that I couldn't find again.

## Depends

- https://github.com/marigold-dev/deku/pull/800

The bash test is annoying; I initially tried to make it automatic but it wasn't very convincing/failsafe (I'm open to suggestions though).

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -24,6 +24,8 @@ with ocamlPackages; buildDunePackage rec {
     mirage-crypto-ec
     mirage-crypto-rng
     secp256k1-internal
+    cohttp
+    cohttp-lwt-unix
     piaf
     domainslib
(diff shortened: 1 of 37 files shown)


nix/default.nix
```diff
@@ -24,6 +24,8 @@ with ocamlPackages; buildDunePackage rec {
     mirage-crypto-ec
     mirage-crypto-rng
     secp256k1-internal
+    cohttp
+    cohttp-lwt-unix
```
aguillon, line 28: I had a Piaf version initially but it wasn't working (POST requests were failing silently). I'll reproduce it so we can decide.

PR #804, opened 2022-09-17
Parametric Deku

d4hines:
Adds Deku-P support from v0 to v1. 

There are lots of FIXME's and I haven't tested the code, but I wanted to get something out quick so people could see the approximate interface.

One thing to note is that  I've modified the external vm protocol to include tickets in the transaction, but I haven't yet modified any of the SDK's, so they're incompatible.


d4hines:
Merged already

PR #805, opened 2022-09-17
support send on the network

EduardoRFS:
## Depends

- [ ] #803

## Problem

Currently there is no way to send a message to a specific node in Deku, this is

## Solution

This makes the network handle the `validator -> uri` map, while it is not ideal especially as validators may want to change their uris, this should work nicely.


d4hines:
Already merged via #821

PR #808, opened 2022-09-18
merge gossip on chain

EduardoRFS:
## Depends

- [ ] #807

## Problem

The chain is intended to represent all the internal logic of a blockchain, but currently this doesn't include gossip, which is extremely important for the chain and should also be serialized.

## Solution

This moves the gossip logic to inside of Chain.t, this ensures that testing of Chain.t represents all the state and pure logic of Deku.

While this makes some testing harder(like, applying same message twice), those kinds of testing should probably happen on smaller components anyway.


d4hines:
Already merged via #821

PR #812, opened 2022-09-19
better request / response system

EduardoRFS:
## Depends

- [ ] #808

## Problem

Current request / response system spams the network and is not very reliable.

## Solution

This implements a better request / response system, it is still not great, but should be good enough for deployment.


d4hines:
Already merged via #821

PR #814, opened 2022-09-20
loads data from disk

EduardoRFS:
## Depends

- [ ] #812

## Problem

Deku currently is not capable of starting after being stopped.

## Solution

This writes the entire Chain.t to disk and tries to load it whenever starting a node, when coupled with the auto rejoin of the network, this allows to both stop a single node and restart it, but also stop the entire network and restart it by adding a single block, which will make all nodes slightly behind request the blocks and get in sync.

This contains some performance regressions that I will be addressing in the near future.


d4hines:
Already merged via #821

PR #817, opened 2022-09-20
V1 test consensus

EduardoRFS:
## Depends

- [ ] #814

## Problem

We don't have a tests for current consensus.

## Solution

This adds some, only a small bug was detected and will be fixed in the near future.


d4hines:
Already merged via #821

PR #818, opened 2022-09-20
fix writing to disk

EduardoRFS:
## Depends

- [ ] #817

## Problem

After #814 the loading of data from disk works, but the writing is partially broken, as Unix.rename doesn't work across devices.

## Solution

Move to a single data folder and make a chain.tmp.bin file after it is written then we rename it, making the chain.bin update stateless.


d4hines:
Already merged via #821

PR #819, opened 2022-09-20
V1 workaround when loading from disk

EduardoRFS:
## Depends

- [ ] #817

## Problem

When the node stops the chain contain some properties that assume liveness, this leads to some bugs such as gossip always having a pending request.

## Solution

I added Chain.clear, that clears some of the state on chain allowing it to always start again, this is not ideal but makes the behaviour more predictable.


d4hines:
Already merged via #821

PR #821, opened 2022-09-21
V1 json instead of marshal

EduardoRFS:
## Depends

- [ ] #819

## Problem

Because of https://github.com/ocaml/ocaml/issues/11548 we have some very nasty memory leaks

## Solution

Use deriving yojson for everything, this is not ideal but should work.


d4hines:
Woops - merged into wrong branch.

PR #829, opened 2022-09-24
V1 wasm

d4hines:
WASM support for v1


d4hines:
This feature isn't finished but I'm going to merge now to keep everyone in sync

PR #832, opened 2022-09-26
V1 alphanet restart

Pilou97:
# Problem:
chain.json does not contain the last state of the chain

# Solution
Sqlite has all the blocks, we just need to apply the missing blocks


d4hines (approved):
Thanks for this!

PR #840, opened 2022-09-27
To string instead of json string

ulrikstrid:
## Problem

```
Failure("cannot deserialize \"2b97aeab16c6ce67abd1fbc66782540a7a6f584800d44da84e2afa316747c7ff\" to blake2b")
```

## Solution

Create a proper string instead of json formatted string


aguillon:
Thank you.

PR #842, opened 2022-09-28
Update flakes and related fixes

ulrikstrid:
## Problem

`Piaf-lwt` has a known FD leak that is fixed in a later version.

## Solution

Update flake inputs to get a newer version of `Piaf-lwt` that has the fix.


src/stdlib/parallel.ml
```diff
@@ -5,7 +5,7 @@ module Pool = struct
   type t = pool
 
   let make ~domains =
-    let pool = Task.setup_pool ~num_additional_domains:domains () in
+    let pool = Task.setup_pool ~num_domains:domains () in
```
ulrikstrid, line 8: `Domainslib` was updated

src/network/network.ml
```diff
@@ -22,6 +22,16 @@ let internal_error error =
 let raw_expected_hash_header = "X-Raw-Expected-Hash"
 let broadcast_endpoint = "/messages"
 let request_endpoint = "/request"
+let clients : (Uri.t, Piaf_lwt.Client.t) Hashtbl.t = Hashtbl.create 16
+
+let get_client uri =
```
ulrikstrid, line 27: Re-use the same connection so that we don't get FD leaks when blocks are created too fast

PR #846, opened 2022-09-30
Deku toolkit nodejs

fraidev:
## Problem

Deku-toolkit doesn't work with Nodejs.

## Done with this PR

- [x] Use rollup to create bundles to CommonJS and ES6
- [x] Publish a new package on npm (0.1.4)
- [x] Fix react example
- [x] Create a node example that works on nodejs 18

## Why only node 18?

With this PR, deku-toolkits only works on node 18, because it's using `fetch`, but node only have fully support to `fetch` on node 18. 
I tried install `node-fetch`, but that library doesn't have some streamer functions that it's needed for this project. 
I recommend another task to replace `fetch` usages to another library that works on the old node versions too.

## Testing
The example shows a address balance:

image


aguillon:
Thank you.

PR #857, opened 2022-10-04
Using rapper ppx

Pilou97:
# Problem:

Requests of the indexer were poorly typed

# Solution:
Using ppx rapper


EduardoRFS (approved):
anime girl smilling at code improvement

PR #858, opened 2022-10-04
fix: support slash at the end of urls

aguillon:
Fixes (one of) Gauthier's bugs:
```
    DekuToolkit({
        dekuRpc: "https://deku-canonical-vm0.deku-v1.marigold.dev/"
```

was making the library fail.


fraidev:
Your PR is right! 
But you are using a node library. It's not a big problem because even for frontend users the most cases there are bundlers to help with it. But if we can make this library easy to use even without a bundler, I recommend you use another library or solution for that.

d4hines (approved):

PR #861, opened 2022-10-04
Add prefix to crypto

InfiniteSwerve:
# Problem
To ensure wallets can sign Deku transactions, we need to identify Deku transactions in the first place.
# Solution
This PR adds an optional prefix to all crypto `sign` and `verify` functions, allowing the inclusion of information about the Deku chain in transaction hashes. This uniquely identifies a transaction hash with a chain, and prevents double spending attacks across different Deku chains.

### DIFF
--- a/src/crypto/BLAKE2b.ml
+++ b/src/crypto/BLAKE2b.ml
@@ -30,10 +30,11 @@ struct
   let digest_size = P.digest_size
 
   module With_alg (Alg : Hash_intf.Alg) = struct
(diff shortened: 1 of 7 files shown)


d4hines:
Can you add the following:
- propagate the `prefix` option to `Identity.sign` and `Identity.verify` respectively
- use the `"\080"^chain_id` as the `prefix` everywhere in `Operation` module for signing and verifying
- pass in the correct `chain_id` everywhere we  constract an operation.

Then this will be fully supported.

PR #863, opened 2022-10-05
fix: intercept "close" message in SDK and update examples/tutorial

aguillon:
The JS SDK is giving the "close" message to the VM, which causes annoying messages when mocking a transaction.

Please take some time to if the changes to the package-lock.json files make sense, I find them surprisingly long. I guess we haven't committed them for a long time?

(Edit: the name of the branch is a misnomer, this isn't really about interop)


d4hines:
Merged in #865

PR #864, opened 2022-10-05
Deku cli v1 for Deku-P tutorial

aguillon:
Based on https://github.com/marigold-dev/deku/pull/863 and a first commit from @Pilou97.

Adds a deku-cli exe and three commands:
- mock-transaction (was create-mock-transaction in v0)
- submit-transaction (was create-custom-transaction in v0)
- generate-identity (was a different executable, which is removed) to generate wallets.

All three commands are used by @gsebil08 in his tutorial.

The remaining commands (transaction, withdraw, etc.) will be added in a separate PR.

PR #865, opened 2022-10-05
Arthur/doc/deku p tutorial

aguillon:
Based on https://github.com/marigold-dev/deku/pull/863 and https://github.com/marigold-dev/deku/pull/864 

@gsebil08 branch rebased. Contains everything for the Deku-P tutorial.


d4hines (approved):
Rebased and approved. One caveat - I want to keep the `generate-identity` util for a while longer since I use it to setup tesnets. We can probably get rid of it at some point though.

Issue #874, opened 2022-10-11
Optimizing fees used for Tezos interop


d4hines:
We should ensure we're paying a reasonable fee for posting state root updates. The default by Taquito is too high.

aguillon:
The fee formula is here: https://www.opentezos.com/tezos-basics/economics-and-rewards#transaction-cost . The question though, is why does Taquito return a significantly higher estimate than what''s actually needed?

PR #876, opened 2022-10-11
feat: set commit fee to a fixed amount

aguillon:
Adds an option to set the amount to a fixed amount in mutez. In the code, the fee is of type `int option` to account for situations where we're not committing on Tezos at all (i.e, where Deku is not used as a sidechain).

### Related issue
https://github.com/marigold-dev/deku/issues/874

### Check list
On flextesa, using `source scripts/common.sh`, when running with the `--preferred-fee=1000` option:
* `tezos_client get balance for alice` show that the balance decreases by 0.001 for each commit
* I can still make a deposit, a withdraw and generate a proof and the

### Limitations
* Not tested on Ghostnet or Mainnet
* The user has to know the minimum fee to use, which varies with the number of validators


src/bin/node/deku_node.ml
```diff
@@ -172,7 +175,7 @@ let setup_log ?style_renderer ?level () =
     (Logs.Src.list ())
 
 let main () =
-  setup_log ~level:Logs.Debug ();
+  setup_log ~level:Logs.Info ();
```
d4hines, line 178: Let's not sneak this change in.

aguillon: Right, thanks.

PR #877, opened 2022-10-12
Crypto add prefix

InfiniteSwerve:
# Problem
To ensure wallets can sign Deku transactions, we need to identify Deku transactions in the first place.

# Solution
This PR adds magic bytes (0x80) and the chain_id in each operation.


EduardoRFS:
Please split the chain_id change from the crypto prefix one

InfiniteSwerve:
@EduardoRFS #[666387c](https://github.com/marigold-dev/deku/pull/877/commits/666387c8b0bee3f14c3a61f2033aae4c32d198e8) is split into #[a41c077](https://github.com/marigold-dev/deku/pull/877/commits/a41c0773f9230984d9273824c49b774c1bc26db5) and #[fb54008](https://github.com/marigold-dev/deku/pull/877/commits/fb5400822bb1f3b40d52cab9b4065a78085ff16a) and has been rebased.

d4hines:
We added the magic byte separately. It's tested and in main now.

PR #879, opened 2022-10-12
Split API

Pilou97:
# Problem:
The api is embeded in the node. 
If we want to make some changes to the API we have to redeploy the node and the api.
If the API crashes, we have to restart the whole node.

# Solution:
Separate the API from the node,

# Notes
The e2e test is still working.


d4hines (approved):
:+1: 
image

PR #901, opened 2022-10-14
Add okrs githook

ulrikstrid:
Not sure this is the best way to go about it since pre-commit-hooks seems to not allow any input


d4hines:
I don't think we want a git hook after all.

PR #902, opened 2022-10-17
[deku-toolkit] add vmState endpoint and example

gsebil08:
## Context

Get VM state was missing in Deku Toolkit.
This PR aims to add it.

## This PR contains

- new function available in deku-toolkit to fetch VM State
- example of the previous point in the `react-app`
- use local deku-toolkit in the react example
- changed version of deku-toolkit to 0.1.9 (not published)

### DIFF
--- a/deku-p/deku-toolkit/examples/react/package.json
+++ b/deku-p/deku-toolkit/examples/react/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@airgap/beacon-sdk": "^3.1.3",
-    "@marigold-dev/deku-toolkit": "^0.1.6",
+    "@marigold-dev/deku-toolkit": "^0.1.9",
     "@taquito/beacon-wallet": "^13.0.1",
     "@taquito/signer": "^13.0.1",
     "@taquito/taquito": "^13.0.1",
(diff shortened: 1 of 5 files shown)


client/examples/react/src/App.tsx
```diff
@@ -4,8 +4,9 @@ import logo from "./logo.png";
 import {
   DekuToolkit,
   fromMemorySigner,
-} from "@marigold-dev/deku-toolkit";
+} from "deku-toolkit";
 import { InMemorySigner } from "@taquito/signer";
+import { JSONType } from "../../../lib/utils/json";
```
Pilou97, line 9: Do not use the JSONType, it's a private type, it's not exposed by the client

client/examples/react/src/App.tsx
```diff
@@ -187,6 +202,10 @@ const App = () => {
           onClick={() => withdrawExample()}
           children="Withdraw (to same address as the ticketer)"
         />
+        <button
+          onClick={() => getVmStateExample()}
+          children="VM State"
+        />
```
Pilou97, line 208: You don't need to add this button.
But if you want to set the state when you click on this button then you have to do the same as your useEffect

Pilou97 (changes requested):

client/examples/react/src/App.tsx
gsebil08: Done, thank you 🙏

PR #910, opened 2022-10-21
Deku toolkit cleanup (wip)

aguillon:
Hi guys, this is me trying to learn JavaScript.

Can you please check that
1) deku-toolkit only test is correctly configured and it works?
2) the eslint, jest and prettier configurations make sense?

This should also fix the bug you reported @fraidev, by usying @tzstamp/helpers as you suggested. It looks ok in the browser as well.

For me, the next steps are to continue cleaning the console.logs everywhere and include a TzPortal-like test to the library.


fraidev (approved):
I tested locally with `deku-tester` and it seems to be working now. 

Please, publish it and update the version on package.json and yarn.

image

Pilou97 (approved):

PR #912, opened 2022-10-24
Fix e2e tests

Pilou97:
# Problem:
 - The e2e tests are not working anymore after the migration to the monorepo architecture

# Solution: 
 - Adapt some paths in test scripts
 - force exit in eio


d4hines:
Closing since it was merged in #879

d4hines:
Actually this test is broken in main but working on this last commit.

PR #931, opened 2022-10-26
adds interface to tezos data used in testing

InfiniteSwerve:
# Problem 
We want to prevent regressions in `Deku_crypto`
# Solution
This is the head of a PR chain which will eventually add tests using data generated on Tezos as the source of truth.
This PR adds an interface to the generated Tezos data


EduardoRFS:
Please make the PRs a bit bigger, I don't have enough context of what this interface file will do by this PR alone. Small PRs are not about small contributions it is about small functionalities.

This interface alone has 21 values, it would be better if it had 1 value but also the implementation using it.

PR #932, opened 2022-10-26
Add Tezos-generated data

InfiniteSwerve:
# Problem 
To use Tezos as a source of truth, we need Tezos-generated data
# Solution
This PR adds data generated on Tezos for access in tests in future PR's.
# Dependencies 
- [ ] #931


EduardoRFS:
Okay this is great, but the same comment as #931 , you added a lot of different tests, it would be better to do a single one completley first.

EduardoRFS:
Also please, follow the PR template.

InfiniteSwerve:
Closing this and all related PR's, and opening new ones with vertical slicing.

PR #935, opened 2022-10-27
fix: replace eprintf with logs

aguillon:
…except in tests

### DIFF
--- a/deku-p/src/core/bin/api/api_state.ml
+++ b/deku-p/src/core/bin/api/api_state.ml
@@ -49,7 +49,7 @@ let find_withdraw_proof ~operation_hash state =
           withdrawal_handles_hash )
   | _ ->
       (* FIXME? fragile *)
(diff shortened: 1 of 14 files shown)


ulrikstrid:
Some of these should maybe be `warn`?

aguillon:
@ulrikstrid for gossip-related failures, you mean? I'm not sure I understand the subtleties of warning vs error.

ulrikstrid:
As I see it a `error` is something that is not expected to happen so when it does we want to be notified (via alerts).
A `warning` is something that's not "good" but we don't expect anyone to act on it immediately.

@d4hines and @Laucans, do you have any other ideas here?

d4hines:
@ulrikstrid @aguillon @Laucans I don't expect us to be able to figure out the best policy without trial and error tbh. I'll do a quick comb-through of these changes and see if any make more sense as alerts, then I'll merge. We can follow up with changes as needed.

PR #937, opened 2022-10-27
Initial crypto tests 

InfiniteSwerve:
## Problem

We want to prevent regressions in the Deku_crypto library

## Solution

This PR sets up all the infrastructure for running tests on `Alg_intf.S` and provides a single example test.

## Related

This is the head of a chain which will incorporate regression tests for `Alg_intf.S`.

### DIFF
--- a/deku-p/src/core/crypto/tests/tezos_test_data.ml
+++ b/deku-p/src/core/crypto/tests/tezos_test_data.ml
@@ -1,0 +1,14 @@
+module type Tezos_data = sig
+  val public_keys : string list
+end
+
+module Ed25519_data : Tezos_data = struct
+  let public_keys =
+    [
+      "edpktgopG88M5eE8M6N1ZtHbYzDCXRnGXk9vhNrLnYp9CA6aVyMRXa";
+      "edpkvLs7dfWXcdx62iEW5wpYfn8yKPuj446BCRhEbevMMbSSE9G1Yn";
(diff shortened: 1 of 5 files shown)


deku-p/src/core/crypto/tests/tezos_test_data.ml
```diff
@@ -0,0 +1,14 @@
+module type Tezos_data = sig
+  val public_keys : string list
+end
+
+module Ed25519_data : Tezos_data = struct
+  let public_keys =
```
EduardoRFS, line 6: You can add a comment showing which line you executed on `dune utop`.

Or alternatively you can just add a dependency to tezos-crypto and use here directly, instead of hard coding the cases, this would be better as it would also detect changes to tezos-crypto, but can be done as a future improvement.

EduardoRFS (changes requested):
It looks okay, besides of the comment, my only recommendation is to avoid indirections.

deku-p/src/core/crypto/tests/tezos_test_data.ml
InfiniteSwerve: The tezos-crypto data generation was done through a script that requires `tezos-crypto` to run. Until we can have it available on Deku as a dev dependency, the script will be included as a .txt file.

Once all the data is included and `tezos-crypto` can be used on Deku, the generated data will be replaced by direct comparisons between the output of `tezos-crypto` and `deku-crypto`.

d4hines (approved):
I added `tezos-crypto` to deku main. Agree with Edu's concerns, but we'll incorporate `tezos-crypto` in a follow-up pr.

PR #939, opened 2022-10-28
fix/wasm-vm: error in serialization/deserialzation

Pilou97:
# Problem:
Impossible to apply a vm operation

# Solution:
Fix the serialization and deserializarion thanks to @Zett98

### DIFF
--- a/deku-p/src/core/protocol/protocol.ml
+++ b/deku-p/src/core/protocol/protocol.ml
@@ -64,18 +64,20 @@ let apply_operation ~current_level protocol operation :
             let receipt = Receipt.Vm_transaction_receipt { operation = hash } in
             let result () =
               let%some ledger =
-                Ledger.with_ticket_table ledger (fun ~get_table ~set_table ->
-                    let tickets =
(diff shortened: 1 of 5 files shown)


Zett98 (changes requested):
This line should be dropped

deku-p/src/core/protocol/protocol.ml
```diff
@@ -110,6 +112,7 @@ let apply_operation ~current_level protocol operation :
               with
               | vm_state, ledger -> Some (ledger, Some receipt, vm_state, None)
               | exception External_vm_client.Vm_execution_error error ->
+                  let () = failwith error in
```
Zett98, line 115: This is a mistake from my end, forgot to drop it before sending the patch

Pilou97: no worries, I'll remove it

gsebil08 (approved):
Tested on my side, was able to mint a cookie!!

d4hines (approved):
LGTM

PR #941, opened 2022-10-28
feat: endpoint to query operation receipt

Pilou97:
# Problem:

It's impossible to know if an operation has been included. Except by pulling each second the state

# Solution:
Add a way to get the receipt of an operation


d4hines (approved):
LGTM :+1:

PR #948, opened 2022-10-28
move chain to data encoding

EduardoRFS:
## Problem

Currently we're using Yojson for both serializing to disk and for broadcasting blocks, this is not ideal.

## Solution

Here I added all the encodings need to put the entire Chain on data encoding. The plan is to remove Yojson altogether.


EduardoRFS:
This makes the chain much slower, due to the time spent serializing, but this will be fixed soon.

PR #949, opened 2022-10-28
Ocaml_vm

Zett98:
## Problem

We dont have a wasm vm written in ocaml on Deku

## Solution

Add it


deku-p/src/core/protocol/protocol.ml
```diff
@@ -112,6 +112,7 @@ let apply_operation ~current_level protocol operation :
               with
               | vm_state, ledger -> Some (ledger, Some receipt, vm_state, None)
               | exception External_vm_client.Vm_execution_error error ->
+                  let () = failwith error in
```
d4hines, line 115: This doesn't look right.

d4hines (changes requested):

EduardoRFS (changes requested):
Why did you vendor the interpreter? I assume it is because you made changes, if so which changes? Clean the git history.

EduardoRFS:
Can we go without the WASM changes for now? Specially without the unicode one's, they're tricky to review and I would lilke more time on those.

EduardoRFS:
Also @renatoalencar please review it

deku-c/wasm-vm-ocaml/env.mlid
```diff
@@ -0,0 +1,15 @@
+(* module Contract_address = Deku_ledger.Contract_address
+   module Address = Deku_ledger.Address
+   module Ledger = Deku_ledger.Ledger
+
```
EduardoRFS, line 4: Delete this file?

deku-p/src/core/ledger/ticket_id.mli
```diff
@@ -4,7 +4,7 @@ type ticketer =
 [@@deriving eq, ord, show]
 
 type ticket_id = private Ticket_id of { ticketer : ticketer; data : bytes }
-and t = ticket_id [@@deriving ord, yojson, show]
+and t = ticket_id [@@deriving ord, eq, yojson, show]
```
EduardoRFS, line 7: Why do you need this eq?

deku-c/wasm-vm-ocaml/value.mli
```diff
@@ -0,0 +1,36 @@
...
+  and t =
+    | Int of Z.t
+    | String of string
+    | Bytes of bytes
+    | Bool of int
+    | Pair of t * t
+    | Union of union
+    | List of t list * typ
+    | Option of t option
+    | Unit
```
EduardoRFS, line 18: Unit should be on tuna

EduardoRFS:
Those changes are okay, most of the code is gonna die, so if possible to remove the changes to the interpreter, let's go with it.

deku-c/wasm-vm-ocaml/env.mlid
Zett98: updated the mli

deku-p/src/core/ledger/ticket_id.mli
Zett98: to not do polymorphic eq, but the current vm will be dropped anyways, so i'll remove the eq derivation for now

Zett98:
> Those changes are okay, most of the code is gonna die, so if possible to remove the changes to the interpreter, let's go with it.

I'd keep the changes for now, as they simplified gas metering and decoding and encoding of idents.
With regards to utf-8, Does it even matter? Vm will be rewritten soon enough, and it works with current contracts

deku-p/src/core/protocol/protocol.ml
d4hines: Pavlo explained that it's fixed in the next PR.

d4hines:
From my POV it would be better to merge https://github.com/marigold-dev/deku/pull/954 into this one and merge them together atomically. Other than that LGTM

EduardoRFS:
From my POV this is bad, but it is an improvement over whatever we have. So let's go with it.

deku-c/wasm-vm-ocaml/operation.ml
```diff
@@ -0,0 +1,17 @@
+open Deku_ledger
+open Deku_concepts
+
+type t = {
+  operation : Operations.t;
```
renatoalencar, line 5: I don't see why there's an Operation and an Operations module and how operations could more than one thing, the names should probably change in order to settle the confusion.

deku-c/wasm-vm-ocaml/env.ml
```diff
@@ -0,0 +1,229 @@
...
+                         in
+                         ignore @@ Wasm.Eval.invoke func [ arg ]);
+                    Some (fun (k : (a, _) continuation) -> continue k ())
+                | _ -> None);
+          }
+    with
+    | Failure s -> Error ("unexpected failure. Failure " ^ s)
+    | Wasm.Eval.Crash _ | Wasm.Eval.Exhaustion _ | Wasm.Eval.Link _
+    | Wasm.Eval.Trap _ ->
+        Error "Invalid contract supplied to the vm"
```
renatoalencar, line 173: Don't ignore the actual errors, that makes debugging harder. I think something similar to line 53 is enough for now.

renatoalencar (changes requested):
Looks fine in general, just two small things I've noticed.

renatoalencar (approved):
Looks good

d4hines (approved):
:+1:

PR #951, opened 2022-10-29
Fixes newlines in logs and cycles drops

aguillon:
Newlines in logs were introduced by https://github.com/marigold-dev/deku/commit/de6c7ec6acc00def73ee7ede7e7a3b816e74b6dc - sorry about that

The second commits fixes a production bug which goes like this:
- node 3 fails for some reason, say at level 205185
- the other nodes continue
- when node 3 gets back online, the other nodes reached level 205211, but every block below 205200 was dropped
- node 3 starts receiving and accumulating blocks > 205201, doesn't apply any of them, fills its RAM and dies.

Example of output (before fix) with a cycle of 30 instead of 600:
```
3: deku-node: [INFO] Chain started at level: 40
3: deku-node: [INFO] Incoming block Block [hash:
Db2n1fvMKHCUifW6ggPG4oucDjSsJFg9bEXFSmpZFVLzWTX1Vsgi, level: 61]
3: deku-node: [INFO] accepted: 61
3: deku-node: [INFO] Incoming block Block [hash:
Db2QD8ibFzbYMgb68MV6cX2ywGcG7JdroD1T7BqgqFKAc9z7nNeG, level: 63]
3: deku-node: [INFO] accepted: 63
3: deku-node: [INFO] Incoming block Block [hash:
Db2YZpyZoqMBLNsryy8hfrnGTX2TtoazpPUK6vUrmppLPeDYStcw, level: 62]
3: deku-node: [INFO] accepted: 62
3: deku-node: [INFO] Incoming block Block [hash:
Db2QD8ibFzbYMgb68MV6cX2ywGcG7JdroD1T7BqgqFKAc9z7nNeG, level: 63]
3: deku-node: [INFO] accepted: 63
```

### DIFF
--- a/deku-p/src/core/bin/node/deku_node.ml
+++ b/deku-p/src/core/bin/node/deku_node.ml
@@ -15,7 +15,7 @@ let make_dump_loop ~sw ~env ~folder =
     let chain = Eio.Promise.await promise in
     (try Storage.Chain.write ~env ~folder chain
      with exn ->
-       Logs.err (fun m -> m "storage.failure: %s\n%!" (Printexc.to_string exn)));
+       Logs.err (fun m -> m "storage.failure: %s" (Printexc.to_string exn)));
     loop ()
   in
   let dump chain =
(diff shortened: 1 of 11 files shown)


d4hines (changes requested):
The last commit `fix: drop previous cycle of blocks instead of current one` should be it's own PR, since that's a significant change.

It looks right to me, but it's something I would want to test first.

Other than that, looks good :+1:

aguillon:
As requested on Slack, here's how I tested it:
- change cycle length to 30
- ./start.sh
- wait for network to reach cycle > 30
- kill one node (e.g., 3)
- wait for network to reach cycle > 60
- kill the whole network
- restart, see if 3 catches up

I had to kill the whole network because start.sh/start_node wasn't tested but it should work by just restarting node 3.

d4hines (approved):
I'm satisfied with the testing

PR #954, opened 2022-10-29
@zett98/drop external vm

Zett98:
## Depends

- [ ] #949
## Problem
We need to have a wasm vm written in ocaml, instead of the one in external vm
## Solution

drop the external_vm and plug in the ocaml vm


deku-p/src/cli/mock_transaction.ml
```diff
@@ -1,72 +1,72 @@
-open Deku_concepts
-open Deku_protocol
-open Deku_stdlib
-open Deku_external_vm
-open Cmdliner
-open Common
+(* open Deku_concepts
```
d4hines, line 1:
```suggestion
(* We've dropped support for the External_vm currently.
TODO: uncomment this when we add it back *)
(* open Deku_concepts
```

deku-c/wasm-vm-ocaml/env.ml
```diff
@@ -4,14 +4,16 @@ module Ledger = Deku_ledger.Ledger
 open Deku_stdlib
 
 type t = {
-  self : Contract_address.t;
   sender : Address.t;
```
EduardoRFS, line 7: Why the runtime changes here?

EduardoRFS (changes requested):
Please remove the external vm library, and clear a bit of the commit history, especially the changes regarding the wasm-vm-ocaml library

deku-c/wasm-vm-ocaml/env.ml
Zett98: moved to prev commit

deku-p/src/core/bin/node/deku_node.ml
```diff
@@ -129,7 +127,6 @@ let main params style_renderer log_level =
   (* The VM must be started before the node because this call is blocking  *)
   Logs.info (fun m ->
```
renatoalencar, line 128: This log message is misleading without an actual process/ipc being started, I think it could be removed alongside with the `External_vm_client.start_vm_ipc` call.

renatoalencar (changes requested):
For now I think the only issue is that portion of code that should be removed alongside the rest of the external vm interface.

renatoalencar (approved):

PR #968, opened 2022-11-03
feat: minimum block latency

d4hines:
## Problem

There are times when you want to slow the chain. The "natural" way is to have a higher workload, but sometimes the ideal spot is not on the workload-latency curve (e.g. when you're on a laptop or cheap hardware).

## Solution

Add a config for how many seconds to sleep in between blocks.

There's a problem with this patch, and that is that it's coupled to saving blocks. But what if we save blocks in batch later? Ideally we would sleep immediately before/after applying. But how do we want to inject config into the `Chain` module?

### DIFF
--- a/deku-p/src/core/chain/chain.ml
+++ b/deku-p/src/core/chain/chain.ml
@@ -116,6 +116,12 @@ let commit ~current_level ~block ~votes ~validators =
       withdrawal_handles_hash;
     }
 
+(* TODO: move this to a transparently configurable thing in Cmdliner in Deku_node *)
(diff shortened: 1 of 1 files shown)


d4hines:
Where ever we add this in the end, we don't want it to slow down actual block application. I.e, applying blocks should go as fast as possible - it's only when we're trying to advance the chain that we should potentially slow down.

Which makes me think maybe we sleep before producing blocks instead of before applying..

d4hines:
merged into main.

PR #969, opened 2022-11-04
fix: insert or ignore block level

d4hines:
## Problem

Every time we load from disk, we try to insert blocks already in the database. 

## Solution

I don't understand our startup sequence enough to know if this is a good patch or not. It might be that we want to fix the problem
upstream (maybe we should not be raising the `Chain_save_block` event at all in this this case).

But if we want to allow idempotent writes to block storage, this works.


d4hines:
Closing since this is not the right fix.

PR #972, opened 2022-11-07
fix/test: e2e tests

Pilou97:
# Problem

The e2e test is broken again, because wrong arguments are passed to the deku-node

# Solution

Remove start of vms, and remove useless deku-node argument

# Warning
If you want to test the e2e test, please change your docker-compose.yml file to use the last version of flextesa (this bug is not related with this PR)


d4hines (approved):
LGTM

PR #973, opened 2022-11-07
Js cli

aguillon:
Base deku-cli programs for Deku-P and Deku-C. Unfortunately it took me a while to have Deku-C version running so it's very incomplete right now, but I'll update it in parallel of testing Deku-C.


deku-c/deku-c-toolkit/src/contract.ts
```diff
@@ -82,48 +83,97 @@ export class Contract {
...
    */
-  async getState(): Promise<any | null> {
-    const state = await this.getRawState();
-    if (state === null) return null;
-    return parseContractState(state);
+  async invokeLigo(
+    code: string,
+    expression: string,
+    ligoRpc: string,
+    dekuRpc: string
```
Pilou97, line 118: Why do you need the dekuRpc ? Can't it be passed to the contract via it's constructor ? And be a attribute of the contract ?

d4hines:
FYI, when the file already exists, I get this:
```
npx deku-cli generate-identity -o wallet.json
node:internal/process/promises:279
            triggerUncaughtException(err, true /* fromPromise */);
            ^

[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "File already exists".] {
  code: 'ERR_UNHANDLED_REJECTION'
  ```

File: deku-c/deku-c-toolkit/src/contract.ts (hunk 1 above)
Comment (aguillon) [reply deku-c/deku-c-toolkit/src/contract.ts:131]: In needed it to compile the Michelson code before posting the operation:

```js
const result = await fetch(ligoRpc + "/api/v1/ligo/expression", options);
const { expression } = await result.json();
return compileExpression(dekuRpc, { expression, address });
```

This was before being able to access dekuRpc into the toolkit, but I guess we don't need to pass it around now.

PR #975, opened 2022-11-08
Fix toolkit parsing

Pilou97:
# Problems:

 - The deku-c-toolkit can not parse several types: Union, List, Unit 
 - Bug in the Pair parsing

# Solution:

- Implements parsing for missing types
- Fix the pair parsing bug

# Publishing

A new version has been set: 0.1.3
The website was updated to this version.


gsebil08 (approved):
LGTM

PR #978, opened 2022-11-11
Prepare benchmarks for cli usage

InfiniteSwerve:
## Problem

We want to have easy access to benchmarks, and also be able to easily export them to csv for use in the dashboard

## Solution

This PR parameterizes all benchmarks to use a common module interface while preserving original functionality.


d4hines (approved):
LGTM :+1: 
gotr grr fast

PR #979, opened 2022-11-12
Add csv output option and domain selection to deku-benchmark 

InfiniteSwerve:
## Problem

We want to be able to write benchmark data to csv files, and choose the amount of domains from the command line
## Solution

This PR uses cmdliner to allow optional writing to a csv file, with a mandatory domain choice

PR #980, opened 2022-11-12
Add basic statistics for deku benchmarks

InfiniteSwerve:
## Depends
- [ ] #979 

## Problem

We want some helper statistics when using deku-benchmark
## Solution
adds average and per-second statistics to terminal output of the benchmark


d4hines (commented):
LGTM
speed racer approves

PR #981, opened 2022-11-13
JS-CLI: invoke with ligo expression

aguillon:
Depends on https://github.com/marigold-dev/ligo-deku-rpc/pull/5


aguillon:
Sorry Edu for the pings.

@Pilou97 so you tell me that I should not use "file:.." in package.json files?

PR #983, opened 2022-11-14
network conf: limit promtail logs collection to avoid flood on loki

Laucans:
When deku-node bug it produce 1.5m logs per hours, that's a lot for loki

adding this conf to limit collection to 50 per second (instead of 1000 right now..)
          limit = {
            rate = 50;
            burst = 50;
          };

Not tested yet, gonna check on the futur deployment of the RC


scripts/deploy_contracts.sh
```diff
@@ -18,7 +18,7 @@ deploy_contract() {
 
     echo "Originating $1 contract"
     sleep 2
-    octez_client \
+    tezos-client \
```
ulrikstrid, line 21: I think this is fixed to be `octez-client` on latest `tezos-nix`

Laucans: Should not be in this PR I removed it

PR #986, opened 2022-11-16
JS Interop: better error management and logging

Pilou97:
# Problem:

- Wrong level of error from the js interop
- Errors from the interop are not explicit

# Solution:
 - Parsing the errors that come from the interop and apply a good log level on them

Is a screen of the new log in case a operation can't be submitted due to an insufficient balance:

Capture d’écran 2022-11-16 à 12 17 05


EduardoRFS (approved):
anime girl approve a straightforward PR

PR #989, opened 2022-11-21
[deku-c-toolkit] Expose cancelOnNewState function

Gau-thier:
## Depends

- [ ] #988 

## Problem

There is no function exposed by the toolkit to cancel the interval of `onNewState`.
Bump version to 0.1.6

## Solution

Add a `cancelOnNewState` function to clear the interval.

### DIFF
--- a/deku-c/deku-c-toolkit/src/contract.ts
+++ b/deku-c/deku-c-toolkit/src/contract.ts
@@ -95,7 +95,7 @@ export class Contract {
   }
 
   /**
-   * Invoke a deku-c smart contrat
+   * Invoke a Deku-C smart contrat
    * @param parameter the parameter of the contract
    * @returns the hash of the operation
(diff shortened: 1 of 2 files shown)


deku-c/deku-c-toolkit/src/contract.ts
```diff
@@ -152,6 +152,10 @@ export class Contract {
           return null;
         })
         .catch(console.error);
-    }, 2000);
+    }, 1000);
+  }
+
+  cancelOnNewState(): void {
+    if (this.fetchInterval) clearInterval(this.fetchInterval);
   }
```
Pilou97, line 160: Could you add a doc to explain what this function does ?

gsebil08: Done.

PR #991, opened 2022-11-23
WIP add contract address in receipts

gsebil08:
## Problem

Contract address is not available in the receipts. Which means we cannot get it from `/operations/{hash}` in the API

## Solution

Add the contract address to the receipts

## Test it
- run deku locally via `./start.sh`
- originate a contract (if you want to originate decookies via deku_helper, please copy/paste the decookies.wasm from feat/decookie-dekuc branch)
- get the hash of the operation in the logs
- execute
```bash
$ curl http://localhost:8080/api/v1/operations/Do3q6fNiXe2rEJ6ui8fJ6vGqiuT47R9rTnLbAZthrrBKCZsAchS4

# Return must be something like following:
["Vm_transaction_receipt",{"operation":"Do3q6fNiXe2rEJ6ui8fJ6vGqiuT47R9rTnLbAZthrrBKCZsAchS4","contract_address":"DK19EoKv6622gK1icUwG8E3ziPqKyqEMsx3N"}
```


deku-p/src/core/ledger/address.mli
```diff
@@ -1,7 +1,9 @@
 open Deku_stdlib
 open Deku_crypto
 
-type address
+type address =
```
gsebil08, line 4: I had to expose the `Implicit` and `Originated` constructors

deku-p/src/core/protocol/protocol.ml
```diff
@@ -73,7 +73,23 @@ let apply_operation ~current_level protocol operation :
...
+            let contract_address =
+              match operation.operation with
+              | View { address; _ } -> Some address
+              | Originate _ ->
+                  Some
+                    (Contract_address.of_user_operation_hash
+                    @@ Operation_hash.to_blake2b hash)
+              | Call { address; _ } -> (
+                  match address with
+                  | Address.Implicit _ -> None
```
gsebil08, line 86: This case is the reason I defined:
```ocaml
    Vm_transaction_receipt of {
      operation : Operation_hash.t;
      contract_address : Deku_ledger.Contract_address.t option;
    }
```
I am not totally convinced this is the best way to do it, but don't have anything else in mind for now...

gsebil08: @d4hines I quickly exchanged with @Zett98 about that, and inside `Env.ml` something similar is done. Hence, I just did the same.
What do yo think? Did I misunderstand something?

deku-p/src/core/protocol/protocol.ml
```diff
@@ -73,7 +73,21 @@ let apply_operation ~current_level protocol operation :
             | Ok ledger -> (ledger, Some receipt, vm_state, None)
             | Error error -> (ledger, Some receipt, vm_state, Some error))
         | Operation_vm_transaction { sender; operation } -> (
-            let receipt = Receipt.Vm_transaction_receipt { operation = hash } in
+            let open Deku_ledger in
+            let contract_address =
+              match operation.operation with
+              | View { address; _ } -> address
+              | Originate _ ->
+                  Contract_address.of_user_operation_hash
```
d4hines, line 81:
```suggestion
                  (* TODO: wrap this in a function exposed from the VM *)
                  Contract_address.of_user_operation_hash
```

deku-p/src/core/protocol/protocol.ml
```diff
@@ -73,7 +73,21 @@ let apply_operation ~current_level protocol operation :
...
+                  @@ Operation_hash.to_blake2b hash
+              | Call { address; _ } ->
+                  Deku_ledger.Address.to_contract_address address
+                  |> Option.map fst |> Option.get
```
d4hines, line 85: This will throw an exception, and the user will not get any notification. This could be improved by giving the user an error receipt instead.

d4hines (approved):
LGTM!

PR #1006, opened 2022-12-01
Initial block storage test

InfiniteSwerve:
## Problem
We want to ensure the block storage works properly. Making sure it can save and load blocks effectively, and can do so for blocks of sufficient size.

## Solution
This PR includes the minimal working example of a single empty block being saved and loaded. It also makes some necessary changes to the worker api to allow the tests to finish.

### DIFF
--- a/deku-p/src/core/block_storage/tests/block_storage_tests.ml
+++ b/deku-p/src/core/block_storage/tests/block_storage_tests.ml
@@ -17,7 +17,11 @@ let block ~default_block_size =
   Producer.produce ~identity ~default_block_size ~above ~withdrawal_handles_hash
     producer
 
-let uri = Uri.of_string "sqlite3:/tmp/database.db" 
+let file_hash =
+  let randn = Stdlib.Random.int 230 in
+  Deku_crypto.BLAKE2b.hash (Int.to_string randn) |> BLAKE2b.to_hex
+
+let uri = Uri.of_string (Format.sprintf "sqlite3:/tmp/%s.db" file_hash)
 
 let make_block_storage env sw =
   let domains = Eio.Stdenv.domain_mgr env in
--- a/deku-p/src/core/block_storage/tests/block_storage_tests.ml
+++ b/deku-p/src/core/block_storage/tests/block_storage_tests.ml
@@ -1,0 +1,64 @@
+open Deku_stdlib
+open Deku_crypto
+open Deku_consensus
+open Deku_concepts
+
+let block_testable = Alcotest.testable Block.pp Block.equal
+
+let identity =
+  let secret = Ed25519.Secret.generate () in
+  let secret = Secret.Ed25519 secret in
+  Identity.make secret
+
+let block ~default_block_size =
+  let above = Genesis.block in
+  let withdrawal_handles_hash = BLAKE2b.hash "potato" in
+  let producer = Producer.empty in
+  Producer.produce ~identity ~default_block_size ~above ~withdrawal_handles_hash
+    producer
+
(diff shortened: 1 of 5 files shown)


deku-p/src/core/stdlib/parallel.mli
```diff
@@ -7,6 +7,7 @@ module Worker : sig
   type t = worker
 
   val make : domains:#Eio.Domain_manager.t -> sw:Eio.Switch.t -> worker
+  val teardown : worker -> unit
```
EduardoRFS, line 10: I understand why this was done, but it is quite bad as now you may have a copy of a worker that is not working, so some piece of code may try to schedule work and never receive the result.

But you can achieve the same by killing the ~sw in make, this will make everything on that switch dies and you can control granularity by making smaller switches.

deku-p/src/core/block_storage/tests/block_storage_tests.ml
```diff
@@ -0,0 +1,68 @@
...
+  let withdrawal_handles_hash = BLAKE2b.hash "potato" in
+  let producer = Producer.empty in
+  Producer.produce ~identity ~default_block_size ~above ~withdrawal_handles_hash
+    producer
+
+let file_hash =
+  let randn = Stdlib.Random.int 230 in
+  Deku_crypto.BLAKE2b.hash (Int.to_string randn) |> BLAKE2b.to_hex
+
+let uri = Uri.of_string (Format.sprintf "sqlite3:/tmp/%s.db" file_hash)
```
EduardoRFS, line 24: Can we have in memory database?

deku-p/src/core/block_storage/tests/block_storage_tests.ml
```diff
@@ -0,0 +1,68 @@
...
+    ~msg:"hash loaded block is not equal to saved block" ~expected:block
+    ~actual:retrieved_block;
+
+  let (Deku_block_storage.Block_storage.Storage { worker; _ }) =
+    block_storage
+  in
+  Parallel.Worker.teardown worker
+
+let eio_test_case : (Eio.Stdenv.t -> unit -> unit) -> unit -> unit =
+ fun f () -> Eio_main.run (fun env -> f env ())
```
EduardoRFS, line 55: you should never run Eio_main.run many times.

EduardoRFS (changes requested):

deku-p/src/core/stdlib/parallel.mli
InfiniteSwerve: I see. I'll keep generating more tests for now like we discussed offline, but I'll make sure to address this in a future PR.

deku-p/src/core/block_storage/tests/block_storage_tests.ml
InfiniteSwerve: Not sure. It looks like this is an option in [SQLite itself](https://www.sqlite.org/inmemorydb.html), but probably won't be available in the OCaml ecosystem.

deku-p/src/core/block_storage/tests/block_storage_tests.ml
InfiniteSwerve: Ok. As it's not a bug, and matters for performance, I'll address this in a future PR.

InfiniteSwerve:
@EduardoRFS I've made the changes you requested. Please review them, and if there are no blocking issues, merge this PR.

InfiniteSwerve:
@EduardoRFS I've rebased and, except for the in-memory sqlite database, implemented the changes you requested. Please take a look and if there are no blocking issues, merge the PR.

PR #1008, opened 2022-12-02
bug: on_message callback on listen to node function

Pilou97:
Commit cherry-picked from rc-4

# Problem

The API does not define the on_message callback when listening to the node

# Solution

Define the callback


d4hines (approved):
LGTM :+1:

PR #1017, opened 2022-12-07
Fix push bool primitives and NOW

renatoalencar:
## Problem

`PUSH bool True` and `PUSH bool False`  doesn't work. Also the `$now` import was previously removed.

## Solution

They needed to be implemented, also put `$now` back.


d4hines (approved):
LGTM

PR #1018, opened 2022-12-07
chore: kubernetes wants entrypoint to be a string array

ulrikstrid:
## Problem

Kubernetes will complain if EntryPoint is a string instead of string array

```
error: code = Unknown desc = failed to pull and unpack image "ghcr.io/marigold-dev/deku:latest": unmarshal image config: json: cannot unmarshal string into Go struct field ImageConfig.config.Entrypoint of type []string
```

## Solution

Make it into a array


d4hines:
Merged into main already.

PR #1019, opened 2022-12-09
feat: python client library

aguillon:
Edit: I completely failed to install the necessary requirements with nix.

I think this is mergeable as such because I would use it in other projects (before making PRs to improve it). It still lacks
- tests
- proper MyPy setup
- strengthening against errors in http.py
- contract state
- support for Ligo and (non-Michelson) WASM
- ~~doc~~ docstrings
- ~~black setup~~

`examples/` both work in a fresh environment, when targeting Canonical (the default). I don't expect it to work with `main` though, but I believe we have the same issue with the JS library?

### DIFF
--- a/deku-c/python-client/README.md
+++ b/deku-c/python-client/README.md
@@ -1,39 +1,5 @@
 # Python Deku Client Library
 
-## Installation
-
-The Deku library requires a few librairies for cryptography, which are dependencies of
-[PyTezos](https://pytezos.org/quick_start.html):
-
-```bash
-sudo apt install libsodium-dev libsecp256k1-dev libgmp-dev
-```
-
-The Deku library also currently requires Python 3.10. We recommend using
-[pyenv](https://realpython.com/intro-to-pyenv/), which allows to manage several different version of
-Python on your system, as well as various virtual environments. Once `pyenv` is installed, create a
-virtual environment by using the following commands:
-
-```bash
-$ pyenv install 3.10  # If not already installed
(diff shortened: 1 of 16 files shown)


docs/Deku-Canonical/deku_c_python.md
```diff
@@ -0,0 +1,48 @@
...
+## Examples
+
+Two examples are given in the `examples/` directory:
+* `examples/bridge.py` demonstrates a deposit from Tezos to Deku using a pre-deployed contract on
+  Ghostnet, as well as the withdraw operation. This contract uses a pre-defined implicit account on
+  Ghostnet, but you may have to change it and/or use the
+  [faucet](https://faucet.ghostnet.teztnets.xyz/) to refill it.
+* `examples/contracts.py` shows the origination of a contract on Deku and invocation of an
+  entrypoint of this contract. You can then observe that the contract storage has changed using the
+  [Deku TypeScript CLI](https://deku.marigold.dev/docs/Deku-Canonical/deku_c_cli).
```
d4hines, line 48:
```suggestion
  [Deku CLI](https://deku.marigold.dev/docs/Deku-Canonical/deku_c_cli).
```

deku-c/python-client/README.md
```diff
@@ -0,0 +1,39 @@
+# Python Deku Client Library
```
d4hines, line 1: Can we just link to the docs from this page? Less chance of failing to keep the two docs in sync.

PR #1022, opened 2022-12-14
Fix memory leak

Pilou97:
# Problem

There is a memory leak, and the cause seems to be the Data_encoding.Json of the Block_storage

# Solution

Insert the block as binary


d4hines:
Merged into main already.

PR #1023, opened 2022-12-14
Fix slow down

Pilou97:
# Problem

The chain is very slow, it seems that `Data_encoding.Json` is the cause of this slow down

# Solution

Change `Data_encoding.Json` into `Data_encoding.Binary` in message.ml


d4hines:
Merged into main already.

PR #1024, opened 2022-12-15
handle set while parsing contract state

gsebil08:
## Problem

If the state of the contract contains a `set`, toolkit won't parse it correctly and return `null` value.

## Solution

Parse a `set` as a list.

### DIFF
--- a/deku-c/client/package.json
+++ b/deku-c/client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@marigold-dev/deku",
-  "version": "0.1.4",
+  "version": "0.1.3",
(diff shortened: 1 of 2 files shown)


deku-c/client/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@marigold-dev/deku",
-  "version": "0.1.3",
+  "version": "0.1.4",
```
d4hines, line 3:
```suggestion
  "version": "0.1.3",
```

d4hines (approved):
LGTM :+1:

d4hines:
I'm going to cherry-pick this to main directly.

PR #1027, opened 2022-12-19
Nix: Update ligo reference to the one in nix pkgs

Laucans:
Use the ligo reference wrapped in nix-core instead of the one wrapped in ligo repository

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -15,7 +15,6 @@
     dream2nix.url = "github:nix-community/dream2nix";
     dream2nix.inputs.nixpkgs.follows = "nixpkgs";
 
-    ligo.url = "gitlab:ligolang/ligo";
     tezos.url = "github:marigold-dev/tezos-nix";
     tezos.inputs = {
       nixpkgs.follows = "nixpkgs";
@@ -33,7 +32,6 @@
     dream2nix,
     tezos,
     deploy-rs,
-    ligo,
     treefmt,
     ...
   }:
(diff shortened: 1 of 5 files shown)


Laucans:
Please @ulrikstrid can you validate stuff and if possible update the nix flake.lock correctly. I'm not sure about update inside

flake.nix
```diff
@@ -78,7 +76,7 @@
             inherit pkgs;
             ligo =
               if system == "x86_64-linux"
-              then ligo.packages.${system}.ligoLight
+              then pkgs.ligo
```
ulrikstrid: I think this could be changed to just always be `pkgs.ligo` since the one in `nixpkgs` works on both m1 and linux

nix/deku-p/default.nix
```diff
@@ -42,10 +42,10 @@
     ligo =
       # TODO: support ligo on more systems in our flake
       if system == "x86_64-linux"
-      then inputs.ligo.packages.${system}.ligoLight
+      then pkgs.ligo
```
ulrikstrid: This can always be `pkgs.ligo`

ulrikstrid (changes requested):

flake.nix
Laucans: I did it. Please check my other comment

ulrikstrid (approved):
The `flake.lock` changes looks good, LGTM

